PR #130, opened 2020-08-21
checks that debug, error, info, trace, warn ranges don't overlap

benmkw:
this is not a fix for an issue I had, just noticed this while reading the code and did it for no reason ;)


japaric (approved):
Thanks!

PR #138, opened 2020-08-26
Look for quoted and unquoted version symbol

jonas-schievink:
Should hopefully close https://github.com/knurling-rs/probe-run/issues/36


mattico:
I can confirm that this fixes the issue.

PR #143, opened 2020-08-27
improve parser errors for missing `:` in params

### LINKED ISSUE #141: Better error messages for missing type specifier
I started experimenting with format strings and got these errors that aren't too helpful with the actual problem.
They should probably indicate that defmt requires type specifiers unlike std formatting macros.

```
error: missing `}` in format string
 --> src/bin/hello.rs:9:19
  |
9 |     defmt::error!("Hello, world: {}", x);
  |                   ^^^^^^^^^^^^^^^^^^
```

```
error: malformed format string
 --> src/bin/hello.rs:9:19
  |
9 |     defmt::error!("Hello, world: {0}", x);
  |                   ^^^^^^^^^^^^^^^^^^^
```

birkenfeld:
Fixes #141

### DIFF
--- a/parser/src/lib.rs
+++ b/parser/src/lib.rs
@@ -93,7 +93,7 @@ fn parse_param(mut s: &str) -> Result<Param, Cow<'static, str>> {
     let mut index = None;
     let colon_pos = s
         .find(|c: char| !c.is_digit(10))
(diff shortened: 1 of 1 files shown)


jonas-schievink:
Thanks, great improvement!

Issue #142, opened 2020-08-27
Support formatting &[u8] as ASCII strings


birkenfeld:
I'm often working with byte slices that have string (or mixed) contents, however it's not worth doing the `str::from_utf8` dance *just* for logging purposes.

So it would be very convenient to have a format specifier that outputs the slice contents as ASCII, with non-printable and non-ASCII bytes either replaced or escaped.

I'm willing to implement it in case it's deemed acceptable :)

jonas-schievink:
This seems useful to have, yeah. There's an interesting design question though: This is, for all intents and purposes, still a `{:[u8]}`, just with a display hint, so it is very similar to a hypothetical `{:u8:x}` modifier that prints an integer in hexadecimal (syntax tbd, we don't support this at the moment). I'd very much like to see the hex modifier at some point, so maybe it makes sense to make this an `:ascii` modifier, like `{:[u8]:ascii}`?

japaric:
For the same rationale as https://github.com/knurling-rs/defmt/issues/145#issuecomment-681887216 I would like to see this handled in the frontend.

Now if we are talking about formatting a `AsciiStr` type (which sadly is not in core) I would be in favor of adding a new formatting paremeter for that type. Though I don't think that's what birkenfeld is after.

birkenfeld:
@japaric I see the point, especially for integers it's nice to get different reprs automatically, without the firmware specifying it explicitly.

Still, since I (and many others) will be happy with the simple stdout logging that `probe-run` provides, it'd be nice to get a commitment that it will support this, and not only some (hypothetical) web/GUI frontend :)

japaric:
this feature is being considered in the 'display hints' RFC (see #270). let's centralize the conversion in that thread. I'm going to close this thread to avoid having parallel on-going conversations

PR #144, opened 2020-08-27
add "bstr" format, for readable ASCII byte strings

birkenfeld:
As I wrote in issue #142, ASCII/mixed byte strings are very common in embedded, and usually we don't want to make a UTF8 check roundtrip just for logging purposes.

The only question here is the format for the unprintable/non-ASCII bytes, I chose `<FF>` since it's easier to read in large quantities than `\xFF` but I'm fine with any change you request.

### DIFF
--- a/book/src/primitives.md
+++ b/book/src/primitives.md
@@ -29,4 +29,5 @@ The available types are:
 - `:f32`, 32-bit floating point type
 - `:[u8; N]`, byte array
 - `:[u8]`, byte slice
+- `:bstr`, byte slice represented as ASCII string
 - `:str`, string slice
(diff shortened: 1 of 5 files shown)


jonas-schievink:
This seems reasonable to have, but it might be best if this was a presentation hint similar to the ones proposed in https://github.com/knurling-rs/defmt/issues/145#issuecomment-712235944 (if we decide to go that route) instead of a brand new formatting type.

japaric:
Thanks for the PR. We are unlikely to implement this format as a separate type but we'll likely implement this as a 'display hint'. The design of this 'display hints' feature is being discussed in RFC #270.

Until the design of the feature is finalized and approved I'm going to close PRs related to the feature -- without an approved design those PRs cannot be reviewed.

If you would like to share your implementation of RFC #270 (note: the design is still in flux so you may need to update your implementation several times) with others you can post the link to your branch in thread #270 and I'll link it from the issue description.

PR #147, opened 2020-08-28
Default zero timestamp

### LINKED ISSUE #137: Default timestamp
Not all apps need timestamped logs, but currently every app must provide a `#[defmt::timestamp]` function.

Instead, we could provide a default zero timestamp source as a weak symbol, and have `#[defmt::timestamp]` override that. The host-side tooling would then omit time stamps if they're zero (or rather, until any timestamp is received that's non-zero).

We could also do something more fancy, like turning timestamps off altogether (so they'd take 0 Bytes of transferred data instead of at least 1 per log frame), but I feel like that ties into support for custom timestamp formats/encodings, which is a bigger feature.

jonas-schievink:
Closes #137

### DIFF
--- a/book/src/setup-app.md
+++ b/book/src/setup-app.md
@@ -47,38 +47,6 @@ The following `global_logger`s are provided as part of the project:
 
 Information about how to write a `global_logger` can be found in the [`#[global_logger]` section](./global-logger.md).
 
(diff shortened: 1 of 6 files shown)


japaric (approved):
Great!

PR #148, opened 2020-08-28
extract location info (file+line number) from DWARF

### LINKED ISSUE #19: [research] extract span information from DWARF?
`proc_macro::Span::{source_file,start,end}` are unstable. See if we can instead extract span information from DWARF information, which is part of the ELF metadata.

If we add this metadata to a log frame then we could (a) display it in the console output or (b) implement go to definition in interactive front ends.

japaric:
closes #19 
this is used to display location info in `probe-run`

### DIFF
--- a/elf2table/src/lib.rs
+++ b/elf2table/src/lib.rs
@@ -147,12 +147,8 @@ pub fn get_locations(object: &object::File) -> Result<Locations, anyhow::Error> 
                 let mut decl_file = None;
                 let mut decl_line = None; // line number
                 let mut name = None;
-                let mut has_type = false;
-                let mut is_external = false;
                 let mut location = None;
-                let mut has_linkage_name = false;
 
-                let mut has_unexpected_attribute = false;
                 cached_attrs.clear();
                 while let Some(attr) = attrs.next()? {
                     match attr.name() {
@@ -160,17 +156,6 @@ pub fn get_locations(object: &object::File) -> Result<Locations, anyhow::Error> 
                             if let gimli::AttributeValue::DebugStrRef(off) = attr.value() {
                                 name = Some(off);
                             }
-                        }
-
-                        gimli::constants::DW_AT_type => {
-                            has_type = true;
-                        }
-
-                        // #[export_name]
-                        gimli::constants::DW_AT_external
-                            if attr.value() == gimli::AttributeValue::Flag(true) =>
-                        {
-                            is_external = true;
                         }
 
                         gimli::constants::DW_AT_decl_file => {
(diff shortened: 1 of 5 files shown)


elf2table/src/lib.rs
```diff
@@ -1,17 +1,15 @@
...
-use object::{File, Object, ObjectSection};
+use object::{Object, ObjectSection};
 
 /// Parses an ELF file and returns the decoded `defmt` table
 ///
 /// This function returns `None` if the ELF file contains no `.defmt` section
-pub fn parse(elf: &[u8]) -> Result<Option<Table>, anyhow::Error> {
-    let elf = File::parse(elf)?;
-
+pub fn parse(elf: &object::File) -> Result<Option<Table>, anyhow::Error> {
```
japaric, line 12: now I'm no long sure if changing `[u8]` to `object::File` here was the right call. It avoids re-parsing the ELF in `probe-run` but makes this API (semver) depend on the version of `object` as one of its types is publicly exposed. This would be the only type from a dependency that's exposed in the public API (I think)

jonas-schievink: Yeah I went from `xmas_elf::ElfFile` to `&[u8]` when switching to `object` to avoid having to care about public dependencies.

I expect host-side tools to always be built optimized, so parsing the ELF headers twice should have negligible impact.

elf2table/src/lib.rs
```diff
@@ -92,3 +91,187 @@ pub fn parse(elf: &[u8]) -> Result<Option<Table>, anyhow::Error> {
...
+    } else {
+        bail!("no `FileEntry` for index {}", index)
+    };
+
+    let mut s = String::new();
+    if let Some(dir) = file.directory(header) {
+        let dir = dwarf.attr_string(unit, dir)?;
+        let dir = dir.to_string_lossy()?;
+
+        if !dir.starts_with('/') {
```
jonas-schievink, line 244: Will this still work on Windows? Maybe converting to a `PathBuf` and using `is_absolute` would be more portable.

elf2table/src/lib.rs
```diff
@@ -92,3 +91,187 @@ pub fn parse(elf: &[u8]) -> Result<Option<Table>, anyhow::Error> {
...
+            if let Some(ref comp_dir) = unit.comp_dir {
+                s.push_str(&comp_dir.to_string_lossy()?);
+                s.push('/');
+            }
+        }
+        s.push_str(&dir);
+        s.push('/');
```
jonas-schievink, line 251: The `/` theoretically would break on Windows when the base path uses `\\?\` UNC paths. Not sure if that's a problem, but might be worth at least a `// FIXME`.

elf2table/src/lib.rs
japaric: change the whole function to use the Path API

jonas-schievink (approved):

Issue #152, opened 2020-08-31
defmt ignoring log when called as log argument


daschl:
Sorry for the confusing title, but here is what I found.

So let's say I have  a struct which has the following function:

```rs
pub fn read(&self) -> i32 {
        defmt::info!("read called");
}
```

Now when I call it like so:

```rs
let external_temp = external_temp_sensor.read();
defmt::info!("Temp is: {:i32}", external_temp);
```

Everything is fine and both info lines are printed. But when it it is called like this:

```rs
defmt::info!("Temp is: {:i32}",  external_temp_sensor.read());
```

Only  the "outer" info is logged and the inner `read called` info event is ignored and does not show up. Is this expected? I'm using probe-run 0.1.3 from master and defmt main branch from github as well (as described in the blog post on how to use it)

japaric:
>  Is this expected?

no, this is a bug in the evaluation order. the arguments should be evaluated before we acquire the logger

the current expansion (`cargo expand`):
``` rust
defmt::info!("{:i32}", foo());
// ^ expands into
if true && true || !true && true {
    if let Some(mut _fmt_) = defmt::export::acquire() {
        match (defmt::export::timestamp(), &(foo())) {
            (ts, arg0) => {
                _fmt_.istr(&defmt::export::istr(match () {
                    #[cfg(not(target_arch = "x86_64"))]
                    () => {
                        #[link_section = ".defmt.info.{:i32}"]
                        #[export_name = "{:i32}@#7 bytes(139..169)"]
                        static DEFMT_LOG_STATEMENT: u8 = 0;
                        &DEFMT_LOG_STATEMENT as *const u8 as usize
                    }
                }));
                _fmt_.leb64(ts);
                _fmt_.i32(arg0);
                _fmt_.finalize();
                defmt::export::release(_fmt_)
            }
        }
    }
}
```
I think the correct order should be: evaluate log level (the `if true && ..` line); evaluate format arguments (`foo()`); `acquire` the logger; compute the `timestamp`; push data into logger (the `_fmt_` method calls); `release` call

PR #153, opened 2020-09-01
Fix argument/acquire evaluation order

jonas-schievink:
Fixes https://github.com/knurling-rs/defmt/issues/152

### DIFF
--- a/macros/src/lib.rs
+++ b/macros/src/lib.rs
@@ -449,9 +449,9 @@ fn log(level: MLevel, ts: TokenStream) -> TokenStream {
     let logging_enabled = is_logging_enabled(level);
     quote!({
         if #logging_enabled {
-            if let Some(mut _fmt_) = defmt::export::acquire() {
-                match (defmt::export::timestamp(), #(&(#args)),*) {
-                    (ts, #(#pats),*) => {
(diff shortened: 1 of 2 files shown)


macros/src/lib.rs
```diff
@@ -449,9 +449,9 @@ fn log(level: MLevel, ts: TokenStream) -> TokenStream {
     let logging_enabled = is_logging_enabled(level);
     quote!({
         if #logging_enabled {
-            if let Some(mut _fmt_) = defmt::export::acquire() {
-                match (defmt::export::timestamp(), #(&(#args)),*) {
-                    (ts, #(#pats),*) => {
+            match (defmt::export::timestamp(), #(&(#args)),*) {
```
japaric, line 452: let's evaluate `timestamp` last so that when you use the AtomicUsize timestamp function the timestamps will appear in order. I think right now something like this `info!("x={:?}", foo())` will print
``` text
0.000002 foo evaluated
0.000001 x=something
```

japaric (approved):
Changes looks good to me but I think it'd be better to evaluate `timestamp` after the arguments

macros/src/lib.rs
jonas-schievink: Yeah, sounds good

PR #166, opened 2020-09-07
fix impl Format for Str (interned string)

### LINKED ISSUE #164: crash when printing interned string with {:?}
this code:

```rust
    info!("this works: {:istr}", intern!("hello world!"));
    info!("this crashes: {:?}", intern!("hello world!"));
    info!("this will never be reached");
```

prints

```
0.000004 INFO this works: hello world!
stack backtrace:
   0: 0x00029e4c - HardFault
      <exception entry>
...
```

The 2nd line should either work, or fail to compile, but not crash at runtime.

japaric:
closes #164

### DIFF
--- a/src/impls.rs
+++ b/src/impls.rs
@@ -93,7 +93,7 @@ impl Format for f32 {
 impl Format for Str {
     fn format(&self, fmt: &mut Formatter) {
         if fmt.needs_tag() {
-            let t = internp!("{:str}");
+            let t = internp!("{:istr}");
             fmt.u8(&t);
         }
         fmt.istr(self);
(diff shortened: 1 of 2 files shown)


src/impls.rs
```diff
@@ -93,7 +93,7 @@ impl Format for f32 {
 impl Format for Str {
     fn format(&self, fmt: &mut Formatter) {
         if fmt.needs_tag() {
-            let t = internp!("{:str}");
+            let t = internp!("{:istr}");
```
japaric, line 96: we forgot to update this format string when we renamed `{:str}` to `{:istr}` some time ago

jonas-schievink (approved):

PR #167, opened 2020-09-07
add module path to location info

### LINKED ISSUE #151: extract module path info from DWARF
while doing #19 it seemed possible to also compute the module path (`krate::module::function`) of each log statement.
we could include that info in the `probe-run` output as a lighter version alternative to the full path (or as complementary info)

implementation notes:
- [while traversing the DIE (Debugging Information Entry) tree](https://github.com/knurling-rs/defmt/blob/0af5bd2ccc70bcfb88fc9cbcc06be402a08baef1/elf2table/src/lib.rs#L145)
- when entering a `DW_TAG_namespace` node read the `DW_AT_name` attribute and use that as a segment; push that into a `path`
- when leaving the `DW_TAG_namespace` node remove (pop) the segment you added to `path`
- when you encounter a `DEFMT_LOG_STATEMENT` variable (a log statement) attach the current `path` to it
- I recommend running [`dwarfdump`](https://github.com/gimli-rs/gimli/blob/e046ca50e2d16ebe775a798d20ad947b7430859f/examples/dwarfdump.rs) on the firmware image you are working on to visualize the DIE tree
- I'm unsure if DIE from other crates need special handling

japaric:
closes #151

### DIFF
--- a/elf2table/src/lib.rs
+++ b/elf2table/src/lib.rs
@@ -101,6 +101,7 @@ pub fn parse(elf: &[u8]) -> Result<Option<Table>, anyhow::Error> {
 pub struct Location {
     pub file: PathBuf,
     pub line: u64,
+    pub module: String,
 }
 
 impl fmt::Debug for Location {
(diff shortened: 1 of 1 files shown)


elf2table/src/lib.rs
```diff
@@ -149,9 +150,30 @@ pub fn get_locations(elf: &[u8]) -> Result<Locations, anyhow::Error> {
...
-            if entry.tag() == gimli::constants::DW_TAG_variable {
+            if entry.tag() == gimli::constants::DW_TAG_namespace {
+                let mut attrs = entry.attrs();
+
+                while let Some(attr) = attrs.next()? {
+                    match attr.name() {
+                        gimli::constants::DW_AT_name => {
+                            if let gimli::AttributeValue::DebugStrRef(off) = attr.value() {
+                                let s = dwarf.string(off)?;
+                                for _ in (depth as usize)..segments.len() + 1 {
```
japaric, line 167: this is a bit magic but seems to work. I think it would be more correct to record the depth at which the first path segment appears instead of hardcoding a `1` there

jonas-schievink (approved):

PR #174, opened 2020-09-21
defmt-rtt: add non-blocking mode; set it as the default

japaric:
~~do not merge until there's a probe-run PR that changes the mode back to blocking at runtime~~
should be merged around at the same time as knurling-rs/probe-run#69


japaric:
this should be tested outside a `probe-run` environment: it should not block and it should not corrupt memory (have not tested myself)

Issue #176, opened 2020-09-21
proc-macro derive produced unparseable tokens


Dirbaio:
```rust
#[derive(defmt::Format)]
pub struct Set<'a, 'b: 'a> {
    foo: &'a u32,
    bar: &'b u32,
}
```

produces this error

```
error: expected one of `,` or `>`, found `:`
   --> foobar/src/lib.rs:140:22
    |
140 | pub struct Set<'a, 'b: 'a> {
    |                      ^ expected one of `,` or `>`

error: proc-macro derive produced unparseable tokens
   --> foobar/src/lib.rs:139:10
    |
139 | #[derive(defmt::Format)]
    |          ^^^^^^^^^^^^^
```

defmt version cf066cb7c9
rustc 1.48.0-nightly (6af1bdda5 2020-09-15)

PR #178, opened 2020-09-22
derive: fix handling of generic bounds

jonas-schievink:
This fixes bounds on generic parameters getting duplicated into the wrong place of the generated impl, which closes https://github.com/knurling-rs/defmt/issues/176.

This also makes the custom derive emit bounds on *type parameters* instead of *field types*, since the latter can apparently run into very obscure and impossible to debug compiler bugs like https://github.com/rust-lang/rust/issues/77052. This matches the behavior of Rust's built-in derives.


japaric (approved):
LGTM. we should fix the docs that currently say we add bounds on field types.

Issue #101, opened 2020-08-11
bitfield: further compression


japaric:
`info!("x={:10..14}", u16_value)` sends `u16_value` over the wire.
We can save bandwidth by sending only the higher byte `(u16_value >> 8) as u8`.

Lotterleben:
Thinking out loud:
- `Codegen::new()` already truncates unneeded high bits https://github.com/knurling-rs/defmt/blob/main/macros/src/lib.rs#L764 so we could extend that to be a bit more sophisticated (shift octet overlaps into octets) and truncate low bits as well
- on the decoder side, we'd have to pay attention to `range.start` in https://github.com/knurling-rs/defmt/blob/main/decoder/src/lib.rs#L628 and shift+cast the data we've read using `self.bytes.read_u*...`accordingly
- my gut feeling is that cutting out "middle bytes" involves more headscratching esp. on the parser side (but fundamentally we have the info to decipher it in the fmt string)

PR #184, opened 2020-10-08
enhance bitfield compression

Lotterleben:
resolves https://github.com/knurling-rs/defmt/issues/101 without the middle bytes compression part.

imo the code is pretty wonky in parts, so I'd be grateful for any feedback on how to make it more idiomatic (I've marked some of the weird parts with `TODO`). The bitfield deduplication in `prepare_params()` has been the biggest fight and I'm not really happy with the tangled index/tracking vars I ended up with.

### DIFF
--- a/decoder/src/lib.rs
+++ b/decoder/src/lib.rs
@@ -18,7 +18,7 @@ use std::{
 use byteorder::{ReadBytesExt, LE};
 use colored::Colorize;
 
-use defmt_parser::{Fragment, Level, Type};
+use defmt_parser::{Fragment, Level, Parameter, Type};
 
 include!(concat!(env!("OUT_DIR"), "/version.rs"));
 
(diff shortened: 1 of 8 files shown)


decoder/src/lib.rs
```diff
@@ -382,6 +382,83 @@ impl<'t, 'b> Decoder<'t, 'b> {
...
+        let mut num_params_read = 0;
+        while num_params_read < initial_num_params {
+            match &(&mut params[i]).ty {
+                Type::BitField(range) => {
+                    let range_start = range.start;
+                    let range_end = range.end;
+                    params.remove(i);
+
+                    match &mut curr_bitfield_range {
+                        Some(r) if range_start < r.start => {
```
Lotterleben, line 417: note to self: merge match arms into one, otherwise we will miss bigger ends if range_start is smaller too

decoder/src/lib.rs
```diff
@@ -382,6 +382,84 @@ impl<'t, 'b> Decoder<'t, 'b> {
...
+        // deduplicate bitfields by merging them into a new one with range min..max
+        // TODO refactor when `drain_filter()` is stable
+        // sorry about the wonky vars but accessing enum fields is too messy to just use a Param{}
+        let mut curr_bitfield_range: Option<Range<u8>> = None;
+        let mut curr_bitfield_index = 0;
+        let mut i = 0;
+        let initial_num_params = params.len();
+        let mut num_params_read = 0;
+        while num_params_read < initial_num_params {
+            match &(&mut params[i]).ty {
```
jonas-schievink, line 410: Any reason this doesn't work?

```suggestion
            match &params[i].ty {
```

decoder/src/lib.rs
```diff
@@ -382,6 +382,84 @@ impl<'t, 'b> Decoder<'t, 'b> {
...
+                _ => {
+                    i += 1;
+                }
+            }
+
+            num_params_read += 1;
+
+            // if we're at the end of the param list or handling a new param, flush our current bitfield
+            if (num_params_read == initial_num_params)
+                || ((&mut params[i]).index != curr_bitfield_index)
```
jonas-schievink, line 442:
```suggestion
                || (params[i].index != curr_bitfield_index)
```

decoder/src/lib.rs
```diff
@@ -623,26 +675,32 @@ impl<'t, 'b> Decoder<'t, 'b> {
...
-                        0..=16 => {
+                        2 => {
                             data = self.bytes.read_u16::<LE>()? as u64;
                         }
-                        0..=24 => {
+                        3 => {
                             data = self.bytes.read_u24::<LE>()? as u64;
                         }
-                        0..=32 => {
+                        5 => {
```
japaric, line 694: shouldn't this be `4` instead of `5`? The largest range is `0..32`; that gives `lowest_byte = 0`; `highest_byte = 3`; and `truncated_sz = 4`

decoder/src/lib.rs
```diff
@@ -623,26 +675,32 @@ impl<'t, 'b> Decoder<'t, 'b> {
...
                 Type::BitField(range) => {
-                    let data: u64;
-
-                    match range.end {
-                        0..=8 => {
+                    let mut data: u64;
+                    let lowest_byte = range.start / 8;
+                    // -1 because `range` is range-exclusive
+                    let highest_byte = (range.end - 1) / 8;
+                    let truncated_sz = highest_byte - lowest_byte + 1; // in octets
```
japaric, line 682: I think `size_after_truncation` (if truncation is an actual word) would be a clearer name.

japaric (commented):
Gave it a first pass but need to give the 'merge bitfields' part another read later

decoder/src/lib.rs
Lotterleben: you're right– fixed and added a test to catch any future regressions

decoder/src/lib.rs
Lotterleben: is an actual word; renamed

macros/src/lib.rs
```diff
@@ -759,26 +759,43 @@ impl Codegen {
                 defmt_parser::Type::BitField(_) => {
                     // TODO reused in decoder::parse_args(), can we share this somehow without Cargo bug troubles?
                     let all_bitfields = parsed_params.iter().filter(|param| param.index == i);
+                    // TODO can I get rid of use after move err other than with clone?
```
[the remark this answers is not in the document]
Lotterleben: okay! it felt a bit clunky to do this so I wasn't quite sure. Thank you for the explanation!

decoder/src/lib.rs
```diff
@@ -382,6 +382,84 @@ impl<'t, 'b> Decoder<'t, 'b> {
...
+                        a_range.start.cmp(&b_range.start)
+                    }
+                    _ => Ordering::Equal,
+                }
+            } else {
+                a.index.cmp(&b.index)
+            }
+        });
+
+        // deduplicate bitfields by merging them into a new one with range min..max
```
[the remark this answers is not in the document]
Lotterleben: good point. Done(ish), I've already found a bug and an idea on how to uncomplicate the whole thing (merge before sorting, don't try to merge in place). will implement in the afternoon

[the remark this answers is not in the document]
Lotterleben: done

parser/src/lib.rs
```diff
@@ -205,6 +205,35 @@ fn push_literal<'f>(
     Ok(())
 }
 
+/// returns Some(smallest_bit_index, largest_bit_index) contained in `params` if
+///         `params` contains any bitfields.
+///         None otherwise
+pub fn get_max_bitfield_range<'a, I: std::clone::Clone>(params: I) -> Option<(u8, u8)>
```
japaric, line 211: nit: you could move the `: Clone` bound to the where clause below. Also, it shouldn't be necessary to specify the full path of the `Clone` trait; it is always imported by the `core` / `std` [prelude](https://doc.rust-lang.org/core/prelude/v1/index.html).

japaric (approved):
Looks good to me! Feel free to squash and merge.

PR #191, opened 2020-10-16
remove common crate

Lotterleben:
As discussed with @jonas-schievink , remove the common crate since it only contains the `Level` enum used by `defmt-decoder`.
I also merged `Level` and `MLevel` and moved them into `parser` (not really the most intuitive location...) in a second commit in order to avoid future Level inconsistency woes, happy to revert that if undesired (not 100% sure if I got the implications of this change for https://github.com/knurling-rs/defmt/pull/175 right)


parser/src/lib.rs
```diff
@@ -303,6 +312,68 @@ pub fn parse<'f>(format_string: &'f str) -> Result<Vec<Fragment<'f>>, Cow<'stati
...
+            Level::Trace => "trace",
+            Level::Debug => "debug",
+            Level::Info => "info",
+            Level::Warn => "warn",
+            Level::Error => "error",
+        }
+    }
+
+    // returns a list of features of which one has to be enabled for this Level to be active
+    pub fn necessary_features(self, debug_assertions: bool) -> &'static [&'static str] {
```
jonas-schievink, line 327: This should maybe go into the macro crate, since it's only needed there

jonas-schievink (commented):
Seems like the parser doesn't actually need `Level`, but it should be fine to leave it there (it's better to get rid of the common crate entirely, I think)

Lotterleben:
I think we were at the point where we thought we could just remove earlier (during bool compression) but it turned out that there were some dependencies

parser/src/lib.rs
Lotterleben: if I understand it correctly I can't do that because that would spread the impl across different crates which is not allowed?

jonas-schievink: Yeah, it would have to be a free function

Lotterleben: welp, that makes sense. done

japaric (approved):
Looks good to me! Less implementation detail crates published on crates.io is always better. Feel free to merge after a rebase.

PR #192, opened 2020-10-19
store raw symbols in decoder::Table

japaric:
the raw symbols are used to check if a symbol is "live" when extracting location information from dwarf
fixes location information in PR #175

### DIFF
--- a/decoder/src/lib.rs
+++ b/decoder/src/lib.rs
@@ -50,6 +50,26 @@ impl Tag {
 }
 
 #[derive(Debug)]
+pub struct TableEntry {
+    string: StringEntry,
+    raw_symbol: String,
+}
+
+impl TableEntry {
+    pub fn new(string: StringEntry, raw_symbol: String) -> Self {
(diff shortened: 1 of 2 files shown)


decoder/src/lib.rs
```diff
@@ -113,8 +133,9 @@ impl Table {
         self.entries.is_empty()
     }
 
+    /// Iterates over the raw symbols of the table entries
     pub fn symbols<'s>(&'s self) -> impl Iterator<Item = &'s str> + 's {
```
japaric, line 137: this method relies on `Item` being a 'raw symbol' (as it appears in the `.symtab`)

jonas-schievink (approved):
Nice!

decoder/src/lib.rs
jonas-schievink: We could rename it to `raw_symbols` if that doesn't cause too much churn, to make it a bit clearer

japaric: sure!

PR #193, opened 2020-10-19
check defmt version before demangling symbols

japaric:
otherwise `probe-run` prints a cryptic "Error: expected value at line 1 column 1" when it's not
compatible with `defmt` used by the application

### DIFF
--- a/decoder/src/lib.rs
+++ b/decoder/src/lib.rs
@@ -103,10 +103,10 @@ Try `cargo install`-ing the latest git version of `probe-run` AND updating your 
 }
 
 impl Table {
-    pub fn new(entries: BTreeMap<usize, TableEntry>, version: &str) -> Result<Self, String> {
-        check_version(version)?;
-
-        Ok(Self { entries })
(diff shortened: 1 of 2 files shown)


decoder/src/lib.rs
```diff
@@ -87,15 +87,24 @@ pub struct Table {
...
 impl Table {
     pub fn new(entries: BTreeMap<usize, TableEntry>, version: &str) -> Result<Self, String> {
-        if version != DEFMT_VERSION {
-            return Err(format!(
-                "defmt version mismatch (firmware is using {}, host is using {}); \
-                 are you using the same git version of defmt and related tools?",
-                version, DEFMT_VERSION,
-            ));
-        }
+        check_version(version)?;
```
jonas-schievink, line 107: Seems like this could be removed now, and we could just always have the caller do it. This would allow skipping the version check too, which I've wanted a couple of times.

japaric: will remove it and document that the caller should do `check_version`

jonas-schievink (approved):

PR #196, opened 2020-10-20
make version check compatible with crates.io releases

### LINKED ISSUE #189: add support for crates-io to version logic
put the major part of the crate version into the linker script (instead of commit hash)
this is being done by build.rs right now; adjust that to not use git (but figure out if/how we need to keep this for git installs?)

japaric:
tested by `rm -rf .git`-ing a local checkout; got `"_defmt_version_ = 0.1" = 1;` in the `defmt.x` linker script
closes #189

### DIFF
--- a/build.rs
+++ b/build.rs
@@ -1,3 +1,4 @@
+use semver::Version;
 use std::{env, error::Error, fs, path::PathBuf, process::Command};
 
 fn main() -> Result<(), Box<dyn Error>> {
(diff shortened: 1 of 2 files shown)


japaric:
note to self: improve the error message on version mismatch (should mention crates.io release vs git, etc.)

build.rs
```diff
@@ -1,12 +1,25 @@
...
+        String::from_utf8(output.stdout).unwrap()
+    } else {
+        // no git info -> assume crates.io
+        let semver = Version::parse(&std::env::var("CARGO_PKG_VERSION")?)?;
+        if semver.major == 0 {
+            // minor is breaking when major = 0
+            format!("{}.{}", semver.major, semver.minor)
+        } else {
+            // ignore minor, patch, pre and build
+            semver.major.to_string()
```
jonas-schievink, line 19: I think if there's pre-release info that is also a breaking change, but I guess we could just not use that

jonas-schievink (approved):
LGTM!

japaric:
will submit a separate PR to improve the error message. Merging this one

PR #197, opened 2020-10-20
Allow printing '@' character.

Dirbaio:
With the new JSON implementation it's trivial, just escape it. The JSON decoder automatically unescapes it.

### DIFF
--- a/firmware/qemu/src/bin/log.rs
+++ b/firmware/qemu/src/bin/log.rs
@@ -308,6 +308,9 @@ fn main() -> ! {
 
     defmt::info!("after nested log: {:?}", nested());
 
+    // printing @ is now allowed
+    defmt::info!("I can now print the @ symbol!");
+
     loop {
         debug::exit(debug::EXIT_SUCCESS)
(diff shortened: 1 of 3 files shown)


jonas-schievink:
Nice! Can you add a test to the QEMU test suite?

Dirbaio:
Done!

firmware/qemu/src/bin/log.rs
```diff
@@ -308,6 +308,9 @@ fn main() -> ! {
 
     defmt::info!("after nested log: {:?}", nested());
 
+    // printing @ is now allowed
+    defmt::info!("I can now print the @ symbol!");
```
japaric, line 312: Could you also add a test for `intern!` macro? Something like lines 66 and 68 but the interned string should contain the  `@` character.

japaric (approved):
Awesome! Could you please add one more test?

Dirbaio:
Donw

japaric:
Thanks!

PR #198, opened 2020-10-20
Add support for printing i64 and u64.

### LINKED ISSUE #160: Implement Format for i64/u64?
When trying to do this:

```rust
#[derive(defmt::Format)]
pub struct Instant {
    ticks: i64,
}
```

it fails because Format is not implemented for i64:

```
error[E0277]: the trait bound `i64: defmt::Format` is not satisfied
```

It would be great to be able to print 64-bit integers.

Dirbaio:
Implemented them the same way as existing integer types.

Seems to work great:

```rust
    info!("I can print u64s: {:u64} {:u64}!", u64::MIN, u64::MAX);
    info!("I can print i64s: {:i64} {:i64}!", i64::MIN, i64::MAX);
```
prints
```
0.003178 INFO  I can print u64s: 0 18446744073709551615!
0.003179 INFO  I can print i64s: -9223372036854775808 9223372036854775807!
```

Fixes #160

### DIFF
--- a/decoder/src/lib.rs
+++ b/decoder/src/lib.rs
@@ -610,6 +610,10 @@ impl<'t, 'b> Decoder<'t, 'b> {
                     let data = self.bytes.read_i32::<LE>()?;
                     args.push(Arg::Ixx(data as i64));
                 }
+                Type::I64 => {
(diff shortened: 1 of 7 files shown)


japaric (approved):
Nice! Thank you

PR #200, opened 2020-10-20
impl Format for str

### LINKED ISSUE #177: Implement Format for str
This code
```rust
        let s: &str = "hello";
        info!("my string is {:?}", s);
```

produces this error

```
error[E0277]: the trait bound `str: Format` is not satisfied
   --> foobar.rs:292:9
    |
292 |         info!("my string is {:?}", s);
    |         ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^ the trait `Format` is not implemented for `str`
    |
    = note: required because of the requirements on the impl of `Format` for `&str`
```

Not a huge issue because there are better ways to print strings, using `intern!()` or with `{:str}`. However being able to use `{:?}` can come in handy when writing macros or generic functions that need to log arbitrary types that could potentially be `str`.

Dirbaio:
fixes #177

### DIFF
--- a/firmware/qemu/src/bin/log.rs
+++ b/firmware/qemu/src/bin/log.rs
@@ -81,6 +81,7 @@ fn main() -> ! {
 
     let interned = defmt::intern!("interned string");
     defmt::info!("&str = {:str}", "string slice");
+    defmt::info!("&str = {:?}", "string slice");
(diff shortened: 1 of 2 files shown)


Lotterleben (approved):
This is awesome, thank you for your work!

PR #201, opened 2020-10-20
impl Format for alloc types (Box & co)

Dirbaio:
This implements Format for the `alloc` types: Box, Rc, Arc, Vec, String.

The impls are only there if  `alloc` Cargo feature is enabled in order to not pull in `alloc` if the user doesn't want it. I'm not really sure if this is the best way to do it, and maybe this interacts weirdly with the `#![cfg_attr(not(target_arch = "x86_64"), no_std)]` attribute the crate has. Maybe we should change the crate to have `alloc` and `std` features, and do these impls if either is enabled?

### DIFF
--- a/src/impls.rs
+++ b/src/impls.rs
@@ -262,3 +262,50 @@ where
         }
     }
 }
+
+#[cfg(feature = "alloc")]
+mod if_alloc {
+    use crate::{Format, Formatter};
+
+    impl<T> Format for alloc::boxed::Box<T>
+    where
+        T: ?Sized + Format,
+    {
+        fn format(&self, f: &mut Formatter) {
+            self.as_ref().format(f)
+        }
+    }
+
+    impl<T> Format for alloc::rc::Rc<T>
+    where
+        T: ?Sized + Format,
+    {
+        fn format(&self, f: &mut Formatter) {
+            self.as_ref().format(f)
+        }
+    }
+
+    impl<T> Format for alloc::sync::Arc<T>
+    where
+        T: ?Sized + Format,
+    {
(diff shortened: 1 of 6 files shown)


src/impls.rs
```diff
@@ -252,3 +262,50 @@ where
...
+#[cfg(feature = "alloc")]
+mod if_alloc {
+    use crate::{Format, Formatter};
+
+    impl<T> Format for alloc::boxed::Box<T>
+    where
+        T: ?Sized + Format,
+    {
+        fn format(&self, f: &mut Formatter) {
+            self.as_ref().format(f)
```
japaric, line 275: I would suggest using UFCS syntax (or whatever it was called) here to make sure we calling `T`'s format method and not recursively calling `Box<T>`'s. That is:

``` rust
T::format(self, f)
```

The first argument may need to be `&**self` or something like that. This comment applies to the other `Format` implementations.

japaric (approved):
Thanks for the PR!

> The impls are only there if alloc Cargo feature is enabled in order to not pull in alloc if the user doesn't want it

That's good. We want to make `alloc` opt-in because *applications* that use the `alloc` crate require a nightly toolchain and we don't want to force nightly on all users.

Apart from my inline comment we should have some tests for this but that will require modifying the CI script to enable the `alloc` feature, maybe create a second `qemu/src/bin/log.rs` file that depends on a memory allocator, etc. Let me know if you would be interested in working on that! Otherwise we can merge the PR as it is and create an issue to track adding the tests in the future.

Dirbaio:
All done!

japaric:
This is excellent work. Thank you!

Issue #188, opened 2020-10-12
make experimental members private


Lotterleben:
Desired result: when someone pull in the decoder crate they'll get an empty crate, unless they use the `unstable` cargo feature, in which case they'll get the API but with a warning that this is subject to change

## notes:
- mark decoder features as unstable in docs
- mark `common`, `macros` as unstable and internal (shouldn't be used at all)
- use `async-std` as a reference
- document usage in README or book?

PR #202, opened 2020-10-20
Hide features

Lotterleben:
resolves https://github.com/knurling-rs/defmt/issues/188

I've added some docs to `qemu-run` since it seemed like it could be useful to someone outside of knurling, but marked it as unstable just in case.

### DIFF
--- a/Cargo.toml
+++ b/Cargo.toml
@@ -49,3 +49,7 @@ debug = false
 debug-assertions = false
 opt-level = 0
 overflow-checks = false
+
+[package.metadata.docs.rs]
+features = ["unstable"]
+rustdoc-args = [ "--cfg docsrs" ]
--- a/elf2table/src/lib.rs
+++ b/elf2table/src/lib.rs
@@ -1,4 +1,9 @@
-//! Reads ELF metadata and builds an interner table
+//! Reads ELF metadata and builds an interner table.
+//! Used by [`defmt`](https://github.com/knurling-rs/defmt).
+
+#![cfg(feature="unstable")]
+#![cfg_attr(docsrs, feature(doc_cfg))]
+#![cfg_attr(docsrs, doc(cfg(unstable)))]
 
 mod symbol;
 
--- a/parser/Cargo.toml
+++ b/parser/Cargo.toml
@@ -4,3 +4,10 @@ edition = "2018"
 license = "MIT OR Apache-2.0"
 name = "defmt-parser"
 version = "0.1.0"
+
+[features]
+unstable = []
+
+[package.metadata.docs.rs]
+features = ["unstable"]
+rustdoc-args = [ "--cfg docsrs" ]
--- a/qemu-run/Cargo.toml
+++ b/qemu-run/Cargo.toml
@@ -8,5 +8,5 @@ version = "0.1.0"
 
 [dependencies]
 anyhow = "1.0.32"
-defmt-decoder = { path = "../decoder" }
-defmt-elf2table = { path = "../elf2table" }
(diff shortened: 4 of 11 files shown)


Cargo.toml
```diff
@@ -44,3 +44,6 @@ debug = false
 debug-assertions = false
 opt-level = 0
 overflow-checks = false
+
+[package.metadata.docs.rs]
+rustdoc-args = [ "--cfg docsrs" ]
```
jonas-schievink, line 49: I think we need to add this to all crates that make use of the `docsrs` cfg

elf2table/src/lib.rs
```diff
@@ -1,3 +1,4 @@
+//! INTERNAL; DO NOT USE. Please use the `defmt` crate to access the functionality implemented here.
```
jonas-schievink, line 1: `defmt` doesn't reexport this crate's contents, so `elf2table` should probably be unstable instead of internal

qemu-run/Cargo.toml
```diff
@@ -8,5 +8,11 @@ version = "0.1.0"
...
 anyhow = "1.0.32"
-defmt-decoder = { path = "../decoder" }
+defmt-decoder = { path = "../decoder", features = ["unstable"] }
 defmt-elf2table = { path = "../elf2table" }
+
+[features]
+unstable = []
+
+[package.metadata.docs.rs]
+rustdoc-args = [ "--cfg docsrs" ]
```
jonas-schievink, line 18: This package is marked `publish = false`, so it won't end up on crates.io, which means we don't have to bother with stability (unless we *do* decide that we want to publish it, but I don't really think we will – this program is just meant to power defmt's test suite, nothing else)

qemu-run/src/main.rs
```diff
@@ -1,3 +1,12 @@
+//! An alternative to the [`probe-run`](https://github.com/knurling-rs/probe-run) printer,
+//! used by [`defmt`](https://github.com/knurling-rs/defmt).
+//! Parses data sent by QEMU over semihosting (ARM Cortex-M only).
+//! *Printers* are *host* programs that receive log data, format it and display it.
```
jonas-schievink, line 4: The docs look good though, we can keep those

qemu-run/Cargo.toml
Lotterleben: wait, if that's possible why don't we just do that for e.g. `elf2table` too?

parser/Cargo.toml
```diff
@@ -4,3 +4,9 @@ edition = "2018"
 license = "MIT OR Apache-2.0"
 name = "defmt-parser"
 version = "0.1.0"
+
+[features]
+unstable = []
+
+[package.metadata.docs.rs]
+rustdoc-args = [ "--cfg docsrs" ]
```
jonas-schievink, line 12: Oh, I think we'll need an additional line in the docs.rs configuration:

```toml
features = ["unstable"]
```

Otherwise the crates will be empty when the docs are generated

japaric: I agree with Jonas. All the Cargo.toml that have an `unstable` feature should have that line under the `[docs.rs]` section. (Though I guess not showing any documentation at all may be another way to say that the API is not meant to be used..)

japaric (approved):
Looks good! Feel free to squash and merge after adding the `features` line to the Cargo.toml-s

parser/Cargo.toml
Lotterleben: yeah I thought that was the goal here :D will change in a minute

PR #204, opened 2020-10-21
Skip empty symbol names while attempting to parse json

MathiasKoch:
Avoid `Error: EOF while parsing a value at line 1 column 0` if `.defmt` section should contain a symbol with empty name.

This might occur if you run `arm-none-eabi-objcopy ${IN} ${OUT}`, where IN is the output ELF of `cargo build`.

### DIFF
--- a/elf2table/src/lib.rs
+++ b/elf2table/src/lib.rs
@@ -61,6 +61,8 @@ pub fn parse(elf: &[u8]) -> Result<Option<Table>, anyhow::Error> {
     // second pass to demangle symbols
     let mut map = BTreeMap::new();
     for (_, entry) in elf.symbols() {
+        // Skipping symbols with empty string names, as they may be added by
+        // `objcopy`, and breaks JSON demangling
         let name = match entry.name() {
             Some(name) if !name.is_empty() => name,
             _ => continue,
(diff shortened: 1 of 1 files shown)


japaric:
Thanks for the PR! Could you describe in more detail how to cause the error that this is fixing? Does the error come from `objcopy`? Or you running `objcopy` on a single ELF file before passing that to `probe-run`?

MathiasKoch:
Ahh, yeah. After reading my description again, I can see how it might be a bit ambiguous.  

If i compile a device side firmware file using defmt, and then run that ELF file through ? `objcopy`, the resulting ELF file looks 100% identical when looking at it with `nm` and `objdump`, but it contains `""` (empty string name) symbols in the `.defmt` section among others. I still don't understand why this happens though?!

japaric:
Thanks for clarifying. I'll reproduce the issue locally to get a better understanding at what `objcopy` is doing. (Mostly because I'm curious)

I'm OK with this PR. Could you add a comment to the source code indicating why we are skipping empty symbol names? (i.e. "they may be added by `objcopy`")

MathiasKoch:
Yeah sure. If you end up figuring out why it produces these empty strings, please let me know somewhere :) eg Matrix.. 

I have spent the better part of two full days trying to find out what happened, and have split my entire firmware to pieces in the process :detective:

japaric (approved):
Thanks!

PR #211, opened 2020-10-22
implement Format arrays, like {:[?;16]}

### LINKED ISSUE #206: how to implement Format for [T; N]?
It would be quite handy if we could print arrays of any Format type. I had a go at implementing it, but it seems there is already a conflicting implementation for [u8; N] (overlaps because u8 is also Format).

I think this is not possible without specialization. What do we do?

- Not implement it?
- Remove the [u8] impl?
- use `min_specialization` behind a feature, so you can at least have both in nightly?

Dirbaio:
this adds support for printing arrays of `Format`s.

The Format impl is changed to work on `[impl Format; N]` for all types. The `[u8; N]` impl is removed because it would overlap, as discussed in #206.

I've renamed "slice" and "array" to "u8_slice" and "u8_array" in a few places, so the names are clearer now that we have multiple kinds of arrays/slices.

Fixes #206


japaric (approved):
Nice job, thank you!

PR #215, opened 2020-10-23
don't try to parse `_defmt_version_` as JSON

japaric:
when linking with GNU LD `_defmt_version_` ends in the `.defmt` section (doesn't happen with LLD) so `elf2table` tries to parse that symbol as JSON; this results in this `probe-run` error: "Error: expected value at line 1 column 1".
This PR fixes that issue.

### DIFF
--- a/elf2table/src/lib.rs
+++ b/elf2table/src/lib.rs
@@ -77,7 +77,9 @@ pub fn parse(elf: &[u8]) -> Result<Option<Table>, anyhow::Error> {
         };
 
         if is_defmt_version(name) {
-            // not a defmt symbol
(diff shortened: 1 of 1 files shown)


elf2table/src/lib.rs
```diff
@@ -73,6 +76,11 @@ pub fn parse(elf: &[u8]) -> Result<Option<Table>, anyhow::Error> {
             _ => continue,
         };
 
+        if is_defmt_version(name) {
+            // not a defmt symbol
```
jonas-schievink, line 80: Can you copy (part of) the PR description here to explain why this is necessary?

jonas-schievink (approved):

PR #216, opened 2020-10-23
explicitly place the .defmt section at address 0

japaric:
before this PR LLD was placing `.defmt` at address 0 but GNU LD was placing the section at some arbitrary address
with this PR both linkers set the start address of `.defmt` to zero, which is what we want (we rely on the indices / addresses being small, etc.)
with this change linking with GNU LD gives me log messages (before it I was getting nothing from apps linked with GNU LD)


jonas-schievink (approved):
*sigh* linkers

PR #218, opened 2020-10-27
Hide all Formatter methods

### LINKED ISSUE #217: Hide the remaining `defmt::Formatter` API
Some methods aren't `#[doc(hidden)]`, we should try putting `#[doc(hidden)]` on the impl block instead of each method.

jonas-schievink:
Just a `#[doc(hidden)]` on the impl takes care of it.

Closes #217


Lotterleben:
neat!

Lotterleben (approved):

PR #219, opened 2020-10-29
CI improvements

jonas-schievink:
We slipped in a warning, let's make this fail CI.

### DIFF
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -32,10 +32,6 @@ jobs:
         profile: minimal
         toolchain: ${{ matrix.rust }}
         override: true
-    - name: Setup mdBook
-      uses: peaceiris/actions-mdbook@v1
-      with:
-        mdbook-version: latest
     - name: Install C libraries for tooling on ubuntu
       if: matrix.os == 'ubuntu-latest'
       run: sudo apt-get update && sudo apt-get install libudev-dev libusb-1.0-0-dev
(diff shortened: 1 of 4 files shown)


.github/workflows/ci.yml
```diff
@@ -69,12 +61,45 @@ jobs:
...
       run: RUSTFLAGS='--deny warnings' cargo build --verbose --target ${{ env.NO_STD_TARGET }} --bins
 
+  mdbook:
+    strategy:
+      matrix:
+        rust:
+          - stable
+        os:
+          - ubuntu-latest
+          - windows-latest
```
Lotterleben, line 82: out of curiosity/wild guessing: are we building for ubuntu + windows bc we're only expecting additional hickups from non-unixoid systems?

firmware/qemu/src/bin/log.rs
```diff
@@ -339,28 +339,39 @@ fn main() -> ! {
...
     defmt::info!("empty tuple: {:?}", ());
-    defmt::info!("tuple of ints: {:?}", (1,2,3));
-    defmt::info!("nested tuple of ints: {:?}", (1,2,(3,4,5), (6,7,8)));
-    defmt::info!("crazy nested tuples: {:?}", ((((((((),),),),),),), (((((((), (),),),),),),),));
-    defmt::info!("slice of tuples: {:?}", &[(1,2), (3,4), (5,6)][..]);
-    defmt::info!("tuple of slices: {:?}", (&[1,2,3][..], &[4,5,6][..]));
+    defmt::info!("tuple of ints: {:?}", (1, 2, 3));
+    defmt::info!("nested tuple of ints: {:?}", (1, 2, (3, 4, 5), (6, 7, 8)));
+    defmt::info!(
+        "crazy nested tuples: {:?}",
```
Lotterleben, line 345: while we're at it nitpick for language; I think some users might be uncomfortable with the choice of words here
```suggestion
        "super nested tuples: {:?}",
```

Lotterleben (approved):

PR #222, opened 2020-10-29
Add snapshot testing for the QEMU test

### LINKED ISSUE #214: snapshot test the output of QEMU test program
we should be testing the *output* of `firmware/qemu/src/bin/log.rs` to ensure the output doesn't change unintentionally (that could indicate a bug in the decoder for example).

snapshot testing can be used for this: you commit the expected output as a file and on each change you check the output of the `log` program against that file. If the output doesn't match the file that errors the CI. Differences need to be reviewed as they sometimes may be intended, e.g. a log statement was added to `log.rs`.

jonas-schievink:
Closes #214


jonas-schievink:
nightly+alloc tests are broken, I expected that to fail. Will fix later.

Lotterleben (approved):
ooooh this is nice. ✨

jonas-schievink:
```
     Running `/home/runner/work/defmt/defmt/target/debug/qemu-run /home/runner/work/defmt/defmt/target/thumbv7m-none-eabi/debug/log`
Error: defmt version symbol not found
```

Very weird. This doesn't happen locally.

jonas-schievink:
:tada:

PR #226, opened 2020-11-03
Fix encoding of single-variant enums

### LINKED ISSUE #225: Corrupted stream with enum with 1 case
```rust
#[derive(defmt::Format)]
pub enum Payload {
    Number(u32),
}

let payload = Payload::Number(31337);
info!("print of death: {:?}", payload);
```

I guess it happens because the format string has no `|` char, so decoding doesn't expect the discriminator...?

jonas-schievink:
Fixes #225


firmware/qemu/src/bin/log.release.out
```diff
@@ -81,3 +81,5 @@
 0.000080 INFO [[u32;4];4]: [[1, 2, 3, 4], [2, 3, 4, 5], [3, 4, 5, 6], [4, 5, 6, 7]]
 0.000081 INFO [Option<u32>;4]: [Some(1), None, Some(3), None]
 0.000082 INFO [(u32,u32);4]: [(1, 2), (3, 4), (5, 6), (7, 8)]
+0.000083 INFO 1-variant enum: A { fld: 123 }
+0.000084 INFO wrapped: A(A { fld: 200 })
```
Lotterleben, line 85: nitpick: *opens can of worms* could you add a test to `tests/encode.rs` for this as well? I was super excited about your snapshot testing PR because it's nifty and helpful but right now I'm a bit worried that this may lead to diverging tests and no "single source of truth"

jonas-schievink: Turns out we already had tests for this, which subsequently broke. I've pushed da5ec41 to update them.

Lotterleben (approved):

Issue #208, opened 2020-10-22
corrupted stream when mixing bools, options and arrays


Dirbaio:
```rust
    #[derive(defmt::Format)]
    pub struct DhcpRepr {
        pub broadcast: bool,
        pub a2: Option<[u8;6]>,
        pub a3: Option<[u8;6]>,
    }

    let mut dhcp_repr = DhcpRepr {
        broadcast: true,
        a2: Some([1, 2, 3, 4, 5, 6]),
        a3: None,
    };

    defmt::info!("print of death: {:?}", dhcp_repr);

```

if you stick it in the qemu firmware, it outputs this:

```
Error: couldn't decode all data (remaining: [70, 49, 6f, d, 1, b, 1, 2, 3, 4, 5, 6, d, 0, 1])
```

reproduces on latest `main` branch

jonas-schievink:
I'm seeing this instead:

    0.000087 INFO print of death: DhcpRepr { broadcast: false, a2: Some([1, 3, 4, 5, 6, 15]), a3: 1 }

(which is arguably worse :smile: )

jonas-schievink:
The derived `Format` impl contains a `_fmt_.finalize();` call, which is wrong. `finalize` should only be called by the logging macros once the whole frame has been encoded, otherwise we end up with additional compressed bool fields in the message.

jonas-schievink:
Hmm, that does not seem to be the cause of this issue though. The same bug is in the decoder, so things end up working out fine.

jonas-schievink:
Ah, the `Format` impl of `bool` also had a bug. Opened https://github.com/knurling-rs/defmt/pull/228 to fix this.

PR #228, opened 2020-11-04
Derive and Decoder fixes around bools

jonas-schievink:
Fixes https://github.com/knurling-rs/defmt/issues/208

This makes our bool compression scheme match the documentation in the book – previously we've flushed the compression after every `Format` impl, which wastes bandwidth.

### DIFF
--- a/src/impls.rs
+++ b/src/impls.rs
@@ -171,7 +171,7 @@ impl Format for bool {
             let t = internp!("{:bool}");
             fmt.u8(&t);
         }
-        fmt.u8(&(*self as u8));
+        fmt.bool(self);
     }
 }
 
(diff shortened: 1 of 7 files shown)


Lotterleben (approved):
I've read the code, seemed plausible and haven't seen anything I'd find suspicious but I'm trusting your tests & you here; for a more in-depth review you'd probably have to wait for Jorge to return

jonas-schievink:
Should be fine to merge this now. If something does come up I can always open a follow-up PR.

src/impls.rs
```diff
@@ -171,7 +171,7 @@ impl Format for bool {
             let t = internp!("{:bool}");
             fmt.u8(&t);
         }
-        fmt.u8(&(*self as u8));
```
japaric, line 174: that looks like the pre-bool-compression Format implementation. Whoops

PR #230, opened 2020-11-09
Handle crates.io versions in decoder build script

jonas-schievink:
Looks like this was missed when the defmt build script was changed


japaric (approved):
Nice catch. (Feel free to merge once CI is green)

Issue #213, opened 2020-10-23
review docs


japaric:
before the crates.io release we should review the documentation to check that all features that have landed recently are documented there. (I did a quick pass in #212)
I'd say the API docs are more important to get right since they'll end on docs.rs and updating that requires a new crates.io release.
The mdbook (https://defmt.ferrous-systems.com) we can easily update at any time.

japaric:
crates.io release has shipped :rocket:

PR #232, opened 2020-11-09
Improve defmt rustdoc

jonas-schievink:
cc #213


src/lib.rs
```diff
@@ -1,13 +1,14 @@
-//! `defmt`
-//!
-//! > **PRE-ALPHA PREVIEW** `defmt` wire format has not been finalized yet. When using the
```
japaric, line 3: (the book also has this "alpha" warning)

japaric (approved):
:+1:

PR #233, opened 2020-11-10
remove old panic-probe copy

japaric:


japaric:
I think the build error is due to the single workspace configuration. The `panic-probe` cratre is being compiled for x86 but that crate has a `compile_error!` to prevent compilation to that target.

jonas-schievink:
I think that's why the other crate had `#![cfg(target_os = "none")]`, seems like we might want to add that to the public panic-probe crate too

japaric:
CI build failure:
```
error[E0152]: found duplicate lang item `panic_impl`
   --> firmware/qemu/src/bin/log.rs:450:1
    |
450 | fn panic(_: &core::panic::PanicInfo) -> ! {
    | ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
    |
    = note: the lang item is first defined in crate `std` (which `stable_deref_trait` depends on)
    = note: first definition in `std` loaded from /usr/share/rust/.rustup/toolchains/stable-x86_64-unknown-linux-gnu/lib/rustlib/x86_64-unknown-linux-gnu/lib/libstd-f14aca24435a5414.rlib
    = note: second definition in the local crate (`log`)
```

more single workspace issues? the `log` app is being `cargo-check`-ed on the host and because in that dependency graph the `stable_deref_trait` has the `std` feature (unsure if that's related to the build-dependencies Cargo bug) enabled the above error is produced

jonas-schievink:
Yeah I ran into the same issue when I tried this. Not quite sure what's wrong.

japaric:
CI is green. r? @jonas-schievink

firmware/nrf52/Cargo.toml
```diff
@@ -15,7 +15,7 @@ cortex-m-rt = "0.6.10"
 cortex-m-semihosting = "0.3.3"
 embedded-hal = "0.2.4"
 nrf52840-hal = "0.10.0"
-panic-probe = { path = "../panic-probe" }
+panic-probe = { git = "https://github.com/knurling-rs/probe-run", branch = "main" }
```
jonas-schievink, line 18: Should we pin this / publish a new panic-probe version first?

japaric: hmm, yeah. I think we should pin to a commit and then move this to a crates.io version.

japaric: (we don't actually run these apps so we won't hit panics but it's best to pin the dependency)

jonas-schievink: Yeah, I was thinking of deleting the nrf52-specific stuff for that reason. But it doesn't hurt to keep it I guess. I think we do build them in CI though.

jonas-schievink (approved):

PR #234, opened 2020-11-10
touch up crate level docs

japaric:
note that some crates are implementation details atm and not meant to be directly used. Also link to
the defmt repo

cc #213


firmware/defmt-rtt/src/lib.rs
```diff
@@ -1,6 +1,13 @@
-//! `defmt` global logger over RTT
+//! [`defmt`](https://github.com/knurling-rs/defmt) global logger over RTT
```
jonas-schievink, line 1:
```suggestion
//! [`defmt`](https://github.com/knurling-rs/defmt) global logger over RTT.
```

jonas-schievink (approved):

PR #239, opened 2020-11-11
add crate metadata & crates.io dependencies

### LINKED ISSUE #236: add `version = ..` to dependencies
to be able to publish defmt on crates.io all the crates in this repo need to depend on their crates.io version. That means that dependencies that look like in the Cargo.toml like this:

``` toml
[dependencies]
defmt-macros = { path = "macros" }
```

should look like this:

``` toml
[dependencies]
defmt-macros = { path = "macros", version = "0.1.0" } # add version field
```

the specified version needs to match the current version of the crate. As everything will be release for the first, "0.1.0" should be used everywhere

japaric:
also update defmt's README and create a README for defmt-rtt

closes #236

### DIFF
--- a/firmware/defmt-rtt/README.md
+++ b/firmware/defmt-rtt/README.md
@@ -1,4 +1,8 @@
-# `defmt`
+# `defmt-rtt`
+
+> Transmit [`defmt`] log messages over the RTT (Real-Time Transfer) protocol
+
+[`defmt`]: https://github.com/knurling-rs/defmt
 
 `defmt` ("de format", short for "deferred formatting") is a highly efficient logging framework that targets resource-constrained devices, like microcontrollers.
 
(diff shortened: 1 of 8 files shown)


firmware/defmt-rtt/README.md
```diff
@@ -0,0 +1,42 @@
...
+
+[`defmt`]: https://github.com/knurling-rs/defmt
+
+`defmt` ("de format", short for "deferred formatting") is a highly efficient logging framework that targets resource-constrained devices, like microcontrollers.
+
+The fastest way to get started with `defmt` is to use our [app-template] to set up a new Cortex-M embedded project.
+
+[app-template]: https://github.com/knurling-rs/app-template
+
+For more details about the framework check the book at https://defmt.ferrous-systems.com
```
Mirabellensaft, line 13:
```suggestion
For more details about the framework check the book at https://defmt.ferrous-systems.com.
```

Mirabellensaft (approved):

firmware/defmt-rtt/README.md
japaric: adding a period there may break the URL (not sure myself) so I'll add this with a space after the URL

PR #241, opened 2020-11-11
decoder: make DEFMT_VERSION public

japaric:
so we can print it from `probe-run`


japaric:
P.S. this is all unstable API

Mirabellensaft (approved):

PR #242, opened 2020-11-12
improve defmt version mismatch diagnostics

japaric:
offer a different suggestion based on whether a git or crates.io version is being used

should be merged after #244

### DIFF
--- a/decoder/src/lib.rs
+++ b/decoder/src/lib.rs
@@ -98,13 +98,69 @@ pub struct Table {
 /// Checks if the version encoded in the symbol table is compatible with this version of the
 /// `decoder` crate
 pub fn check_version(version: &str) -> Result<(), String> {
+    enum Kind {
+        // "1" or "0.1"
+        Semver,
+        // commit hash "e739d0ac703dfa629a159be329e8c62a1c3ed206"
+        Git,
+    }
+
+    impl Kind {
+        fn of(version: &str) -> Kind {
+            if version.contains('.') {
+                // "0.1"
+                Kind::Semver
+            } else if version.parse::<u64>().is_ok() {
+                // "1"
+                Kind::Semver
+            } else {
+                // "e739d0ac703dfa629a159be329e8c62a1c3ed206" (should be)
+                Kind::Git
+            }
+        }
+    }
+
     if version != DEFMT_VERSION {
-        return Err(format!(
-            "defmt version mismatch (firmware is using {}, host is using {}); \
-             are you using the same git version of defmt and related tools?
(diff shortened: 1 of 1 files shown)


decoder/src/lib.rs
```diff
@@ -98,13 +98,69 @@ pub struct Table {
...
+            (Kind::Git, Kind::Git) => {
+                write!(
+                    msg,
+                    "pin _all_ `defmt` related dependencies to revision {0}; modify Cargo.toml files as shown below
+
+[dependencies]
+defmt = {{ git = \"https://github.com/knurling-rs/defmt\", rev = \"{0}\" }}
+defmt-rtt = {{ git = \"https://github.com/knurling-rs/defmt\", rev = \"{0}\" }}
+# ONLY pin this dependency if you are using the `print-defmt` feature
+panic-probe = {{ git = \"https://github.com/knurling-rs/defmt\", features = [\"print-defmt\"], rev = \"{0}\" }}",
```
japaric, line 139: `panic-probe` currently lives in the `probe-run` repo so it's not really possible to pin to the correct defmt rev (`panic-probe` also already pins to a particular `defmt` and that cannot be changed from an application project). I think that to simplify the use of the git-version of defmt we'll have to move panic-probe (back) into the defmt repo. This PR is blocked on that migration.

japaric: migration PR: #244

decoder/src/lib.rs
```diff
@@ -98,13 +98,69 @@ pub struct Table {
...
 /// `decoder` crate
 pub fn check_version(version: &str) -> Result<(), String> {
+    enum Kind {
+        // "1" or "0.1"
+        Semver,
+        // commit hash "e739d0ac703dfa629a159be329e8c62a1c3ed206"
+        Git,
+    }
+
+    impl Kind {
```
Lotterleben, line 108: question: why are these defined within the function instead of outside?

japaric: the `enum` is only used within the function body so putting it inside avoids adding yet another type to the outer (`src/lib.rs`) type namespace.

Another way to keep the namespace tidy is use modules so this could have been written as: `mod version { enum Kind { .. } pub fn check(version: &str) -> Result<(), String> { .. } }`. We have not used modules in this crate at all so the file is over 1.6K lines ... we should eventually come back and refactor this into modules.

Lotterleben: alrighty, thank you for the clarification! +1 to the refactoring, but that's for another day ofc

japaric:
reminder to self: publish a new patch version of `defmt-decoder` after this is merged

Lotterleben (approved):

PR #244, opened 2020-11-12
Import panic-probe crate

japaric:
from the probe-run repo, including its git history
also make it depend on defmt via a path-dependency (instead of a git-dependency)
this makes it easier to use a git version of defmt in e.g. an app-template project


japaric:
the only real change is the last commit ("path-depend on defmt"); the rest is the existing probe-run repo git history

Lotterleben (approved):

PR #248, opened 2020-11-13
Import defmt-test crate

japaric:
from the defmt-test repo, including its git history
also make it depend on defmt via a path-dependency (instead of a git-dependency)
also add missing crate metadata so we can publish this on crates.io
this makes it easier to use a git version of defmt in e.g. an app-template project

Commits before "Merge external defmt-test git history" are existing commits from the defmt-test repo; commits after it are new commits that need reviewing

### DIFF
--- a/firmware/defmt-test/src/lib.rs
+++ b/firmware/defmt-test/src/lib.rs
@@ -1,10 +1,10 @@
-//! A test harness for embedded devices
+//! A test harness for embedded devices.
 //!
 //! This crate has a single API: the `#[tests]` macro. This macro is documented in the project
 //! README which can be found at:
 //!
-//! - https://crates.io/crates/defmt (crates.io version)
-//! - https://github.com/knurling-rs/defmt/tree/main/firmware/defmt-test (git version)
+//! - <https://crates.io/crates/defmt-test> (crates.io version)
+//! - <https://github.com/knurling-rs/defmt/tree/main/firmware/defmt-test> (git version)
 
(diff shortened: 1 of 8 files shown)


Mirabellensaft (approved):

firmware/defmt-test/src/lib.rs
```diff
@@ -0,0 +1,13 @@
+//! A test harness for embedded devices
```
jonas-schievink, line 1:
```suggestion
//! A test harness for embedded devices.
```

firmware/defmt-test/src/lib.rs
```diff
@@ -0,0 +1,13 @@
...
+//!
+//! This crate has a single API: the `#[tests]` macro. This macro is documented in the project
+//! README which can be found at:
+//!
+//! - https://crates.io/crates/defmt (crates.io version)
```
jonas-schievink, line 6:
```suggestion
//! - <https://crates.io/crates/defmt-test> (crates.io version)
```

firmware/defmt-test/src/lib.rs
```diff
@@ -0,0 +1,13 @@
...
+//! - https://github.com/knurling-rs/defmt/tree/main/firmware/defmt-test (git version)
```
jonas-schievink, line 7:
```suggestion
//! - <https://github.com/knurling-rs/defmt/tree/main/firmware/defmt-test> (git version)
```

Issue #145, opened 2020-08-27
Hex formating is missing {:x}


andresv:
I saw that `{:x}` was announced in https://ferrous-systems.com/blog/defmt/. However it seems to be actually missing.
I couldn't find it from here: https://github.com/knurling-rs/defmt/blob/main/parser/src/lib.rs.

japaric:
> I saw that {:x} was announced in https://ferrous-systems.com/blog/defmt/. 

the `{:x}` there is making reference to `core::fmt::Hex`. `defmt` has a single "formatting" trait (`Format`) so `{:x}`, `{:b}`, etc. are not planned. 

The rationale for a single trait is that since formatting is done on the host, it should be the host side who decides how to format the string and not the device. Interactive web / GUI front ends can provide a mechanism (e.g. click the number) to change how the information is presented; that way you don't have to re-build, re-flash and re-run the program to change the representation. That however is harder to do with the current non-interactive terminal output so we have considered printing different representations at once (e.g. `{16,0x10,0b10000}` instead of `16`) but that "bloats" the output.

andresv:
Got it.

AndWass:
> the {:x} there is making reference to core::fmt::Hex. defmt has a single "formatting" trait (Format) so {:x}, {:b}, etc. are not planned.
> The rationale for a single trait is that since formatting is done on the host, it should be the host side who decides how to format the string and not the device. Interactive web / GUI front ends can provide a mechanism (e.g. click the number) to change how the information is presented; that way you don't have to re-build, re-flash and re-run the program to change the representation. That however is harder to do with the current non-interactive terminal output so we have considered printing different representations at once (e.g. {16,0x10,0b10000} instead of 16) but that "bloats" the output.

But allowing the target side to provide formatting hints (via `{:x}` and `{:b}`) doesn't stop interactive web / GUI frontends from providing the functionality to change how information is presented anyway. And isn't it the case that the person who most often will look at the output is also the person who will be writing the code for the target? So for that person it does provide immediate QoL-improvements for the current non-interactive terminal output, especially since the need for currently non-existing tools diminishes.

Dirbaio:
+1000 to this. I've been using `defmt` in my projects since it came out, and not being able to print stuff in hex is the only big usability issue I've ran into. And it's quite a big one IMO. Hex is especially important for embedded dev. 

[...]

PR #253, opened 2020-11-14
#145: Implement display hints to optionally format values as hex

twitchyliquid64:
This PR implements the beginnings of formatting values as hex, as per design in https://github.com/knurling-rs/defmt/issues/145#issuecomment-712235944.

This allows you to do:

```rust
defmt::error!(
    " double-fault at 0x{:u32:X} !",
    fault_addr,
);
```

There's plenty more options to support, but my hope with this PR is to start the conversation on:

1. Where the logic should go and how it should be represented
2. What we should call it (format? display hints?)
3. What we should support (padding? what formats? just integer types or everything?)

To start, this PR just implements the `{<optional-arg-idx>:<type>:<display-hint>}` form, where (currently) the only two display hints are `:x` (lower hex) and `:X` (upper hex).

### DIFF
--- a/decoder/src/lib.rs
+++ b/decoder/src/lib.rs
@@ -24,7 +24,7 @@ use byteorder::{ReadBytesExt, LE};
 use colored::Colorize;
 
 pub use defmt_parser::Level;
-use defmt_parser::{get_max_bitfield_range, Fragment, Parameter, Type};
+use defmt_parser::{get_max_bitfield_range, DisplayHint, Fragment, Parameter, Type};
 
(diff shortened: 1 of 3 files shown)


japaric:
Thanks for the PR. The design for this 'display hints' feature is currently being discussed in RFC #270. Until the design of the feature is finalized and approved I'm going to close PRs related to the feature -- without an approved design those PRs cannot be reviewed.

If you would like to share your implementation of RFC #270 (note: the design is still in flux so you may need to update your implementation several times) with others you can post the link to your branch in thread #270 and I'll link it from the issue description.

PR #259, opened 2020-11-16
make crates.io version not depend on the `git` tool

### LINKED ISSUE #256: Using defmt on WSL provides a cryptic error message if git is missing
While `libudev` and `libusb` are mentioned as dependencies, `git` is not. As such when trying to install `probe-run` or use defmt in a project you can get:
```
error: failed to run custom build command for defmt-decoder v0.1.1
Error: Os { code: 2, kind: NotFound, message: "No such file or directory" }
```
This was hard to work out until `git` was mentioned in chat.
Once git was installed, it worked fine.

japaric:
also error if git version is used but `git` is not installed
prompt the user to install `git` in that case

fixes #256

---

we need `git` when the git version is used because we use the commit hash to check if probe-run is compatible with the firmware-side defmt
instead of panicking when git is not installed we could:
- depend on the libgit2 crate and use that to retrieve the commit hash (from what I remember libgit2 is as, or more, annoying to install as git on Windows (not WSL))
- implement `git rev-parse HEAD` using just the std library. It seems reading the `.git/HEAD` file and then reading the file specified there (e.g. `.git/refs/heads/main`, which contains the full hash) should do the trick but it does feel a bit brittle

### DIFF
--- a/build.rs
+++ b/build.rs
@@ -8,18 +8,22 @@ fn main() -> Result<(), Box<dyn Error>> {
     let hash = Command::new("git")
         .args(&["rev-parse", "HEAD"])
         .output()
-        .map(|output| {
+        .ok()
+        .and_then(|output| {
             if output.status.success() {
                 String::from_utf8(output.stdout).ok()
             } else {
-                assert!(!Path::new(".git").exists(), "you need to install the `git` command line tool to use the git version of `defmt`");
-
                 None
             }
         });
-    let version = if let Ok(Some(hash)) = hash {
+    let version = if let Some(hash) = hash {
         hash
     } else {
+        assert!(
+            !Path::new(".git").exists(),
+            "you need to install the `git` command line tool to use the git version of `defmt`"
+        );
+
         // no git info -> assume crates.io
         let semver = Version::parse(&std::env::var("CARGO_PKG_VERSION")?)?;
         if semver.major == 0 {
--- a/build.rs
+++ b/build.rs
@@ -1,13 +1,24 @@
 use semver::Version;
-use std::{env, error::Error, fs, path::PathBuf, process::Command};
+use std::{env, error::Error, fs, path::Path, path::PathBuf, process::Command};
 
 fn main() -> Result<(), Box<dyn Error>> {
     // Put the linker script somewhere the linker can find it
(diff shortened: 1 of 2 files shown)


build.rs
```diff
@@ -1,13 +1,24 @@
...
 fn main() -> Result<(), Box<dyn Error>> {
     // Put the linker script somewhere the linker can find it
     let out = &PathBuf::from(env::var("OUT_DIR")?);
     let mut linker_script = fs::read_to_string("defmt.x.in")?;
-    let output = Command::new("git").args(&["rev-parse", "HEAD"]).output()?;
-    let version = if output.status.success() {
-        String::from_utf8(output.stdout).unwrap()
+    let hash = Command::new("git")
+        .args(&["rev-parse", "HEAD"])
+        .output()
```
jonas-schievink, line 10: The `output()` call itself will fail if the program cannot be found, so I don't think the fix can use `.map`

japaric: the original was not checking the .git path if git was not installed but the new version should

decoder/build.rs
```diff
@@ -1,13 +1,31 @@
...
+            if output.status.success() {
+                String::from_utf8(output.stdout).ok()
+            } else {
+                None
+            }
+        });
+    let version = if let Some(hash) = hash {
+        hash
     } else {
+        assert!(!Path::new(".git").exists(), "you need to install the `git` command line tool to install the git version of `probe-run`");
```
jonas-schievink, line 27: Did you test that this is the right path? I'm not sure in which working directory build scripts are executed. If it's the Cargo workspace root this seems correct, but it seems like it might be the *package* root, which means that this would be wrong for the decoder.

japaric: yes, there's an old commit that removes a `panic!("{:?}, env::current_dir())` statement from the build script. I used to inspect the directory from which the build script was invoked. It was the package root and not the workspace root.

I guess I could also use `CARGO_MANIFEST_DIR` (was that the path to the directory that contains the `Cargo.toml` file?) instead of relying on `current_dir`. Perhaps that'd be more reliable?

jonas-schievink: Either way seems fine to me

jonas-schievink (approved):

PR #261, opened 2020-11-16
defmt-test: update README

japaric:
so example output matches current probe-run output
also steps 2 and 3 won't apply to the newest version of app-template
also recommend crates.io version over git in step 1

### DIFF
--- a/firmware/defmt-test/README.md
+++ b/firmware/defmt-test/README.md
@@ -4,7 +4,7 @@
 
 ## Basic usage
 
-0. If you don't have a project setup yet, start from the [`app-template`] then proceed with step 3. If you have a project version on an old version of `app-template` then proceed with step 1.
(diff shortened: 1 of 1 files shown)


firmware/defmt-test/README.md
```diff
@@ -4,18 +4,17 @@
 
 ## Basic usage
 
-0. If you don't have a project setup yet, start from the [`app-template`] 
-
+0. If you don't have a project setup yet, start from the [`app-template`] then proceed with step 3. If you have a project version on an old version of `app-template` then proceed with step 1.
```
jonas-schievink, line 7:
```suggestion
0. If you don't have a project setup yet, start from the [`app-template`], then proceed with step 3. If you have a project on an old version of `app-template` then proceed with step 1.
```

jonas-schievink (approved):
Looks good, modulo nits

Issue #97, opened 2020-08-10
align `winfo!()` with `info!()`


Lotterleben:
currently, the implementation of `winfo!()`  is prefixed with
```
// TODO share more code with `log`
```

It seems to be used in unit tests for encoding only (i.e. in encode.rs) so it should be reusing the code running for `info!()`– otherwise we can't really trust the tests to break (i.e. if we change info! but forget to update winfo! as well)

PR #263, opened 2020-11-16
`write!` macro

### LINKED ISSUE #104: syntax checked `write!` macro
for now the `write!` macro has been removed from the public API because it's too free form.
We'll need the `write!` macro for things like `svd2rust` integration (`derive` is not flexible enough) so we want a public `write!` macro but that's syntax checked: the format string should be a valid struct ("S {{ field: {:u8} }} }") or valid enum.

We can start with just checked struct syntax. enum syntax is a bit harder: how do we want to expose it to the user so that they cannot misuse?

jonas-schievink:
Closes #104
Closes https://github.com/knurling-rs/defmt/issues/97 (`winfo!` is removed)


jonas-schievink:
This now implements a more limited, but backwards-compatible version where you can only call `write!` once per `Format` impl. In 0.2, we'll allow arbitrarily many calls and introduce a dedicated tag for `write!`, which lets us distinguish it from derived `Format` impls.

firmware/qemu/src/bin/double-write.rs
```diff
@@ -0,0 +1,35 @@
...
+use defmt_semihosting as _; // global logger
+
+use cortex_m_semihosting::debug;
+
+struct MyStruct;
+
+impl Format for MyStruct {
+    fn format(&self, fmt: &mut defmt::Formatter) {
+        defmt::write!(fmt, "one");
+        defmt::write!(fmt, "two");
```
japaric, line 16: :+1:

japaric (approved):
this is excellent work.

before publishing defmt 0.1.2 we should mention this feature in the book but that can be done in a follow-up PR so I'll hit the merge button now

Issue #252, opened 2020-11-13
Ensure that attribute macros retain or reject other attributes


jonas-schievink:
Currently, it isn't clear what happens when defmt's attribute macros are combined with built-in attributes. We should make sure that this works and add tests, and reject any incompatible attributes.

japaric:
done in #293

PR #266, opened 2020-11-17
add defmt::panic!

### LINKED ISSUE #136: `defmt::{panic, assert, assert_eq}`
It would be great if defmt could be used as an efficient replacement for `panic!`, `assert!`, and `assert_eq!`.

These could work like this: If the panic should happen, a `defmt::error!` is logged with the formatting arguments, and then a `core::panic!();` is raised. We might need to think about the panic payload, since a bare `core::panic!()` is the same as `core::panic!("explicit panic");`, which pulls in (or *might* pull in) `core::fmt`.

Also note that we can't use `stringify!` to turn the assert expression into text, since the text has to go into the symbol name. That means that `assert!`-style macros all have to be procedural macros and stringify part of their arguments by themselves.

This won't get rid of *all* inefficient panics. For example, indexing something out of bounds still uses `core::fmt` to format the panic string (and of course all dependencies that don't use defmt will also do that). But it should still help.

japaric:
closes #136
this implements `defmt::panic!` which calls `defmt::error!` and then "panics"

by default, `defmt::panic!` will "panic" by calling `core::panic!`. This preserves the panicking *behavior* (e.g. abort on panic, reset on panic, etc.) but can result in the panic message being printed *twice* -- this is the case if you use `panic_probe` (+print-defmt)

to prevent printing the panic message *twice*, the panicking *behavior* of `defmt::panic` can be overridden using the `#[defmt::panic_handler]` attribute, which works about the same as `#[core::panic_handler]` but expects a `fn() -> !` function (note: no argument)

the idea is that end users will replicate their `#[panic_handler]` panicking *behavior*, and omit the 'print the panic message' part, in the `#[defmt::panic_handler]` function, for example:

``` rust
#[panic_handler]
fn panic(info: &PanicInfo) -> ! {
    print(info); // e.g. using defmt
    reset()
}

#[defmt::panic_handler]
fn defmt_panic() -> ! {
    reset()
}
```

we can do the above for all new users of app-template so they don't run into the 'panic message printed twice':

``` rust
// src/lib.rs
use panic_probe as _; // "print-defmt" is enabled

#[defmt::panic_handler]
fn panic() -> ! {
    cortex_m::asm::udf()
}
```

TODO
- [x] API docs
- [x] book docs
- [x] todo! & unimplemented!
- [x] unreachable!
- [x] assert!
- [x] assert_eq! assert_ne!
- [x] debug_assert, debug_assert_eq, debug_assert_ne
- [x] snapshot tests
- [x] assert_eq! & probe-run integration (colored diff) -> https://github.com/knurling-rs/probe-run/tree/color-diff
- ~~prelude~~ didn't work you can't override core's prelude items using a glob import

FIXME
- [x] probe-run renders `assert_ne` as a color-diff but shouldn't

### DIFF
--- a/book/src/panic.md
+++ b/book/src/panic.md
@@ -1,0 +1,38 @@
+# `panic!` and `assert!`
+
+The `defmt` crate provides its own version of `panic!`-like and `assert!`-like macros.
+The `defmt` version of these macros will log the panic message using `defmt` and then call `core::panic!` (by default).
+Because the panic message is formatted using `defmt!` the format string must use the same syntax as the logging macros (e.g. `info!`).
+
+## `#[defmt::panic_handler]`
+
+Because `defmt::panic!` invokes `core::panic!` this can result in the panic message being printed twice if your `#[core::panic_handler]` also prints the panic message.
+This is the case if you use [`panic-probe`] with the `print-defmt` feature enabled but not an issue if you are using the [`panic-abort`] crate, for example.
+
+[`panic-probe`]: https://crates.io/crates/panic-probe
+[`panic-abort`]: https://crates.io/crates/panic-abort
+
+To avoid this issue you can use the `#[defmt::panic_handler]` to *override* the panicking behavior of `defmt::panic`-like and `defmt::assert`-like macros.
+This attribute must be placed on a function with signature `fn() -> !`.
+In this function you'll want to replicate the panicking behavior of `#[core::panic_handler]` but leave out the part that prints the panic message.
+For example:
+
+<!-- NOTE(ignore) we can't compile this test because the `panic_handler` defined here collides with the one in `std` -->
+
+``` rust, ignore
+#[panic_handler] // built-in ("core") attribute
+fn core_panic(info: &core::panic::PanicInfo) -> ! {
+    print(info); // e.g. using RTT
+    reset()
+}
+
+#[defmt::panic_handler] // defmt's attribute
+fn defmt_panic() -> ! {
+    // leave out the printing part here
+    reset()
+}
+```
+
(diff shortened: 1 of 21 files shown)


book/src/panic.md
```diff
@@ -0,0 +1,38 @@
...
+
+Because `defmt::panic!` invokes `core::panic!` this can result in the panic message being printed twice if your `#[core::panic_handler]` also prints the panic message.
+This is the case if you use [`panic-probe`] with the `print-defmt` feature enabled but not an issue if you are using the [`panic-abort`] crate, for example.
+
+[`panic-probe`]: https://crates.io/crates/panic-probe
+[`panic-abort`]: https://crates.io/crates/panic-abort
+
+To avoid this issue you can use the `#[defmt::panic_handler]` to *override* the panicking behavior of `defmt::panic`-like and `defmt::assert`-like macros.
+This attribute must be placed on a function with signature `fn() -> !`.
+In this function you'll want to replicate the panicking behavior of `#[core::panic_handler]` but leave out the part that prints the panic message.
```
jonas-schievink, line 17:
```suggestion
In this function you'll want to replicate the panicking behavior of the Rust `#[panic_handler]` but leave out the part that prints the panic message.
```

(paths cannot be used to refer to builtin attributes)

macros/src/lib.rs
```diff
@@ -60,6 +59,52 @@ pub fn global_logger(args: TokenStream, input: TokenStream) -> TokenStream {
...
+        || !rety_is_ok
+    {
+        return parse::Error::new(ident.span(), "function must have signature `fn() -> !`")
+            .to_compile_error()
+            .into();
+    }
+
+    let block = &f.block;
+    quote!(
+        #[export_name = "_defmt_panic"]
```
jonas-schievink, line 100: cc https://github.com/knurling-rs/defmt/issues/252

jonas-schievink (approved):
Great!

PR #267, opened 2020-11-20
add Debug2Format adapter

### LINKED ISSUE #126: impl Format for T where T: Debug
I understand the motivation behind the custom `Format` trait. However, some of us have 2MB of flash and want to format library types which only `impl Debug`. It would be nice to provide an opt-in way to do this.

japaric:
this is a temporary solution for #126 
(a better solution is probably something along the lines of https://github.com/knurling-rs/defmt/issues/126#issuecomment-724890594 but that may require changes to the wire format and may *not* make it to the v0.1.x series)
closes #126

### DIFF
--- a/src/lib.rs
+++ b/src/lib.rs
@@ -619,6 +619,8 @@ fn default_panic() -> ! {
 /// `consts` module. Example:
 ///
 /// ```
+/// use defmt::{consts, Debug2Format};
+///
 /// #[derive(Debug)]
 /// struct S { x: i8, y: i16 }
 ///
--- a/src/lib.rs
+++ b/src/lib.rs
@@ -16,7 +16,10 @@
 #[cfg(feature = "alloc")]
 extern crate alloc;
 
-use core::{mem::MaybeUninit, ptr::NonNull};
+use core::{fmt, marker::PhantomData, mem::MaybeUninit, ptr::NonNull};
+
+pub use heapless::consts;
+use heapless::{ArrayLength, String};
 
 #[doc(hidden)]
 pub mod export;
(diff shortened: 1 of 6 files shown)


mattico:
I also had a similar adapter for `Display`, which is perhaps even less in the spirit of defmt since it abandons structured data. I still found it useful.

src/lib.rs
```diff
@@ -461,3 +463,72 @@ pub trait Format {
...
+where
+    N: ArrayLength<u8>,
+{
+    fn format(&self, fmt: &mut Formatter) {
+        use core::fmt::Write as _;
+
+        // `Debug` format into a stack buffer; if the formatted string doesn't fit discard the
+        // excess
+        let mut buf = String::<N>::new();
+        write!(buf, "{:?}", self.value).ok();
```
jonas-schievink, line 513: Could we maybe use an ellipsis here? Maybe put a `…` at the end of the buffer if the full text doesn't fit?

japaric: I can't think of a nice way to do that in a way that ensures the ellipsis will be included if the output is truncated.
Reason being the `write!` may fill `buf` with truncated output. In that scenario we cannot push the ellipsis at the end (because the buffer is full).

> I can't think of a nice way

But I can think of not nice ways to do this. One is passing `String<N - 2>` to `write!` by transmuting a mutable reference (yeah, let's not do that). The other is to use 2 heapless String of sizes N and N + 2: `write!` into one; if that's truncated then memcpy into the larger one + add ellipsis. The last way is to write our own heapless string that has internal capacity N+2 but looks like it has capacity N to the end user; that string would have some method to append an ellipsis.

P.S. The heapless::String's fmt::Write implementation is a bit odd in that it won't write an argument at all if it doesn't fit. That is `write!(s, "x = {}", u32::max_value())` produces "x = " instead of "x = 4294" if the string capacity is 8.

jonas-schievink: Ah, I see, yeah let's just land this as-is then.

jonas-schievink (approved):
Looks good, but needs a rebase

PR #269, opened 2020-11-23
Fix panic in expression context.

### LINKED ISSUE #268: panic! macro should be an expression returning ! (Never type)
Code like this fails to compile, while it compiles with std `panic!()`.

```rust
    let foo: u32 = match version {
        4 => 123,
        5 => 456,
        _ => defmt::panic!("unknown version {:?}", version),
    };
```

```
error: macro expansion ignores token `;` and any following
  --> firmware/qemu/src/bin/panic.rs:14:14
   |
14 |         _ => defmt::panic!("unknown version {:?}", version),
   |              ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^- help: you might be missing a semicolon here: `;`
   |              |
   |              caused by the macro expansion here
   |
   = note: the usage of `defmt::panic!` is likely invalid in expression context

error[E0308]: mismatched types
  --> firmware/qemu/src/bin/panic.rs:14:14
   |
14 |         _ => defmt::panic!("unknown version {:?}", version),
   |              ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^ expected `u32`, found `()`
   |
```

Dirbaio:
Fixes #268.

### DIFF
--- a/firmware/qemu/src/bin/panic.rs
+++ b/firmware/qemu/src/bin/panic.rs
@@ -6,8 +6,14 @@ use cortex_m_rt::entry;
 use defmt_semihosting as _; // global logger
 
 #[entry]
-fn main() -> ! {
-    defmt::panic!("The answer is {:?}", 42);
+fn main() ->  ! {
+    let answer = 42;
+    let foo: u32 = match answer {
+        1 => 123,
+        2 => 456,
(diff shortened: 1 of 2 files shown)


japaric (approved):
ooh, that's subtle. nice catch!

PR #272, opened 2020-11-24
Escape braces in assert expressions.

Dirbaio:
Without this, defmt attempts to interpret the braces as format args, resulting in errors like

```rust
    defmt::assert!({1 + 1} == { 2 });
```

```
error: malformed format string (missing `:`)
  --> firmware/qemu/src/bin/assert.rs:10:5
   |
10 |     defmt::assert!({1 + 1} == { 2 });
   |     ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
   |
```


japaric (approved):
Thanks for the fix!

Issue #161, opened 2020-09-03
unwrap alternative?


Dirbaio:
`uwnrap()` on a Result is quite convenient, but it's not usable with defmt since it requires the error type to be fmt::Debug, which is not useful to print it with defmt.

A solution could be add a trait `defmt::Unwrap` (or maybe `Dewrap`?) and implement it for Result, which does the same as `unwrap()` but printing the result with defmt.

Dirbaio:
Here's a copypaste-ready implementation, in case someone finds it useful...

```rust

pub trait Dewrap<T> {
    /// dewrap = defmt unwrap
    fn dewrap(self) -> T;

    /// dexpect = defmt expect
    fn dexpect<M: defmt::Format>(self, msg: M) -> T;
}

impl<T> Dewrap<T> for Option<T> {
    fn dewrap(self) -> T {
        match self {
            Some(t) => t,
            None => depanic!("unwrap failed: enum is none"),
        }
    }

    fn dexpect<M: defmt::Format>(self, msg: M) -> T {
        match self {
            Some(t) => t,
            None => depanic!("unexpected None: {:?}", msg),
        }
    }
}

impl<T, E: defmt::Format> Dewrap<T> for Result<T, E> {
    fn dewrap(self) -> T {
        match self {
            Ok(t) => t,
            Err(e) => depanic!("unwrap failed: {:?}", e),
        }
    }

    fn dexpect<M: defmt::Format>(self, msg: M) -> T {
        match self {
            Ok(t) => t,
            Err(e) => depanic!("unexpected error: {:?}: {:?}", msg, e),
        }
    }
}
```

kuon:
I started working with this, it works, but it is a bit cumbersome to have both `unwrap()` and `dewrap()` in the code.

It would be really nicer to be able to do `unwrap()`  everywhere and have it being handled properly behind the scene. But I don't know if this is possible.

jonas-schievink:
No, there is no way to override `unwrap()` behavior (besides providing a panic handler, but that will not be able to use defmt).

kuon:
Would it be possible to provide `dewrap()` for `Debug` trait? This would permit having to `dewrap()` everywhere. That's a minor issue, but it can help when dealing with data that mix `defmt::Format` and `Debug`.

jonas-schievink:
I think not without specialization, if we want it to work with both `Debug` and `defmt::Format` types.

japaric:
[...]

PR #273, opened 2020-11-24
Unwrap macro

Dirbaio:
This is a PoC of the `unwrap!` macro discussed in #161, to see how well it works.

Seems to work great, line numbers are printed correctly. The only drawback is the syntax is less nicer than the normal `unwrap()` since it doesn't read left-to-right.

Ignore the `Escape braces in assert expressions.` commit, it's included so the unwrap macro can use the `escape_expr` function.

### DIFF
--- a/macros/src/lib.rs
+++ b/macros/src/lib.rs
@@ -653,20 +653,7 @@ fn assert_binop(ts: TokenStream, binop: BinOp) -> TokenStream {
     };
 
     for val in vals {
-        let mut segments = Punctuated::new();
-        segments.push(PathSegment {
-            ident: Ident2::new(*val, Span2::call_site()),
-            arguments: PathArguments::None,
-        });
-
-        log_args.push(Expr::Path(ExprPath {
-            attrs: vec![],
-            qself: None,
-            path: Path {
-                leading_colon: None,
-                segments,
-            },
-        }));
+        log_args.push(ident_expr(*val));
     }
 
     let log_stmt = match binop {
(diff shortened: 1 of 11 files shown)


japaric:
Thanks for building a PoC.
I tested out locally and I liked that `unwrap!(something)` prints "something" to the console. `core::unwrap` doesn't do that in comparison.
My biggest concern, though, is that the macro "turns off" Rust-Analyzer so if I write a long method chain:
``` rust
unwrap!(x.
    map(|z| z + 1).
    then(something)
)
```
I don't get type annotations next to each method call or in internal closures.

To work around the issue I would split the chain into its own variable:
``` rust
let res = x.
    map(|z| z + 1).
    then(something);
unwrap!(res)
```
But then I won't get the chain expression printed in the panic message (if the unwrap fails) and the reported location will be sort of "off by one line" (but still be more precise than with the `.dewrap()` approach)

japaric:
@Dirbaio the team discussed how we'd like to implement this feature and we are OK with a macro implementation so I'll review this PR once you mark it as ready for review.

Dirbaio:
DOne :D

Dirbaio:
I don't know what's up with the `unused import: crate as defmt` from CI. If I remove it it won't build because it's needed.

macros/src/lib.rs
```diff
@@ -752,6 +739,76 @@ pub fn debug_assert_ne_(ts: TokenStream) -> TokenStream {
...
+            ::core::result::Result::Err(_unwrap_err) => {
+                #log_stmt;
+                defmt::export::panic()
+            }
+        }
+    )
+    .into()
+}
+
+fn ident_expr(name: &str) -> Expr {
```
japaric, line 790: nice refactor :+1:

japaric (approved):
Looks great!

> I don't know what's up with the unused import: crate as defmt from CI. 

I think that's because `internp!` requires it when target = x86 but not when target = ARM. Sticking an `allow` on it would be an OK fix.

PR #274, opened 2020-11-25
Update CI test job on OSX

justahero:
Fixes the CI test task to run all tests on OSX.

Previously the test task did not run the tests of the `defmt-decoder` crate due to the missing feature `unstable`. In order to run these tests as well the `cargo test` command now includes the `--all-features` option to enable all features. For the decoder this enables the `unstable` feature and therefore runs all of its tests.


japaric (approved):
Thanks for the fix!

PR #278, opened 2020-11-25
CI: add backwards compatibility check

### LINKED ISSUE #275: CI: check for backward incompatible changes
the idea: run the snapshot tests with the qemu-run from defmt v0.1.0 
this should be more or less equivalent to running probe-run 0.1.4 (the first probe-run with defmt support) against the latest version of defmt 0.1.x and that setup should always work

japaric:
the check runs the snapshot tests against decoder v0.1.0
the tests must produce the same output even if they are decoder using the oldest v0.1.x decoder
closes #275

### DIFF
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -122,3 +122,8 @@ jobs:
     - name: Build and Run QEMU tests
       working-directory: firmware/qemu
       run: ./test.sh
+    - name: Backward compatibility check against decoder v0.1.0
+      working-directory: firmware/qemu
+      run: |
+        cargo install --debug --git https://github.com/knurling-rs/defmt --branch do-not-delete-v0.1.0-with-no-version-check
+        CARGO_TARGET_THUMBV7M_NONE_EABI_RUNNER=qemu-run ./test.sh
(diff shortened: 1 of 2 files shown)


japaric:
manually checked that this approach catches the breaking change in PR #253 

also enabled bash's pipefail for good measure: the old decoder will exit with non-zero exit code if it can't decode the stream. in that case, pipefail will make the script terminate with non-zero code on the first cargo run invocation that fails *even if the stdout of the program doesn't change* = diff exits successfully (though that scenario seems unlikely given the other test jobs we have)

.github/workflows/ci.yml
```diff
@@ -122,3 +122,8 @@ jobs:
     - name: Build and Run QEMU tests
       working-directory: firmware/qemu
       run: ./test.sh
+    - name: Backward compatibility check against decoder v0.1.0
+      working-directory: firmware/qemu
+      run: |
+        cargo install --debug --git https://github.com/knurling-rs/defmt --branch do-not-delete-v0.1.0-with-no-version-check qemu-run
```
jonas-schievink, line 128: We could also use a tag instead of a branch here, to reduce branch clutter and make the installed version immutable. Up to you though, this also seems fine.

jonas-schievink (approved):
Nice!

.github/workflows/ci.yml
japaric: certainly a tag sounds better! had not thought about that option. will update this

PR #280, opened 2020-11-26
document the unwrap! macro

japaric:
unsure if we should also document the macro in the book -- that may end up being repetitive

### DIFF
--- a/src/lib.rs
+++ b/src/lib.rs
@@ -118,7 +118,33 @@ pub use defmt_macros::todo_ as unimplemented;
 /// [the manual]: https://defmt.ferrous-systems.com/macros.html
 pub use defmt_macros::panic_ as panic;
 
-/// todo
+/// This macro is roughly equivalent to `{Option,Result}::{expect,unwrap}` but invocation looks
+/// a bit different because this is a macro and not a method. The other difference is that
+/// `unwrap!`-ing a `Result<T, E>` value requires that the error type `E` implements the `Format`
+/// trait
+///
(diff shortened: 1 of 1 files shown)


src/lib.rs
```diff
@@ -118,7 +118,33 @@ pub use defmt_macros::todo_ as unimplemented;
 /// [the manual]: https://defmt.ferrous-systems.com/macros.html
 pub use defmt_macros::panic_ as panic;
 
-/// todo
+/// This macro is roughly equivalent to `{Option,Result}::{expect,unwrap}` but invocation looks
```
jonas-schievink, line 121: Maybe add a 1-line description at the top for nicer rustdoc output?

```suggestion
/// Unwraps an `Option` or `Result`, panicking if it is `None` or `Err`.
///
/// This macro is roughly equivalent to `{Option,Result}::{expect,unwrap}` but invocation looks
```

src/lib.rs
```diff
@@ -118,7 +118,33 @@ pub use defmt_macros::todo_ as unimplemented;
...
+/// a bit different because this is a macro and not a method. The other difference is that
+/// `unwrap!`-ing a `Result<T, E>` value requires that the error type `E` implements the `Format`
+/// trait
+///
+/// The following snippet shows the differences between core's unwrap method and defmt's unwrap
+/// macro
```
jonas-schievink, line 127:
```suggestion
/// macro:
```

jonas-schievink (approved):

PR #281, opened 2020-11-26
add Display2Format adapter

japaric:
like #267 but for Display values

### DIFF
--- a/src/lib.rs
+++ b/src/lib.rs
@@ -681,3 +681,75 @@ where
         value,
     }
 }
+
+/// An "adapter" type to feed `Display` values into defmt macros, which expect `defmt::Format` values
+///
+/// This adapter disables compression! You should prefer `defmt::Format` over `Display` whenever
+/// possible
+///
+/// This adapter works by formatting the `Display` value into a stack-allocated buffer. You need to
+/// specify how large that buffer is. If you pick a size that's too small an *incomplete* string
+/// will be transmitted.
+///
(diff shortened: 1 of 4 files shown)


src/lib.rs
```diff
@@ -681,3 +681,75 @@ where
         value,
     }
 }
+
+/// An "adapter" type to feed `Display` values into defmt macros, which expect `defmt::Format` values
```
jonas-schievink, line 685:
```suggestion
/// An "adapter" type to feed `Display` values into defmt macros, which expect `defmt::Format` values.
```

src/lib.rs
```diff
@@ -681,3 +681,75 @@ where
...
+///
+/// This adapter disables compression! You should prefer `defmt::Format` over `Display` whenever
+/// possible
```
jonas-schievink, line 688:
```suggestion
/// possible.
```

src/lib.rs
```diff
@@ -681,3 +681,75 @@ where
...
+        // tell defmt we are formatting a `str` value
+        if fmt.needs_tag() {
+            let t = impls::str_tag();
+            fmt.u8(&t);
+        }
+        fmt.str(&buf);
+    }
+}
+
+/// `Display2Format` constructor
```
jonas-schievink, line 745:
```suggestion
/// `Display2Format` constructor.
```

jonas-schievink (approved):
Nice!

Needs an addition to the Qemu test, but otherwise looks good.

Issue #186, opened 2020-10-08
implement Format for char


japaric:
this should work but triggers a compiler error right now:
``` rust
defmt::info!("x={:?}", 'a');
```
the `{:char}` format specifier would also need to be implemented

japaric:
1. the easiest implementation could cast `char` to `u32` and send the 4 bytes
2. a better implementation would use invariants around UTF-8 to compress the data
3. alternative implementation: send the `char` as `[u8]`; would not compress as much as option 2

jonas-schievink:
Seems like we could also leb128-encode the `char`. That way, any ASCII characters will only take up 1 Byte. I guess this is similar to using UTF-8.

PR #282, opened 2020-11-26
Implement Format for char argument

justahero:
This PR adds support to format a single `char`.

It implements Format for the `char` argument.

```rust
defmt::info!("x = {:char}", 'y');
```

For now it is implemented by encoding / decoding the `char` as a `u32`, the first mentioned option in the related issue #186.

### DIFF
--- a/firmware/qemu/src/bin/log.rs
+++ b/firmware/qemu/src/bin/log.rs
@@ -482,6 +482,8 @@ fn main() -> ! {
         defmt::info!("{:?}", Display2Format::<consts::U32>(&addr));
     }
 
+    defmt::info!("Hello {:char}", '💜');
+
     defmt::info!("QEMU test finished!");
 
     loop {
--- a/firmware/qemu/src/bin/log.rs
+++ b/firmware/qemu/src/bin/log.rs
@@ -482,7 +482,7 @@ fn main() -> ! {
         defmt::info!("{:?}", Display2Format::<consts::U32>(&addr));
     }
 
-    defmt::info!("Hello {:char}", '💜');
+    defmt::info!("Hello {:char} & {:?}", '💜', '🍕');
 
     defmt::info!("QEMU test finished!");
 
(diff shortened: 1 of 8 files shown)


japaric (approved):
Looks great! Extending the format string is a breaking change so this needs to wait until v0.2.0 but I'll mark this as approved.

firmware/qemu/src/bin/log.rs
```diff
@@ -482,6 +482,8 @@ fn main() -> ! {
         defmt::info!("{:?}", Display2Format::<consts::U32>(&addr));
     }
 
+    defmt::info!("Hello {:char}", '💜');
```
japaric, line 485: could you also test formatting a `char` value using `{:?}`?

justahero: Yes, good idea. :+1:

japaric:
>  CI / qemu (nightly) (pull_request) Failing after 2m — qemu (nightly) 

to land breaking changes like these in `main` we'll need to disable the backwards compatibility test. The test needs to remain disabled until we release v0.2.0; at that point the test needs to be re-enabled (and updated to use a v0.2.0 decoder) cc @Lotterleben 
https://github.com/knurling-rs/defmt/blob/a6edbd1285505e09e0d022d9aee7313a7f7af9b0/.github/workflows/ci.yml#L129-L133

Lotterleben:
Added a milestone / issue so we don#t forget :)

Lotterleben:
> > CI / qemu (nightly) (pull_request) Failing after 2m — qemu (nightly)
> 
> to land breaking changes like these in `main` we'll need to disable the backwards compatibility test. The test needs to remain disabled until we release v0.2.0; at that point the test needs to be re-enabled (and updated to use a v0.2.0 decoder) cc @Lotterleben
> 
> https://github.com/knurling-rs/defmt/blob/a6edbd1285505e09e0d022d9aee7313a7f7af9b0/.github/workflows/ci.yml#L129-L133

@justahero @BriocheBerlin beat you to this, so a rebase should make this PR run green too https://github.com/knurling-rs/defmt/pull/284#event-4066470458

PR #284, opened 2020-11-27
Implement i128 and u128.

### LINKED ISSUE #199: impl Format for  i128/u128
Would also be nice to have i128/u128 while at it.

_Originally posted by @MathiasKoch in https://github.com/knurling-rs/defmt/issues/160#issuecomment-687833958_

BriocheBerlin:
Closes #199


firmware/qemu/src/bin/log.rs
```diff
@@ -419,6 +419,22 @@ fn main() -> ! {
...
 
+    defmt::info!(
+        "i128: 0 = {:i128}, -1 = {:i128}, MAX = {:i128}, MIN = {:i128}",
+        0,
+        -1,
+        i128::max_value(),
+        i128::min_value()
+    );
+
+    defmt::info!(
```
japaric, line 430: could you also test here that we can format `u128` and `i128` values using the `{:?}` format specifier?

japaric (approved):
Looks great! Extending the format string is a breaking change so this needs to wait until v0.2.0 but I'll mark this as approved.

firmware/qemu/src/bin/log.rs
BriocheBerlin: Is it ok now?

japaric: yes, looks good; thanks! I'll re-approve the PR but it will need a rebase before it can be merged.

japaric (approved):

Lotterleben:
Awesome, feel free to squash and merge!

PR #288, opened 2020-11-29
Book: Link to defmt-rtt and defmt-semihosting crates

dbrgn:
Makes it easier to get started if I can directly view the crate details :slightly_smiling_face:


japaric (approved):
Thanks!

PR #291, opened 2020-11-30
Enable test-only code via feature instead of target_arch.

### LINKED ISSUE #223: Allow using on x86_64
The library currently hardcodes test-only behaviors for x86_64. This makes it impossible to actually use it on x86_64. 

(The opposite is also true, I think: tests would break if you run them on a Raspberry Pi for example)

Dirbaio:
Fixes #223 

- Replace `target_arch = "x86_64"` with `feature = "unstable-test"` everywhere
- Add features to `Cargo.toml`s
- Add `--features unstable-test` in CI script

### DIFF
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -43,7 +43,7 @@ jobs:
       shell: bash
     - name: Run tests on Ubuntu/Windows
       if: matrix.os != 'macOS-latest'
-      run: cargo test --workspace
+      run: cargo test --workspace --features unstable-test
     - name: Run tests on macOS
       # NOTE defmt does not build for macOS because its `cortex-m-rt` dependency doesn't
       # (see https://github.com/rust-embedded/cortex-m-rt/issues/74), so we cannot use
@@ -97,7 +97,7 @@ jobs:
       with:
         mdbook-version: latest
     - name: Build defmt
-      run: cargo build
+      run: cargo build --features unstable-test
       # NOTE order of steps is important
     - name: Run book tests
       working-directory: book
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -41,6 +41,9 @@ jobs:
     - name: Check that all crates build
       run: RUSTFLAGS='--deny warnings' cargo check --all
       shell: bash
+    - name: Check that crates build with unstable-test feature
+      run: RUSTFLAGS='--deny warnings' cargo check --all --features unstable-test
+      shell: bash
     - name: Run tests on Ubuntu/Windows
       if: matrix.os != 'macOS-latest'
       run: cargo test --workspace --features unstable-test
--- a/macros/Cargo.toml
+++ b/macros/Cargo.toml
@@ -12,6 +12,10 @@ version = "0.1.1"
 [lib]
 proc-macro = true
 
+[features]
+# WARNING: for internal use only, not covered by semver guarantees
+unstable-test = []
+
 [dependencies]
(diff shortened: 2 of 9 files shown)


macros/Cargo.toml
```diff
@@ -12,6 +12,10 @@ version = "0.1.1"
 [lib]
 proc-macro = true
 
+[features]
+# WARNING: for internal use only, not covered by semver guarantees
+unstable-test = []
```
japaric, line 17: ~~I don't think the feature is necessary in this crate. The proc macro code is *emitting* `cfg(feature = "unstable-test")` but not using the Cargo feature.~~

EDIT: after reading the rest of the PR, I see that we do want this

macros/src/lib.rs
```diff
@@ -924,11 +924,11 @@ pub fn internp(ts: TokenStream) -> TokenStream {
     let section = format!(".defmt.prim.{}", sym);
 
     quote!(match () {
-        #[cfg(target_arch = "x86_64")]
+        #[cfg(feature = "unstable-test")]
```
japaric, line 927: this (cfg inside `quote!`) seems fine because `internp` is an implementation detail and only used within defmt non-public / non-user-facing code.

macros/src/lib.rs
```diff
@@ -992,11 +992,11 @@ fn mksym(string: &str, tag: &str, is_log_statement: bool) -> TokenStream2 {
         format_ident!("S")
     };
     quote!(match () {
-        #[cfg(target_arch = "x86_64")]
+        #[cfg(feature = "unstable-test")]
```
japaric, line 995: ~~this `cfg` however will end in the code emitted by the logging macros, `write!`, `intern!`, etc. I'm not sure if that's the behavior we want.~~

~~The other option is make the defmt macros *always* emit the `fetch_add_string_index` version if its "unstable-test" feature is enabled.~~

EDIT: OK, after reading the rest of the PR. I think this should be:
``` rust
if cfg!(feature = "unstable-test") {
    quote!(/* .. */) // fetch add string version
} else {
    quote!(/* .. */) // linker version
}
```

src/lib.rs
```diff
@@ -574,7 +574,7 @@ impl Formatter {
 // these need to be in a separate module or `unreachable!` will end up calling `defmt::panic` and
 // this will not compile
 // (using `core::unreachable!` instead of `unreachable!` doesn't help)
-#[cfg(target_arch = "x86_64")]
+#[cfg(feature = "unstable-test")]
 mod x86_64 {
```
japaric, line 578: nit: this module should be renamed as it's no longer x86 only

.github/workflows/ci.yml
```diff
@@ -43,7 +43,7 @@ jobs:
       shell: bash
     - name: Run tests on Ubuntu/Windows
       if: matrix.os != 'macOS-latest'
-      run: cargo test --workspace
+      run: cargo test --workspace --features unstable-test
```
japaric, line 46: in addition to this I think we should also run: `RUSTFLAGS='--deny warnings' cargo check --features unstable-test` (there's a previous build step that does this without the --features)

japaric (commented):
The approach looks good to me.

I think you meant to put the cfg instead the quote! macros though. See my inline comments.

Dirbaio:
Thanks for the review. Nice catch about the macros, I didn't think about that at all.
Review comments adressed, plus I fixed some comments and docs to reflect that defmt is now usable on x64.

japaric (approved):
This is great :100:  Thanks for implementing this!

PR #294, opened 2020-12-01
Allow use statements in defmt-test

### LINKED ISSUE #289: defmt-test: Allow use in test module
Hey, I'd like to propose a feature request:

Support use statements in the `#[defmt_test::tests]` macro. 
It is pretty common to import the output scope via `use super::*;`. 
But this is currently not possible:

```patch
 testsuite/tests/test.rs | 2 ++
 1 file changed, 2 insertions(+)

diff --git a/testsuite/tests/test.rs b/testsuite/tests/test.rs
index d1002ce..7a8d75b 100644
--- a/testsuite/tests/test.rs
+++ b/testsuite/tests/test.rs
@@ -7,6 +7,8 @@ use {{crate_name}} as _; // memory layout + panic handler
 // feature)
 #[defmt_test::tests]
 mod tests {
+    use super::*;
+
     #[test]
     fn assert_true() {
         assert!(true)
```

```log
$ cargo test
   ...
   Compiling use-test v0.1.0 (/tmp/use-test)
   Compiling testsuite v0.1.0 (/tmp/use-test/testsuite)
error: only function items are allowed
  --> testsuite/tests/test.rs:10:5
   |
10 |     use super::*;
   |     ^^^

error: aborting due to previous error

error: could not compile `testsuite`
```

As we can see, this case is currently just not taken into consideration:

https://github.com/knurling-rs/defmt/blob/2657a8db30a67c7f225611e2c14cb2759df876da/firmware/defmt-test/macros/src/lib.rs#L133-L138

The current workaround I use, it to declare the `use` statement in the function scope, e.g.

```patch
 #[defmt_test::tests]
 mod tests {
-    use super::*;
-
     #[test]
     fn assert_true() {
+        use super::*;
+         
         assert!(true)
```

If you point me into the right directions I could try to implement this
feature myself. But I've no experience writing procedural macros, so any help
is appreciated :)

Sh3Rm4n:
Closes #289 

Thanks to @japaric's [detailed instructions](https://github.com/knurling-rs/defmt/issues/289#issuecomment-735751942) this was straight forward to implement. 

The only thing missing right now, is to update the error message, as "only function items are allowed" is no longer true.

### DIFF
--- a/firmware/defmt-test/macros/src/lib.rs
+++ b/firmware/defmt-test/macros/src/lib.rs
@@ -34,6 +34,7 @@ fn tests_impl(args: TokenStream, input: TokenStream) -> parse::Result<TokenStrea
 
     let mut init = None;
     let mut tests = vec![];
+    let mut imports = vec![];
     for item in items {
         let item_span = item.span();
         match item {
@@ -130,10 +131,14 @@ fn tests_impl(args: TokenStream, input: TokenStream) -> parse::Result<TokenStrea
                 }
             }
 
+            Item::Use(u) => {
+                imports.push(u);
+            }
+
             _ => {
                 return Err(parse::Error::new(
                     item.span(),
-                    "only function items are allowed",
+                    "only `#[test]` functions and imports (`use`) are allowed in this scope",
                 ));
             }
         }
@@ -194,6 +199,7 @@ fn tests_impl(args: TokenStream, input: TokenStream) -> parse::Result<TokenStrea
         .map(|test| format!(".. {} ok", test.ident))
         .collect::<Vec<_>>();
     Ok(quote!(mod #ident {
+        #(#imports)*
         // TODO use `cortex-m-rt::entry` here to get the `static mut` transform
         #[export_name = "main"]
         unsafe extern "C" fn __defmt_test_entry() -> ! {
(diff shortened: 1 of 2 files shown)


firmware/defmt-test/macros/src/lib.rs
```diff
@@ -130,9 +131,14 @@ fn tests_impl(args: TokenStream, input: TokenStream) -> parse::Result<TokenStrea
...
 
+            Item::Use(u) => {
+                imports.push(u);
+            }
+
             _ => {
                 return Err(parse::Error::new(
                     item.span(),
+                    // TODO Make error message generic over Item
                     "only function items are allowed",
```
japaric, line 142: perhaps this should now say "only `#[test]` functions and imports (`use`) are allowed in this scope" (*). I think the error will have the right span and point to the item (e.g. a struct item) that's not allowed so I don't think it's super critical to report e.g. "this struct item is not allowed" from this error message.

(*) I *think* we don't allow non-`#[test]` functions here but can't remember

japaric (approved):
Looks great! I think that with an updated error message this would be good to go.

firmware/defmt-test/macros/src/lib.rs
Sh3Rm4n: > I think we don't allow non-#[test] functions here but can't remember

That is correct. 

This feature is only one addition to supported items. I did update the error to your suggestion. 
When more Items are supported the error should be made more generic, as manually declaring
which is, and which is not allowed seems error prone. 
However, this is nothing to worry about now. 

From my side this PR is ready to go :)

jonas-schievink (approved):
Great, thanks!

PR #296, opened 2020-12-03
Allow skipping the defmt version check

jonas-schievink:
If we add a `--ignore-version-mismatch` flag to probe-run, this will make development a bit easier.

### DIFF
--- a/elf2table/src/lib.rs
+++ b/elf2table/src/lib.rs
@@ -21,10 +21,21 @@ pub use defmt_decoder::Table;
 use defmt_decoder::{StringEntry, TableEntry};
 use object::{Object, ObjectSection};
 
-/// Parses an ELF file and returns the decoded `defmt` table
+/// Parses an ELF file and returns the decoded `defmt` table.
 ///
-/// This function returns `None` if the ELF file contains no `.defmt` section
+/// This function returns `None` if the ELF file contains no `.defmt` section.
 pub fn parse(elf: &[u8]) -> Result<Option<Table>, anyhow::Error> {
+    parse_impl(elf, true)
+}
+
+/// Like `parse`, but does not verify that the defmt version in the firmware matches the host.
+///
+/// Use with caution, this is meant for defmt/probe-run development only.
+pub fn parse_ignore_version(elf: &[u8]) -> Result<Option<Table>, anyhow::Error> {
+    parse_impl(elf, false)
+}
+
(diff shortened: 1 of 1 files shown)


elf2table/src/lib.rs
```diff
@@ -21,10 +21,21 @@ pub use defmt_decoder::Table;
...
 ///
-/// This function returns `None` if the ELF file contains no `.defmt` section
+/// This function returns `None` if the ELF file contains no `.defmt` section.
 pub fn parse(elf: &[u8]) -> Result<Option<Table>, anyhow::Error> {
+    parse_impl(elf, true)
+}
+
+/// Like `parse`, but does not verify that the defmt version in the firmware matches the host.
+///
+/// Use with caution, this is meant for defmt/probe-run development only.
```
Lotterleben, line 33: nitpick: I'd put this at the top of the docs so it can't be overseen. plus the `--ignore-version-mismatch`probe-run flag should get a "CAUTION: .." doc too?

elf2table/src/lib.rs
```diff
@@ -75,7 +86,9 @@ pub fn parse(elf: &[u8]) -> Result<Option<Table>, anyhow::Error> {
         }
     };
 
-    defmt_decoder::check_version(version).map_err(anyhow::Error::msg)?;
+    if check_version {
```
Lotterleben, line 89: personally, I'd rather pull this out of `parse_impl()` and move it into `parse()` right before the call to `parse_impl()`. That way we can avoid having a flag argument* as well as tight coupling between the parsing and defmt consistency check, which imo are two different tasks.

* the function names of `parse_ignore_version()` and `parse()` imo already communicate what's happening better than parse(.. , true) does

jonas-schievink: Doing the version check requires some amount of processing to get the `version`, so I'd have to duplicate that into the caller.

> tight coupling between the parsing and defmt consistency check, which imo are two different tasks.

There is some amount of intentional coupling between them, for example, we always check that there's only one version in use, and we check for the existence of the `.defmt` section and the version symbol to determine whether defmt is in use at all, or the linker script is broken.

This PR is only intended to skip the version check, not any of the other consistency checks.

Lotterleben: hrm , ok then that's maybe one for future refactoring sessions...

Lotterleben (approved):

PR #300, opened 2020-12-14
Use primitive formatting for primitive references

### LINKED ISSUE #163: #[derive] should use primitive formatting for &i32 and similar
``` rust
#[derive(Format)]
struct S<'a> {
    x: &'a i32,
}
```
expands into:
``` rust
fn format(&self, f: &mut defmt::Formatter) {
    match (f, &(self.x)) {
        (ref mut _fmt_, arg0) => {
            if _fmt_.needs_tag() {
                _fmt_.istr(&defmt::export::istr(match () {
                    #[cfg(not(target_arch = "x86_64"))]
                    () => {
                        #[link_section = ".defmt.fmt.S {{ x: {:?} }}"] // <-
                        #[export_name = "S {{ x: {:?} }}@#18 bytes(276..282)"] // <-
                        static S: u8 = 0;
                        &S as *const u8 as usize
                    }
                }));
            }
            _fmt_.fmt(arg0, false); // <-
            _fmt_.finalize();
        }
    };
}
```
The format string should be `S {{ x: {:i32} }}` Instead of the `fmt` method `derive` should use the built-in `i32` method.
This change (from `Format` to built-in formatter) should be applied to all references (`&T` and `&mut T`) to primitives

Sh3Rm4n:
Closes #163

A single recursive call in `as_native_type` is sufficient
to unpack the type from the reference.
Formatting functions (`fmt.i32()` etc.) are already taking
a reference, so this does not need to be changed.

Tests are adjusted, so that they pass.
This introduced change is not explicitly tested though.

```rust
#[derive(Format)]
struct S2<'a> {
    x: &'a i32,
    y: &'a u8,
}
```

now expands into

```rust
fn format(&self, f: &mut defmt::Formatter) {
    if f.needs_tag() {
        f.istr (&defmt::export::istr({ 
            #[link_section = ".defmt.{\"package\":\"use-test\",\"tag\":\"defmt_fmt\",\"data\":\"S2 {{ x: {:i32}, y: {:u8} }}\",\"disambiguator\":\"10404396863279507298\"}"]
            #[export_name = "{\"package\":\"use-test\",\"tag\":\"defmt_fmt\",\"data\":\"S2 {{ x: {:i32}, y: {:u8} }}\",\"disambiguator\":\"10404396863279507298\"}"]
            static S: u8 = 0; 
            &S as * const u8 as usize 
        })) ;
    }
    match self {
        Self { x, y } => {
            f.i32(x);
            f.u8(y);
        }
    }
}
```

### DIFF
--- a/macros/src/lib.rs
+++ b/macros/src/lib.rs
@@ -422,6 +422,7 @@ fn as_native_type(ty: &Type) -> Option<String> {
             }
             None => None,
         },
+        Type::Reference(tref) => as_native_type(&*tref.elem),
         _ => None,
     }
 }
--- a/macros/src/lib.rs
+++ b/macros/src/lib.rs
@@ -447,11 +447,7 @@ fn log(level: Level, log: FormatArgs) -> TokenStream2 {
     let ls = log.litstr.value();
     let fragments = match defmt_parser::parse(&ls) {
         Ok(args) => args,
-        Err(e) => {
-            return parse::Error::new(log.litstr.span(), e)
-                .to_compile_error()
-                .into()
-        }
+        Err(e) => return parse::Error::new(log.litstr.span(), e).to_compile_error(),
     };
 
     let args = log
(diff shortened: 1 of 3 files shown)


japaric (approved):
Nice! This looks good.

> This introduced change is not explicitly tested though.

(the unit test fixed in this PR is testing the derive macro change, at least it shows that the encoding is now smaller)

The CI/nightly build is broken for reasons unrelated to this this PR. We should figure out some workaround for that ...

PR #302, opened 2020-12-15
derive(Format): Support more than 256 variants

### LINKED ISSUE #63: derive(Format): support more than 256 variants
the discriminant should be encoded in the smallest integer (LE encoding) that can enumerate all the variants. [Previous discussion](https://github.com/ferrous-systems/binfmt/issues/61)

Sh3Rm4n:
Closes #63


japaric (commented):
The encoder part looks good but I think we also need to do changes in the decoder.

Could you add the `EnumLarge` as a test to `firmware/log.rs` and test out a small enum and a large (>256) enum?

firmware/qemu/src/bin/log.rs
```diff
@@ -501,6 +501,32 @@ fn main() -> ! {
...
+            A162, A163, A164, A165, A166, A167, A168, A169, A170, A171, A172, A173, A174, A175, A176, A177, A178, A179,
+            A180, A181, A182, A183, A184, A185, A186, A187, A188, A189, A190, A191, A192, A193, A194, A195, A196, A197,
+            A198, A199, A200, A201, A202, A203, A204, A205, A206, A207, A208, A209, A210, A211, A212, A213, A214, A215,
+            A216, A217, A218, A219, A220, A221, A222, A223, A224, A225, A226, A227, A228, A229, A230, A231, A232, A233,
+            A234, A235, A236, A237, A238, A239, A240, A241, A242, A243, A244, A245, A246, A247, A248, A249, A250, A251,
+            A252, A253, A254, A255, A256, A257, A258, A259, A260, A261, A262, A263, A264, A265, A266, A267, A268, A269,
+        }
+
+        defmt::info!("Format EnumLarge Variant: {:?}", EnumLarge::A007);
+        defmt::info!("Format EnumLarge Variant: {:?}", EnumLarge::A269);
```
japaric, line 527: this fails locally with 
``` console
0.000101 INFO Format EnumLarge Variant: A007
failed to decode defmt data: [0, 99, 1]
Error: malformed data
```
I think the decoder expects that the *discriminant* is encoded as a single byte but with these changes it should consider how many the variants the enum has (and use 1,2,3,4,etc. bytes as the discriminant size)

Sh3Rm4n: Whoops interesting. It's failing locally for me as well. I just pushed it to have a failing case to move along and eventually make it pass. 

But that this does not fail in CI is interesting.

Sh3Rm4n: Well I guess it is because I have not added any `.out` files? :thinking:

japaric: I think CI is not running because the PR has merge conflicts. The last CI run is from your very first version, which didn't have the qemu test.

japaric:
Nice! Thanks for the PR.

bors r+

macros/src/lib.rs
```diff
@@ -278,13 +270,34 @@ pub fn format(ts: TokenStream) -> TokenStream {
...
 
-                    let encode_discriminant = if de.variants.len() == 1 {
+                    let len = de.variants.len();
+                    let encode_discriminant = if len == 1 {
                         // For single-variant enums, there is no need to encode the discriminant.
                         quote!()
-                    } else {
+                    } else if let (Ok(_), Ok(i)) = (u8::try_from(len), u8::try_from(i)) {
                         quote!(
                             f.u8(&#i);
```
japaric, line 279: the GitHub diff is not showing this line correctly but after PR #305 the new statements need to be `f.inner.u8(&#i)`.**k b**

Sh3Rm4n: Rebased to resolve this

japaric:
Thanks! (and sorry for the repeated rebase work)

bors r+

Sh3Rm4n:
No problem, these were minor anyways. :)

PR #305, opened 2020-12-15
Change Format::format to take Formatter argument by value

### LINKED ISSUE #277: consider making `Format::format` take the `Formatter` by value
the issue: calling some (other type) `format` method more than once corrupts the defmt data stream
``` rust
struct MyU8 { inner: u8 }

impl defmt::Format for MyU8 {
    fn format(&self, f: &mut defmt::Formatter) {
        self.inner.format(f); // OK
        self.inner.format(f); // corrupts the stream
    }
}
```
a solution could be to change the signature of `Format::format` to `fn(&self, defmt::Formatter)`.
This way a `format` method can only be called once: calling the method more than once produces a compile error.

TBD:
- this compile time check should extend to `write!`
- does `derive(Format)` need to "undo" the value semantics to continue working?

BriocheBerlin:
Closes #277


japaric:
https://github.com/knurling-rs/defmt/pull/305/checks?check_run_id=1558757983#step:6:117

```
  --> firmware/qemu/src/bin/double-write.rs:16:9
   |
15 |         defmt::write!(fmt, "one");
   |         -------------------------- value moved here
16 |         defmt::write!(fmt, "two");
   |         ^^^^^^^^^^^^^^^^^^^^^^^^^^ value used here after move
   |
   = note: move occurs because `fmt.inner` has type `&mut InternalFormatter`, which does not implement the `Copy` trait
```

this test (`double-write.rs`) can be removed. With these changes we don't need the runtime check because the issue is prevent at compile time.

japaric:
Some follow-up work to do here. It doesn't need to be done in *this* PR but I wanted to jot down some notes somewhere:
- remove double/nested `write!` runtime check (no longer needed because we prevent the issue at compile time)
- update book / docs to mention the limitations of `write!` and the new `Format::format` signature
- compile time tests that check that one cannot invoke `write!` twice within a `Format::format` implementation

japaric:
https://github.com/knurling-rs/defmt/pull/305/checks?check_run_id=1558757817#step:6:147

a code snippet needs to be updated in the book https://github.com/knurling-rs/defmt/blob/dd056e6b7ea807e43f851c75b6fa24bfd7cf9f5a/book/src/re-entrancy.md#L16-L26

like with API documentation, we test that the snippets compile to keep them in sync with library changes

PR #308, opened 2020-12-16
compile-fail test new Formatter move semantics

japaric:
now that the signature of `Format::format` has changed we expect that these uses of the `write!`
macro and `format` method fail so let's test that

the `fmt-*` examples were corrupting the defmt data stream before but now are rejected at compile
time

the error message for the `write!` examples could be better: it should not mention the
`InternalFormatter` type, which is an implementation detail. I'll open an issue about improving the
diagnostics of that macro


Lotterleben (approved):
nice!

PR #312, opened 2020-12-17
derive(Format): better encode `str` fields

japaric:
use {:str} instead of {:?}


jonas-schievink (approved):
Looks good! Do we test decoding of this too?

japaric:
added a qemu test (I don't think we can test `derive` in the `defmt_decoder` unit tests)

PR #321, opened 2020-12-20
Always respect DisplayHint::Ascii for byte slices

### LINKED ISSUE #318: Byte slices ignore display hints when they appear behind a `Format` impl
These two lines should print the same but don't.

``` rust
defmt::info!("{=[u8]:a}", bytes); // -> INFO b"Hi"
defmt::info!("{:a}", bytes);      // -> INFO [72, 105]
```

Sh3Rm4n:
This is a naive implementation, which assumes, that the `Uxx` value
is always in `u8` range. It does not consider the given format type `{=$type}`

Also, it assumes that the element args array only contains one element

Fixes #318

### DIFF
--- a/decoder/src/lib.rs
+++ b/decoder/src/lib.rs
@@ -1038,16 +1038,37 @@ fn format_args_real(
                     Arg::IStr(x) => format_str(x, hint, &mut buf)?,
                     Arg::Format { format, args } => buf.push_str(&format_args(format, args, hint)),
                     Arg::FormatSlice { elements } => {
-                        buf.write_str("[")?;
-                        let mut is_first = true;
-                        for element in elements {
-                            if !is_first {
-                                buf.write_str(", ")?;
+                        match hint {
+                            // Filter Ascii Hints, which contains u8 byte slices
+                            Some(DisplayHint::Ascii)
+                                if elements.iter().filter(|e| e.format == "{=u8}").count() != 0 =>
+                            {
+                                let vals = elements
+                                    .iter()
+                                    .map(|e| match e.args.as_slice() {
+                                        [Arg::Uxx(v)] => u8::try_from(*v).expect("the value must be in u8 range"),
+                                        _ => panic!("FormatSlice should only contain one argument"),
+                                    })
+                                    .collect::<Vec<u8>>();
+                                format_bytes(&vals, hint, &mut buf)?
                             }
-                            is_first = false;
-                            buf.write_str(&format_args(element.format, &element.args, hint))?;
+                            _ => {
+                                buf.write_str("[")?;
+                                let mut is_first = true;
+                                for element in elements {
+                                    if !is_first {
+                                        buf.write_str(", ")?;
+                                    }
+                                    is_first = false;
+                                    buf.write_str(&format_args(
+                                        element.format,
+                                        &element.args,
+                                        hint,
+                                    ))?;
+                                }
+                                buf.write_str("]")?;
+                            }
                         }
-                        buf.write_str("]")?;
                     }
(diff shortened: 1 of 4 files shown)


decoder/src/lib.rs
```diff
@@ -1025,18 +1025,38 @@ fn format_args_real(
...
+                        match hint {
+                            Some(DisplayHint::Ascii) => {
+                                let vals = elements
+                                    .iter()
+                                    .filter_map(|e| match e.args[0] {
+                                        Arg::Uxx(v) => Some(u8::try_from(v).ok()?),
+                                        _ => None,
+                                    })
+                                    .collect::<Vec<u8>>();
+                                format_bytes(&vals, hint, &mut buf)?
```
jonas-schievink, line 1037: When formatting a slice of structs (or anything that isn't a `u8`) with `{=[?]:a}`, I'd expect the display hint to be ignored, but the code looks like it'll instead output nothing.

jonas-schievink: For example, you might want to format this struct with `{=[?]:a}` in order to display the `name` field as an ASCII string:

```rust
struct Data<'a> {
    name: &'a [u8],
    value: bool,
}
```

Sh3Rm4n:
I've resolved the concerns in https://github.com/knurling-rs/defmt/pull/321#discussion_r546697911, but I'm still making assumption, where I'm not sure if they hold true for every case. 

1. If any element in a slice has the `"{=u8}"` format, does that mean, that all elements have this format?
2. Does `"{=u8}"` format ensure, that the values are also always in `u8` range?

And I feel, that this code can be written a bit more elegant / readable. But it works for now :)

decoder/src/lib.rs
```diff
@@ -1025,16 +1025,38 @@ fn format_args_real(
...
-                                buf.write_str(", ")?;
+                        match hint {
+                            // Filter Ascii Hints, which contains u8 byte slices
+                            Some(DisplayHint::Ascii)
+                                if elements.iter().filter(|e| e.format == "{=u8}").count() != 0 =>
+                            {
+                                let vals = elements
+                                    .iter()
+                                    // Assumption: Arguments for u8 has always only 1 element
+                                    .filter_map(|e| match e.args[0] {
```
japaric, line 1036: I would suggest using `map` instead of `filter_map` and panicking if `arg` is not a single `Arg::Uxx`. Because: if for some reason (i.e. a bug) the program reaches that state then the current logic will silently ignore the errors and print the wrong thing. With a panic the bug will be immediately exposed.

japaric (approved):
Thanks for the PR. This looks good. I left a suggestion.

Regarding your questions / assumptions

> If any element in a slice has the "{=u8}" format, does that mean, that all elements have this format?

Yes. The format string (field) *can* vary from element to element (of a format slice) but only *if* we are formatting a slice of enums. In that case, the format string is the variant and not the complete enum (e.g. `Foo({:u8})` (variant) vs `Foo({:u8})|Bar({:u16})` (enum)). The format string of a single variant cannot be `{=u8}`, however.

> Does "{=u8}" format ensure, that the values are also always in u8 range?

Yes, because the macros do type checking: the value must have `u8` type.

> Do FormatSlice elements for "{=u8}" always contain only one value, and when is this not true?

Yes.

decoder/src/lib.rs
Sh3Rm4n: Converted both assumptions to panics, if they are not true.

Sh3Rm4n:
Rebased and sqaushed. Ready for review :)

japaric (approved):
Great work! Thanks

PR #323, opened 2020-12-20
Merge I128/Ixx and U128/Uxx variants

### LINKED ISSUE #314: Merge I128/Ixx and U28/Uxx variants
Currently, the `Arg` enum has two matching *xx/*128 variants which are treated the same e.g. in `format_args_real()`. They should be merged into one,

Sh3Rm4n:
Closes #314 

Also lift 32 bit range for bitfields introduced in #13
Could be split into another PR, if concerns are raised.


jonas-schievink:
Looks reasonable, thanks!

I think we should add a QEMU test for the larger bitfield range.

Sh3Rm4n:
I've added a qemu test and as it shows, some more places has to be changed to support 128 bitfields. 

I've also added support for ascii hints for bitfields :)

decoder/src/lib.rs
```diff
@@ -1608,6 +1621,31 @@ mod tests {
         );
     }
 
+    #[test]
+    fn bitfields_u128() {
```
japaric, line 1625: I was a bit surprised about this test because expected that `{0=119..124:b}` is encoded as 2 bytes instead of as 16 bytes. After checking locally, this test also works if I remove the last 14 bytes. It would be nice to make `decode_and_expect` check that the input data was fully consumed to ensure we are not passing extra bytes that will be ignored.

japaric (approved):
Thanks for the PR. This looks good!

japaric:
CI should be fixed now
bors retry

PR #324, opened 2020-12-31
Fix clippy lints

Sh3Rm4n:


Sh3Rm4n:
Feel free to cherry-pick commits, if some changes are not desired. 

Clippy was added to the CI to confirm the correct alliance of the suggested fixes.
I'll can remove or adjust this part, if you find clippy to annoying. (Maybe only show clippy erros and ignore clippy warnings?) 

I'm thinking about adding a rustfmt ci job as well? The code base is mostly consistent with the default rustfmt style, but not everywhere.

I'm not sure about the UI tests failures. Is this directly related to my changes or where these introduced because of the 1.49 rustc release?

.github/workflows/ci.yml
```diff
@@ -78,6 +78,22 @@ jobs:
...
+      - uses: actions/checkout@v2
+      - uses: actions-rs/toolchain@v1
+        with:
+          toolchain: stable
+          override: true
+          profile: minimal
+          components: clippy
+      - uses: actions-rs/clippy-check@v1
+        with:
+          token: ${{ secrets.GITHUB_TOKEN }}
```
japaric, line 93: (why) is this needed? (also, I don't think we have set any secrets for this repo.)

tests/ui/panic-handler-export-name.stderr
```diff
@@ -2,4 +2,4 @@ error: `#[panic_handler]` attribute cannot be used together with `#[export_name]
  --> $DIR/panic-handler-export-name.rs:5:1
   |
 5 | #[export_name = "hello"]
-  | ^^^^^^^^^^^^^^^^^^^^^^^^
```
japaric, line 5: you can undo these changes. I've fixed the UI issues in PR #325

japaric (commented):
Thanks for the PR. The clippy fixes look good to me. I'm OK with adding a clippy CI job. A rustfmt one would also be welcomed.

.github/workflows/ci.yml
Sh3Rm4n: This is what the [`clippy-check`](https://github.com/actions-rs/clippy-check/#example-workflow) action requires and suggest in its default configuration. 

This step even fails, when this toke key is not set. I guess, it's about a hidden token generated by github, which has to be passed to clippy-check, so that it can annotate PRs with clippy warnings? TBH, I haven't looked much further. The correct answer can be probably found at the [`clippy-check`](https://github.com/actions-rs/clippy-check) repository.

Sh3Rm4n:
I've added rustfmt and rebased onto main, so that the UI tests are running again. 

I stopped at `#[deny(clippy::uninit_assumed_init)]`, where I do understand the incentive, but I don't know how to solve it, nor if a solution is even necessary. I'm too inexperienced to evaluate safety / undefined behavior issues. 

Same goes for the `#[warn(clippy::missing_safety_doc)]` warnings. Though, these are warnings and could be technically ignored.

PR #325, opened 2021-01-04
update UI tests

japaric:
span info has changed in the last rust update (stable channel)

PR #326, opened 2021-01-04
GHA: remove ci-failure

japaric:
bors' semantics have changed: skipped checks are now considered failures
the result is that when all build jobs pass the ci-failure reports a "ci" check as skipped
(expected behavior) but that is now considered a failure by bors

removing the ci-failure conditional should fix the issue
if any build job fails then ci-success will report "ci" as skipped, which bors will consider as a
failure


japaric:
right, this needs to land together with PR #325

PR #327, opened 2021-01-06
impl<T> Format for PhantomData<T>

mattico:
I needed this to be able to `#[derive(Format)]` on a few structs.

### DIFF
--- a/src/impls.rs
+++ b/src/impls.rs
@@ -346,6 +346,15 @@ impl Format for () {
     }
 }
 
+impl<T> Format for core::marker::PhantomData<T> {
+    fn format(&self, f: Formatter) {
+        if f.inner.needs_tag() {
+            let t = internp!("PhantomData");
+            f.inner.u8(&t);
+        }
+    }
+}
+
 macro_rules! tuple {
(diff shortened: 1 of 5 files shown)


src/impls.rs
```diff
@@ -346,6 +346,15 @@ impl Format for () {
     }
 }
 
+impl<T> Format for core::marker::PhantomData<T> {
+    fn format(&self, f: Formatter) {
+        if f.inner.needs_tag() {
+            let t = internp!("PhantomData<_>");
```
japaric, line 352: could you change the string to `PhantomData` so that this resembles how `core::fmt::Debug` formats this type?

japaric (commented):
Thanks for the PR! This looks good. Could you please extend the snapshot test in `firmware/qemu/src/bin/log.rs` to test that something like `S { x: PhantomData<u8>, y: u8 }` is correctly formatted by `{=[?]}`? I'd like to confirm that the tag is omitted when you format a slice of `PhantomData`s and the whole thing is properly decoded.

japaric (approved):
Thanks!

Issue #306, opened 2020-12-15
attributes: add more compile fail tests


japaric:
now that there's CI infrastructure in place for compile fail tests (added in #293) we should add tests for these scenarios:

- `timestamp` and `panic_handler` used on a function with the wrong signature. For instance, all these examples are UB (they are rejected at compile time today):

``` rust
#[timestamp]
fn foo(x: u64) -> u64 { x + 1 }
```

``` rust
#[panic_handler]
fn foo(x: bool) -> ! { info!("{:?}", x); loop {} }
```

``` rust
#[panic_handler]
fn foo() {}
```

- `global_logger` used on a struct that has fields. I don't think this is much of an issue (unsure if it's even checked today) but the fields won't be accessed by `defmt` so accepting this would be misleading at best.

these compile fail tests are also a good place to check the helpfulness of our error messages. For example, this:

``` rust
defmt::write!(42, "hello");
```

should say "error: expected `defmt::Formatter`, found `i32`", ideally with a span that points to the `42` expression, instead of the a helpful "method `needs_tag` not found on `i32` value". (not sure what the current error message says)

PR #331, opened 2021-01-11
More compile fail tests

BriocheBerlin:
More ui tests for `timestamp` and `panic_handler`.
Closes https://github.com/knurling-rs/defmt/issues/306.


japaric (approved):
Thanks!

PR #335, opened 2021-01-12
add defmt-itm

japaric:
this is a new defmt transport crate

### DIFF
--- a/firmware/defmt-itm/src/lib.rs
+++ b/firmware/defmt-itm/src/lib.rs
@@ -1,12 +1,14 @@
-//! [`defmt`](https://github.com/knurling-rs/defmt) global logger over RTT.
+//! [`defmt`](https://github.com/knurling-rs/defmt) global logger over ITM.
 //!
-//! NOTE when using this crate it's not possible to use (link to) the `rtt-target` crate
-//!
-//! To use this crate, link to it by importing it somewhere in your project.
+//! To use this crate, call the `enable` function
 //!
 //! ```
 //! // src/main.rs or src/bin/my-app.rs
-//! use defmt_rtt as _;
+//!
+//! fn main() -> ! {
+//!     let p = cortex_m::Peripherals::take().unwrap();
+//!     defmt_itm::enable(p.ITM);
+//! }
 //! ```
 
 #![no_std]
--- a/firmware/defmt-itm/src/lib.rs
+++ b/firmware/defmt-itm/src/lib.rs
@@ -23,7 +23,7 @@ static ENABLED: AtomicBool = AtomicBool::new(false);
 /// Enables defmt logging over the ITM stimulus port 0
 pub fn enable(itm: ITM) {
     // enable stimulus port 0
-    itm.ter[0].write(1);
+    unsafe { itm.ter[0].write(1) }
     drop(itm);
     ENABLED.store(true, Ordering::Relaxed);
 }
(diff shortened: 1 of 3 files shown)


firmware/defmt-itm/src/lib.rs
```diff
@@ -0,0 +1,76 @@
...
+    ptr::NonNull,
+    sync::atomic::{AtomicBool, Ordering},
+};
+
+use cortex_m::{interrupt, itm, peripheral::ITM, register};
+
+static ENABLED: AtomicBool = AtomicBool::new(false);
+
+/// Enables defmt logging over the ITM stimulus port 0
+pub fn enable(itm: ITM) {
```
japaric, line 24: we may want to make the stimulus port configurable in the feature that can be added backwards compatibility by adding a second function with signature `fn(ITM, port_number: u8)`

firmware/defmt-itm/src/lib.rs
```diff
@@ -0,0 +1,76 @@
...
+    // enable stimulus port 0
+    itm.ter[0].write(1);
+    drop(itm);
+    ENABLED.store(true, Ordering::Relaxed);
+}
+
+#[defmt::global_logger]
+struct Logger;
+
+impl defmt::Write for Logger {
```
japaric, line 34: (one thing I noticed is that this implementation is exactly the same as the one in `defmt-rtt`)

firmware/defmt-itm/src/lib.rs
```diff
@@ -0,0 +1,76 @@
+//! [`defmt`](https://github.com/knurling-rs/defmt) global logger over RTT.
+//!
+//! NOTE when using this crate it's not possible to use (link to) the `rtt-target` crate
```
jonas-schievink:
```suggestion
//! [`defmt`](https://github.com/knurling-rs/defmt) global logger over ITM.
```

firmware/defmt-itm/src/lib.rs
jonas-schievink: I'm assuming you mean the `acquire`/`release` impl further down?

firmware/defmt-itm/src/lib.rs
```diff
@@ -0,0 +1,76 @@
...
+use core::{
+    ptr::NonNull,
+    sync::atomic::{AtomicBool, Ordering},
+};
+
+use cortex_m::{interrupt, itm, peripheral::ITM, register};
+
+static ENABLED: AtomicBool = AtomicBool::new(false);
+
+/// Enables defmt logging over the ITM stimulus port 0
```
jonas-schievink, line 25:
```suggestion
/// Enables defmt logging over the ITM stimulus port 0.
///
/// This needs to be called by the application before defmt logging is used, otherwise the logs will be disposed.
```

Could we do this automatically, perhaps? Or would that have any problematic side-effects?

firmware/defmt-itm/src/lib.rs
japaric: yes, the acquire / release implementations look exactly the same

firmware/defmt-itm/src/lib.rs
japaric: you mean writing to `ITM.TER` lazily? that can be done
what I wanted to about though is the user sending unrelated data over the stimulus port 0 (using e.g. `iprintln!`) as that would corrupt the defmt data stream

firmware/defmt-itm/src/lib.rs
jonas-schievink: Ah you fixed this already

firmware/defmt-itm/src/lib.rs
jonas-schievink: Ah, right. Seems fine then.

Issue #309, opened 2020-12-16
improve write! macro diagnostics


japaric:
This code:

``` rust
struct S;

impl defmt::Format for S {
    fn format(&self, f: defmt::Formatter) {
        defmt::write!(f, "hello");
        defmt::write!(f, "world");
    }
}
```

produces the following error:

``` text
error[E0382]: use of moved value: `f.inner`
 --> $DIR/write-twice.rs:6:9
  |
5 |         defmt::write!(f, "hello");
  |         -------------------------- value moved here
6 |         defmt::write!(f, "world");
  |         ^^^^^^^^^^^^^^^^^^^^^^^^^^ value used here after move
  |
  = note: move occurs because `f.inner` has type `&mut InternalFormatter`, which does not implement the `Copy` trait
  = note: this error originates in a macro (in Nightly builds, run with -Z macro-backtrace for more info)
```

it should not mention the `inner` field or the `InternalFormatter` type because those are implementation details.

I think the error could be improved by changing expansion to look like this (maybe?):

``` diff
 struct S;
 impl defmt::Format for S {
     fn format(&self, f: defmt::Formatter) {
         // first write macro call
+        {
+            let f: defmt::Formatter = f; // force a move & type check the argument
             match (f.inner,) {
                 // ..
             };
+        }

         // second write macro call
+        {
+            let f: defmt::Formatter = f;
             match (f.inner,) {
                 // ..
             };
+        }
     }
 }
```

japaric:
relevant proc-macro code:
https://github.com/knurling-rs/defmt/blob/2ffdf3c90b1c4f85d086fbd2ff692aa205414af0/macros/src/lib.rs#L1001-L1011

PR #337, opened 2021-01-12
Improve diagnostics on double `write!`

jonas-schievink:
Fixes https://github.com/knurling-rs/defmt/issues/309


japaric (approved):
Nice! Thanks

PR #338, opened 2021-01-12
add defmt-logger and defmt-print

japaric:
defmt-logger is probe-run's code that formats defmt frames into a log::Log-ger sink
defmt-print is a command-line tool that decodes defmt and prints it to the console

TODO
- [ ]  `defmt-print --version` should print supported defmt version

other thoughts: qemu-run should maybe use defmt-logger instead of its own thing?


jonas-schievink:
This seems good!

bors r+

PR #339, opened 2021-01-12
Make leb64 encoding fully safe

jonas-schievink:
This removes all unchecked indexing as well as the usage of `MaybeUninit::uninit().assume_init()` from the leb64 encoding logic.

Resulting binary size change for the `log` binary:

```
   text	   data	    bss	    dec	    hex	filename
  33708	      0	      8	  33716	   83b4	log-main
  33508	      0	      8	  33516	   82ec	../../target/thumbv7m-none-eabi/release/log
```

That's right, it's 200 Bytes smaller.

I've also tried keeping `MaybeUninit` in place, which reduces binary size by like 400 *more* bytes. That seemed spooky, so I didn't go for that (zeroing 10 bytes shouldn't take 400 Bytes of code space? does it pull in `memset`? does it miscompile something due to the UB?). Haven't investigated that further.

### DIFF
--- a/src/leb.rs
+++ b/src/leb.rs
@@ -1,6 +1,4 @@
-/// # Unsafety
-/// `buf` must be large enough to hold the encoded value
-pub(crate) unsafe fn leb64(x: u64, buf: &mut [u8]) -> usize {
+pub(crate) fn leb64(x: u64, buf: &mut [u8; 10]) -> usize {
     let mut low = x as u32;
     let mut high = (x >> 32) as u32;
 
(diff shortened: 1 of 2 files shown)


japaric (approved):
This is better. Thanks.

> That's right, it's 200 Bytes smaller.

was that with opt-level=3? maybe the indexing prevents loop unrolling and that makes the binary smaller? I would have no idea why keeping the assume-uninit makes the program that much smaller though. llvm optimizations are unpredictable IMO

jonas-schievink:
> was that with opt-level=3?

It was `cargo build --release` in the `firmware/qemu` directory, which should be opt-level=3, LTO=fat.

japaric:
needs a rebase but feel free to submit to bors once that's resolved

jonas-schievink:
bors r=japaric

Issue #292, opened 2020-11-30
support impl Debug and impl Display values directly in logging macros


japaric:
This is different from the existing adapter types: Debug2Format and Display2Format in that formatted string will never be truncated. It still has the disadvantage of "no compression"

the rough idea is to
1. extend defmt wire format to support "unsized values" (streams?). That is values whose length (in bytes once encoded) is not known before the value's data written into a `Formatter`
2. extend the format string with say `{:Debug}`. This string expects an `impl Debug` value. The "unsized" value is then written into the `defmt::Formatter` probably using some adapter to make `defmt::Formatter` look like a `impl core::fmt::Write` type.

jonas-schievink:
Since the result of `Display` and `Debug` is always valid UTF-8, we could use an invalid UTF-8 byte as the end-of-frame delimiter (say, `0xFF`). That way we don't have to implement something more complex or expensive like SLIP or even COBS.

jonas-schievink:
These will unfortunately make our format string syntax deviate quite a bit from `core::fmt`, since these won't be able to use `{}` and `{:?}` like you would expect from `core::fmt`.

Could we maybe force these to go through `{:?}`/`Format` and use some secret syntax that isn't exposed to the user? Then `Display2Format` and `Debug2Format` would remain, but become simple wrappers.

PR #340, opened 2021-01-12
Streaming core::fmt adapters

jonas-schievink:
This removes heapless, the `consts` reexport, and the need to manually specify buffer sizes for the `X2Format` adapters. Instead of formatting to an on-stack buffer, these adapters now stream the formatted data over the defmt sink and terminate it with a `0xff` byte.

Display hints are, perhaps confusingly, ignored. We could pass them to `core::fmt` if we like, but that's a non-breaking change and can be done later.

No user-facing format type (`{=Debug}`) is provided, since I felt that that may cause confusion between `{}`, `{:?}`, `{=Debug}` and `{=Display}`, which would all mean different things.

Closes https://github.com/knurling-rs/defmt/issues/292

### DIFF
--- a/decoder/src/lib.rs
+++ b/decoder/src/lib.rs
@@ -349,6 +349,9 @@ enum Arg<'t> {
     Slice(Vec<u8>),
     /// Char
     Char(char),
+
+    /// `fmt::Debug` / `fmt::Display` formatted on-target.
+    Preformatted(String),
 }
 
 #[derive(Debug, PartialEq)]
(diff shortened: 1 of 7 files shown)


decoder/src/lib.rs
```diff
@@ -873,6 +876,20 @@ impl<'t, 'b> Decoder<'t, 'b> {
                     let c = std::char::from_u32(data).ok_or(DecodeError::Malformed)?;
                     args.push(Arg::Char(c));
                 }
+                Type::Debug | Type::Display => {
+                    // UTF-8 stream without a prefix length, terminated with `0xFF`.
```
japaric, line 880: "isn't a 0xFF delimiter problematic with characters like BOM / U+FEFF?"
"no, because UTF-8 encoding never uses the 0xff value; each byte (after encoding) contains at least one zero bit"
case in point, BOM is encoded as `[239, 187, 191]`

japaric (approved):
The implementation looks good to me.

The user interface seems reasonable. I like that a newtype is still needed. I personally prefer not to make super easy to plug in Debug / Display types into the logs as they are bandwidth expensive and Format should be preferred.

One thing that I'm not sure about is the adapters taking ownership of the inner field. It seems that in most cases you'll want to pass a reference as Debug/Display only need a reference to the value. I'm thinking of cases where you may write:
```
let x = NotCopyThing::new();
info!("{}", Debug2Format(x));
let y = x; // error: value moved in previous line
```
I think being uniform everywhere and always requiring a reference (rather than having the option to use `x` in some cases and `&x` in others) would avoid running into these unintentional moves.

> Display hints are, perhaps confusingly, ignored.

yeah, that's a bit unfortunate. Perhaps add a big warning to the API docs of the X2Format types? I don't think we can emit a warning or an error at call site (from the log macro)

Issue #341, opened 2021-01-14
Another bool compression bug


jonas-schievink:
```rust
#[entry]
fn main() -> ! {
    defmt::info!("{} {=bool}", Timestamp, true);

    loop {
        debug::exit(debug::EXIT_SUCCESS);
    }
}
struct Timestamp;

impl Format for Timestamp {
    fn format(&self, fmt: Formatter<'_>) {
        defmt::write!(fmt, "{=bool}", true);
    }
}
```

Expected output:

```
0.000000 INFO true true
```

Actual output:

```
0.000000 INFO false true
Error: couldn't decode all data (remaining: [1])
```

PR #342, opened 2021-01-14
Don't call `finalize` in `write!`

jonas-schievink:
`finalize` flushes accumulated `bool`s to the output stream. This should only happen once at the very end of a log frame, never within.

Fixes https://github.com/knurling-rs/defmt/issues/341


Lotterleben (approved):

jonas-schievink:
bors r=Lotterleben

Issue #106, opened 2020-08-11
Allow customizing/omitting the timestamp from the stream


japaric:
use cases:
- completely omitting the timestamps saves bandwidth.
- some people may want to use "human time" for the timestamps: (timezone+date+time).

the stream decoder needs to know before hand which binary format it will get (are timestamps included or not?)
one way to inform the decoder is to include a symbol with a know name in the ELF.
the address ("value") of this symbol indicates whether a "feature" is enabled (we do not need to constrain features to boolean like Cargo features though).
For example:
``` text
/* binfmt.x */
_BINFMT_TIMESTAMP = 0
```
could mean no timestamps.
a value of "1" could mean 64-bit absolute timestamps (today's default)

unresolved questions:
- are there any features, other than timestamps, that would benefit from this mechanism?
- are there other mechanism we could use to select the timestamp format, for example?

PR #343, opened 2021-01-14
Customizable timestamps

jonas-schievink:
Closes https://github.com/knurling-rs/defmt/issues/106

Based on https://github.com/knurling-rs/defmt/pull/342, since that was needed to fix the test in here.

This changes the `defmt::timestamp` proc macro to work just like the logging macros, but in item scope. This allows full customization of the timestamp, as well as "real" omission if no timestamp is used (currently we just always have a timestamp of 0).

## Summary of changes

* Add a dummy section to the linker script, whose purpose is just to hold symbols we want to retain but don't otherwise use. This is used to make the linker retain the timestamp format string, since we wouldn't otherwise use its address.
* `#[defmt::timestamp]` is now a regular proc macro instead of an attribute.
* A new `timestamp` tag was added to identify the format string used in `defmt::timestamp!`.
* The default timestamp is not present instead of being 0. elf2table just won't find a format string with `timestamp` tag and omit it accordingly.
* A new `µs` display hint was added to retain the previous time formatting. It interprets an integer as a time in microseconds and formats it like `X.XXXXXX`. We could potentially be more fancy here and allow arbitrary scaling of fixed-point decimal and binary values, but that seemed fairly complex, so I did this instead.

## Limitations

* If you want to use the old timestamp format, you'd use `{=u64:µs}`. However that is no longer LEB-encoded, so it takes up more space if it's a small number (eg. a monotonic counter). We could fix this by encoding integers with LEB128, or by adding a new `leb128` format type.

~~I probably still need to update the book and other docs.~~ Done!


Lotterleben:
regarding the book update, whst's left to do there? would it make sense to open an issue about that so it's transparent and we don't forget?

jonas-schievink:
I think that's all done now

japaric (approved):
This is awesome!

jonas-schievink:
bors r=japaric

Issue #75, opened 2020-08-06
Investigate efficiency of using leb64 for 32-bit values


jonas-schievink:
We currently encode string indices, timestamps, slice lengths and isize/usize with leb64, which makes use of 64-bit arithmetic. We should investigate whether using a 32-bit version for most of these (except timestamps, which are always up to 64 bits) makes formatting more efficient.

jonas-schievink:
After https://github.com/knurling-rs/defmt/pull/343 lands, we can switch leb64 out for leb32.

PR #345, opened 2021-01-15
Avoid 64-bit arithmetic in LEB encoding

jonas-schievink:
Fixes https://github.com/knurling-rs/defmt/issues/75

This uses `usize` instead of `u64` in `leb64`, which is now possible since timestamps aren't hard-coded to LEB64-encoded-`u64` anymore.

Before:

```
   text	   data	    bss	    dec	    hex	filename
  32084	      0	      8	  32092	   7d5c	../../target/thumbv7m-none-eabi/release/log
```

After:
```
   text	   data	    bss	    dec	    hex	filename
  31876	      0	      8	  31884	   7c8c	../../target/thumbv7m-none-eabi/release/log
```

So, ~200 bytes of code saved.

### DIFF
--- a/src/leb.rs
+++ b/src/leb.rs
@@ -3,6 +3,7 @@
 /// This handles 32-bit and 64-bit `usize`.
 pub(crate) fn leb64(x: usize, buf: &mut [u8; 10]) -> usize {
     let mut low = x as u32;
+    // Shift by 16 twice, to avoid a panic/error when shifting a 32-bit usize by 32 bits.
     let mut high = ((x >> 16) >> 16) as u32;
 
     let mut i = 0;
@@ -61,6 +62,8 @@ pub(crate) fn leb64(x: usize, buf: &mut [u8; 10]) -> usize {
     }
 }
 
+/// Encodes an `isize` as a `usize` by pulling its sign bit to the least-significant place (this
+/// makes it have an efficient LEB-encoding for small positive and negative values).
 pub fn zigzag_encode(v: isize) -> usize {
     const USIZE_BITS: usize = core::mem::size_of::<usize>() * 8;
     ((v << 1) ^ (v >> (USIZE_BITS - 1))) as usize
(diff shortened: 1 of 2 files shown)


src/leb.rs
```diff
@@ -58,8 +61,9 @@ pub(crate) fn leb64(x: u64, buf: &mut [u8; 10]) -> usize {
     }
 }
 
-pub fn zigzag_encode(v: i64) -> u64 {
-    ((v << 1) ^ (v >> 63)) as u64
+pub fn zigzag_encode(v: isize) -> usize {
```
Lotterleben, line 64: nit: a 1 sentence doc on what's happening here for those could be helpful who are not familiar with zigzag encoding (not blocking though)

src/leb.rs
```diff
@@ -86,26 +90,28 @@ mod tests {
         assert_eq!(buf[..i], [0x80, 1]);
         buf.iter_mut().for_each(|b| *b = 0x55);
 
-        let i = leb64((1 << 32) - 1, &mut buf);
-        assert_eq!(buf[..i], [0xff, 0xff, 0xff, 0xff, 0xf]);
-        buf.iter_mut().for_each(|b| *b = 0x55);
+        if cfg!(target_pointer_width = "64") {
```
Lotterleben, line 93: another nitpick: it's not very intuitive to figure out what's being tested here... Personally I'd put each check in a dedicated test case anyway (this way they're independent and their purpose is documented on the fly) but in this constellation comments would be helpful and time-saving

Lotterleben (approved):
heh, nice! nitpicks are non-blocking.

src/leb.rs
```diff
@@ -1,6 +1,9 @@
-pub(crate) fn leb64(x: u64, buf: &mut [u8; 10]) -> usize {
+/// LEB128-encodes a `usize` value into `buf`.
+///
+/// This handles 32-bit and 64-bit `usize`.
+pub(crate) fn leb64(x: usize, buf: &mut [u8; 10]) -> usize {
     let mut low = x as u32;
-    let mut high = (x >> 32) as u32;
+    let mut high = ((x >> 16) >> 16) as u32;
```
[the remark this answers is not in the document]
jonas-schievink: Oh, right, should've added a comment for that. Pushed a commit that explains this.

jonas-schievink:
bors r=Lotterleben

PR #347, opened 2021-01-18
Document current format parameter syntax

jonas-schievink:
Compare to https://doc.rust-lang.org/std/fmt/index.html#syntax

This is a helpful reference, and also helps determine what parts are still missing for `core::fmt` parity.

### DIFF
--- a/parser/src/lib.rs
+++ b/parser/src/lib.rs
@@ -62,8 +62,9 @@ pub enum Fragment<'f> {
 ///     'bool' | 'char' | 'u8' | 'u16' | 'u32' | 'u64' | 'u128' | 'usize' | 'i8' | 'i16' | 'i32' |
 ///     'i64' | 'i128 | 'isize' | 'f32'
 /// bitfield := integer '..' integer
-/// format-array := '[?;' ' '* integer ']'
-/// byte-array := '[u8;' ' '* integer ']'
+/// format-array := '[?;' spaces integer ']'
+/// byte-array := '[u8;' spaces integer ']'
+/// spaces := ' '*
 ///
 /// format_spec := type
 /// type := 'a' | 'b' | 'o' | 'x' | 'X' | '?' | 'µs'
--- a/parser/src/lib.rs
+++ b/parser/src/lib.rs
@@ -50,6 +50,24 @@ pub enum Fragment<'f> {
     Parameter(Parameter),
 }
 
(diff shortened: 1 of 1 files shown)


parser/src/lib.rs
```diff
@@ -50,6 +50,24 @@ pub enum Fragment<'f> {
...
+///
+/// ```notrust
+/// param := '{' [ argument ] [ '=' argtype ] [ ':' format_spec ] '}'
+/// argument := integer
+///
+/// argtype := bitfield | '?' | format-array | '[?]' | byte-array | '[u8]' | 'istr' | 'str' |
+///     'bool' | 'char' | 'u8' | 'u16' | 'u32' | 'u64' | 'u128' | 'usize' | 'i8' | 'i16' | 'i32' |
+///     'i64' | 'i128 | 'isize' | 'f32'
+/// bitfield := integer '..' integer
+/// format-array := '[?;' ' '* integer ']'
```
Lotterleben, line 65:
```suggestion
/// format-array := '[?;' ' ' * integer ']'
```

parser/src/lib.rs
```diff
@@ -50,6 +50,24 @@ pub enum Fragment<'f> {
...
+/// byte-array := '[u8;' ' '* integer ']'
```
Lotterleben, line 66:
```suggestion
/// byte-array := '[u8;' ' ' * integer ']'
```

Lotterleben (approved):
Oooh this is really neat. I'm wondering if this could also be a helpful addition to the book? not quite sure where to best put it though.

parser/src/lib.rs
jonas-schievink: Hmm, these are supposed to be [Kleene stars](https://en.wikipedia.org/wiki/Kleene_star), to denote zero or more repetitions of `' '` are allowed, so they shouldn't be spaced apart. Would it be clearer if I moved the `' '*` to a `spaces := ' '*` rule?

Lotterleben: ugh duh, shouldn't review stuff like this before my first coffee :D Yeah, the reason for my suggestion was that I had a hard time figuring out what the "opening" and "closing" ticks were. a `spaces := ' '*` rule would make this clearer imo.

jonas-schievink: Okay, did that

jonas-schievink:
bors r=Lotterleben

Issue #349, opened 2021-01-19
panic-probe should internally use the new Debug2Format adapter


japaric:
instead of formatting to a fixed size buffer

PR #350, opened 2021-01-19
panic-probe: use Display2Format adapter

jonas-schievink:
Fixes https://github.com/knurling-rs/defmt/issues/349


japaric (approved):
> +2 lines; -48 lines

:tada: awesome!

PR #351, opened 2021-01-19
Cleanup defmt tags

jonas-schievink:
This adds a `write` tag to identify format strings from user-defined `write!` invocations, and splits up the `fmt` tag into `prim` for primitives, and `derived` for `#[derive(Format)]`-generated strings. This allows tools to distinguish user-controlled format strings from generated ones.


japaric (approved):
Looks good. Thanks

PR #352, opened 2021-01-21
Do not display full version with --help

Javier-varez:


print/src/main.rs
```diff
@@ -92,15 +100,9 @@ fn main() -> anyhow::Result<()> {
...
-    let mut output = env!("CARGO_PKG_VERSION").to_string();
-
-    output.push_str("\nsupported defmt version: ");
-    output.push_str(defmt_decoder::DEFMT_VERSION);
-
-    // leak (!) heap memory to create a `&'static str` value. `String` won't work due to how
-    // structopt uses the clap API
-    // (this is only called once so it's not that bad)
-    Box::leak(Box::<str>::from(output))
+    println!("{}", env!("CARGO_PKG_VERSION").to_string());
```
japaric, line 105: could you include the tool name here as well? so it prints `defmt-print 0.1.0` on the first line.

japaric (commented):
Thanks for working on this! I left one request in the comments.

print/src/main.rs
Javier-varez: Sure, I will upload a change for it.

japaric (approved):
Thanks!

Issue #336, opened 2021-01-12
defmt crate doesn't compile on macOS


BriocheBerlin:
**Updated description**:
defmt 0.1.3 compiles on macOS but the git version (0.2.0) doesn't
repro:
``` console
$ cargo new --bin hello
$ cd hello
$ cargo add defmt --vers 0.1.3
$ cargo run
$ # ^ doesn't work if you use the git version
```
---

When trying to get started to write compile fail tests , running
`$ cargo test --test ui`
didn't work for me using a mac. I'd always get this unhelpful error:
```
LLVM ERROR: Global variable '{"package":"defmt","tag":"defmt_prim","data":"{=i8}","disambiguator":"2773326613612971761"}' has an invalid section specifier '.defmt.prim.{"package":"defmt","tag":"defmt_prim","data":"{=i8}","disambiguator":"2773326613612971761"}': mach-o section specifier requires a segment whose length is between 1 and 16 characters.
error: could not compile `defmt` 
```

Passing more flags like this did help:
`$ cargo test -p defmt --all-features --test ui`

This makes sense when looking at [ci.yml](https://github.com/knurling-rs/defmt/blob/main/.github/workflows/ci.yml#L50)

japaric:
@BriocheBerlin could you test this on macOS? I'd like to figure out if `defmt` fails to compile natively or if this problem is due to the single workspace configuration used in this repository.

``` console
$ cargo new --bin hello
$ cd hello
$ # or manually add 'defmt = "0.1.3"' to Cargo.toml
$ cargo add defmt 
$ # does this succeed?
$ cargo run
```

BriocheBerlin:
`$ cargo run` succeeded.

japaric:
@BriocheBerlin could you run `cargo test --test ui -j1` in this repo and report which crate fails to build? That command builds one crate at a time rather than in parallel.

BriocheBerlin:
So it seems that in the working version on the `hello` project the command for building the defmt crate is very similar. However it uses a different `src/lib.rs` file which is local `<more_path>/.cargo/registry/src/github.com-1ecc6299db9ec823/defmt-0.1.3/src/lib.rs`.

japaric:
hmm, maybe the macOS issue was introduced recently so it's in the git version but not on crates.io?
could you try these two:
``` console
$ cargo new --bin hello
$ cd hello

[...]

PR #357, opened 2021-01-25
fix issue #336

spookyvision:
Introduce special cased symbol names to satisfy the macos linker


macros/src/lib.rs
```diff
@@ -1039,9 +1050,25 @@ pub fn write(ts: TokenStream) -> TokenStream {
...
 }
 
+/// work around restrictions on length and allowed characters imposed by macos linker
+/// returns (note the comma character for macos):
+///   under macos: ".defmt," + 16 character hex digest of symbol's hash
+///   otherwise:   ".defmt." + prefix + symbol
+fn mksection(prefix: &str, symbol: &str) -> String {
+    let mut sub_section = format!(".{}{}", prefix, symbol);
+
+    if cfg!(target_os = "macos") {
```
jonas-schievink, line 1060: These reflect the system that the macro crate is built for, which is always the host system. We either need to check the `TARGET` env var (if it is even set), or emit `link_section` attributes for all target platforms at once, gated with `cfg_attr`.

spookyvision: …whoops. fixed?

macros/src/lib.rs
```diff
@@ -204,7 +197,8 @@ pub fn timestamp(ts: TokenStream) -> TokenStream {
                 // Unique symbol name to prevent multiple `timestamp!` invocations in the crate graph.
                 // Uses `#symname` to ensure it is not discarded by the linker.
                 #[no_mangle]
-                #[link_section = #section]
+                #[cfg_attr(target_os = "macos", link_section = ".defmt,end_timestamp")]
+                #[cfg_attr(not(target_os = "macos"), link_section = ".defmt.end_timestamp")]
```
jonas-schievink, line 201: This needs to use `end.timestamp` because the linker script special-cases the `end` subsection

macros/src/lib.rs
jonas-schievink: Yeah, seems good! We should add a test like described in https://github.com/knurling-rs/defmt/issues/336#issuecomment-761094732 though, to make sure this really fixed the bug and to prevent regressions.

spookyvision: already messaged Jorge about this 🙃

macros/src/lib.rs
spookyvision: for non-macos only though? Otherwise we need to change the special casing, macos won't allow any dots except the leading one

spookyvision: I spoke too soon - `.defmt,end.timestamp` seems fine under macos, unlike `.defmt.end.timestamp`...?

jonas-schievink: Ah, yeah, for non-macOS should be enough then

Issue #319, opened 2020-12-18
String fields in structs should be formatted using ':?' hint


japaric:
Current output:

``` rust
#[derive(Format)]
struct S { x: &'static str }

info!("{}", S { x: "hi" });
// -> INFO S { x: hi }
```

Desired output: `S  { x: "hi" }`

---

I wasn't sure how to best resolve this one.
In the RFC, we discussed applying the ':?' hint to all struct fields and that does the job in the previous example.
But seems like it would have the adverse effect of display hints never propagating into structs. Example below:

``` rust
// #[derive(Format)]
struct S { x: &'static str, y: u8 }

// derive would expand into
impl defmt::Format for S {
    fn format(&self, f: defmt::Formatter) {
        write!(f, "S {{ x: {=str:?}, y: {=u8:?} }}")
    }
}
```

``` rust
// matches the expected output
info!("{}", S { x: "hi", y: 42 });
// -> INFO S { x: "hi", y: 42 }

// `y` field ignores top-level `:x` display hint
info!("{:x}", S { x: "hi", y: 42 });
// -> INFO S { x: "hi", y: 42 }
```

The second statement ignores the ':x' hint.

We discussed in the RFC that the innermost display hint should not be overridden by an outer one.
Example from the RFC:

``` rust
struct S { x: u8 }
impl Format for S {
    fn format(&self, f: defmt::Formatter) {
        write!(f, "S {{ x: {=u8:x} }}", self.x)
    }
}

// binary hint should be ignored
info!("{:b}", S { x: 42 });
// -> INFO S { x: 0x2a }
```

Given that, using ':?' in fields would make non-string fields ignore display hints specified at the top call which is not what I would expect.

japaric:
Here's one way to make this work.

> We discussed in the RFC that the innermost display hint should not be overridden by an outer one.

We revise this part of the specification to say "innermost display hint should not be overridden by an **applicable** outer one" and introduce the idea of certain hints not being applicable to certain types.

For example, `:x`, `:b` and `:a` do *not* apply to type `=str`, but `:?` does (it escapes the string).
Similarly, `:?` does *not* apply to integer types like `=u8` but `:b` and `:x` do apply.

With that adjusted semantics we can tweak the derive macro to always add the `:?` hint to all struct / enum fields. Example:

``` rust
#[derive(Format)]
struct S { x: &'static str, y: u8 }

// derive would expand into
impl defmt::Format for S {
[...]

PR #359, opened 2021-01-27
Implement precedence of inner display hint

justahero:
This PR implements precedence of display hints inner types, e.g. `u8`, `u16` etc. It implements the display hint precedence rules given in #319.

When formating variables using the `defmt::Format` trait, either by implementing it or using the `derive` macro, the outer display hint is ignored depending on the inner type & its display hint.

For example, a struct that uses the `derive` marco

```rust
#[derive(Format)]
struct S { x: u8 }

// derive would expand to
impl defmt::Format for S {
    fn format(&self, f: defmt::Formatter) {
        write!(f, "S {{ x: {=u8:?} }}", self.x);
    }
}
```

This uses the `Debug` display hint (`:?`) for the inner type `x` of struct `S`. When displaying the above struct `S` with

```rust
info!("{:x}", S { x: 42 });
// -> INFO S { x: 0x2a }
```

the outer display hint `:x` is applied to the inner type `x`. In this case the outer display hint hexadecimal is used.

On the other hand if the inner type is one of `:x` (hex), `:b` (binary), the outer display hint will be ignored and the inner hint has precedence.

This time the implementation uses the `:b` (binary) display hint to format inner type `x`.

```rust
impl defmt::Format for S {
    fn format(&self, f: defmt::Formatter) {
        write!(f, "S {{ x: {=u8:b} }}", self.x);
    }
}
```

When logging the struct this time

```rust
info!("{:x}", S { x: 42 });
// -> INFO S { x: 0b101010 }
```

the outer display hint `:x` is ignored, the inner display hint `:b` is used, therefore the value `42` is formatted as a binary string `0b101010`.

Another precedence rule is related to `str` fields of structs. As given in the RFC #319, the following struct with an inner `str` field is deriving the `defmt::Format` trait as follows.

```rust
#[derive(Format)]
struct S { x: &'static str }

// derive now expands to
impl defmt::Format for S {
    fn format(&self, f: defmt::Formatter) {
        write!(f, "S {{ x: {=str:?} }}", self.x);
    }
}
```

The difference is `x: &'static str` is expanded to `{=str:?}` instead of a plain `{=str}`. This way string in `x` is enclosed by double quotes.

```rust
defmt::info!("{} world", S { x: "hello" });
// -> "hello" world
defmt::info("hello {:?}", S { x: "world" });
// -> hello "world"
```

Both display hints `{}` and `{:?}` now produce the same desired output.


macros/src/lib.rs
```diff
@@ -429,7 +429,11 @@ fn fields(
                         "?".to_string()
                     });
                     if let Some(ident) = f.ident.as_ref() {
-                        core::write!(format, "{}: {{={}}}", ident, ty).ok();
+                        if ty == "str" {
```
japaric, line 432: the problem with this approach (attaching `:?` only to the `=str` type) is that it won't work with generic structures. For example:
``` rust
#[derive(Format)]
struct S1<T> {
    x: T,
    y: u8,
}

// outputs: "S { x: hi, y: 42 }"
defmt::info!("{}", S1 { x: "hi", y: 42 });
// outputs: "S { x: hi, y: 0x2a }"
defmt::info!("{:x}", S1 { x: "hi", y: 42 });
```

This is because the expansion of the macro will use the format string `"S1 {{ x: {=?}, y: {=u8} }}"`. The proc-macro cannot know whether the generic type `T` will be instantiated to `str` or not. Also, you don't the `:?` format specifier if `T` is instantiated to a type that's not a string.

japaric (changes requested):

macros/src/lib.rs
justahero: Thanks for the feedback. The code is changed now to apply the Debug hint `:?` to all inner types. As a side effect this changes the output in the `log` test program as well.

```rust
#[derive(Format)]
struct Data<'a> {
    name: &'a [u8],
    value: bool,
}

let data = &[Data { name: b"Hi", value: true }];
defmt::info!("{=[?]:a}", *data);

// before the change
// -> INFO [Data { name: b"Hi", value: true }]
// after the change
// -> INFO [Data { name: [72, 105], value: true }]
```

firmware/qemu/src/bin/log.out
```diff
@@ -111,6 +111,6 @@
 0.000110 INFO b"Hi"
 0.000111 INFO b"Hi"
 0.000112 INFO [45054, 49406]
-0.000113 INFO [Data { name: b"Hi", value: true }]
+0.000113 INFO [Data { name: [72, 105], value: true }]
```
jonas-schievink, line 114: This slice is formatted with `{=[?]:a}`, why is the `:a` hint ignored?

firmware/qemu/src/bin/hints_inner.rs
```diff
@@ -0,0 +1,142 @@
...
+    }
+
+    {
+        #[derive(defmt::Format)]
+        struct S1<T> {
+            x: T,
+            y: u8,
+        }
+
+        // outputs: "S { x: hi, y: 42 }"
```
jonas-schievink, line 121:
```suggestion
        // outputs: "S { x: "hi", y: 42 }"
```

firmware/qemu/src/bin/hints_inner.rs
```diff
@@ -0,0 +1,142 @@
...
+        defmt::info!("{}", S1 { x: "hi", y: 42 });
+        // outputs: "S { x: hi, y: 0x2a }"
```
jonas-schievink, line 123:
```suggestion
        // outputs: "S { x: "hi", y: 0x2a }"
```

decoder/src/lib.rs
```diff
@@ -588,6 +588,159 @@ mod tests {
...
+            0,
+            TableEntry::new_without_symbol(Tag::Info, "x={:b}".to_owned()),
+        );
+        entries.insert(
+            1,
+            TableEntry::new_without_symbol(Tag::Derived, "S {{ x: {=u8:?} }}".to_owned()),
+        );
+
+        let table = Table {
+            entries,
```
Urhengulas, line 640: What do you think about this pattern? It would also be applicable for some other tests.
```suggestion
        let entries = vec![
            TableEntry::new_without_symbol(Tag::Info, "x={:b}".to_owned()),
            TableEntry::new_without_symbol(Tag::Derived, "S {{ x: {=u8:x} }}".to_owned()),
        ];

        let table = Table {
            entries: entries.into_iter().enumerate().collect::<BTreeMap<_, _>>(),
```

decoder/src/frame.rs
```diff
@@ -269,13 +269,24 @@ fn format_args_real(
...
                                 }
                             }
-                            _ => format_u128(*x as u128, hint, &mut buf)?,
+                            _ => {
+                                if let Some(DisplayHint::Debug) = hint {
+                                    format_u128(*x as u128, parent_hint, &mut buf)?;
+                                } else {
+                                    format_u128(*x as u128, hint, &mut buf)?;
+                                }
+                            }
```
Urhengulas, line 278:
```suggestion
                            _ => match hint {
                                Some(DisplayHint::Debug) => {
                                    format_u128(*x as u128, parent_hint, &mut buf)?
                                }
                                _ => format_u128(*x as u128, hint, &mut buf)?,
                            },
```

macros/src/lib.rs
```diff
@@ -410,7 +410,7 @@ fn fields(
                         "?".to_string()
                     });
                     if let Some(ident) = f.ident.as_ref() {
-                        core::write!(format, "{}: {{={}}}", ident, ty).ok();
+                        core::write!(format, "{}: {{={}:?}}", ident, ty).ok();
```
Urhengulas, line 413: Why is the `:?` needed?

Urhengulas (changes requested):
Looks generally good to me. I have a couple of questions (not all particular about this PR) and style suggestions:
1. Could and should we drop `:?` from our grammar, since "Both display hints {} and {:?} now produce the same desired output."? 
2. Does this PR respect all of `x`, `X`, `#x`, `b`?

decoder/src/frame.rs
```diff
@@ -269,13 +269,24 @@ fn format_args_real(
...
                         }
                     }
                     Arg::Ixx(x) => format_i128(*x as i128, hint, &mut buf)?,
                     Arg::Str(x) | Arg::Preformatted(x) => format_str(x, hint, &mut buf)?,
                     Arg::IStr(x) => format_str(x, hint, &mut buf)?,
-                    Arg::Format { format, args } => buf.push_str(&format_args(format, args, hint)),
+                    Arg::Format { format, args } => match parent_hint {
+                        Some(DisplayHint::Ascii) => {
```
japaric, line 285: I don't quite follow why `:a` is a special case here. (there some bugs / limitations (?) around `:a` and generics but this doesn't seem like it fixes them or creates more)

decoder/src/lib.rs
```diff
@@ -588,6 +588,159 @@ mod tests {
...
+        entries.insert(
+            1,
+            TableEntry::new_without_symbol(Tag::Derived, "S {{ x: {=str:?} }}".to_owned()),
+        );
+
+        let table = Table {
+            entries,
+            timestamp: Some(TableEntry::new_without_symbol(
+                Tag::Timestamp,
+                "{=u8:µs}".to_owned(),
```
japaric, line 678: (this may need to be changed to `:us` after a recent PR but bors will point it out I think)

japaric (approved):
this looks good to me. let's see if bors accepts it

japaric:
@Urhengulas 

> Could and should we drop :? from our grammar, since "Both display hints {} and {:?} now produce the same desired output."?

`{}` is Display-like format. `{:?}` is Debug-like format. This PR doesn't change that distinction. The following code prints the logs as desired (behavior matches `std`) even with this PR
``` rust
defmt::info!("the answer is {}", "42"); // INFO the answer is 42
defmt::info!("the answer is {:?}", "42"); // INFO the answer is "42"

```

japaric:
>failures:
    tests::display_inner_str_in_struct
    tests::display_u8_vec
    tests::display_use_inner_type_hint
    tests::display_use_outer_type_hint

those are the new tests. @justahero could you please update the timestamp format to match PR #522 ? feel free to `bors r+` this again after that

Urhengulas (approved):

PR #367, opened 2021-02-03
Satisfy clippy

Urhengulas:
Fix multiple clippy warnings:
* useless conversion to the same type: `u128`
* useless use of `format!`
* called `.as_ref().map(|s| &**s)` on an Option value. This can be done more directly by calling `mod_path.as_deref()` instead
* you seem to be trying to use match for destructuring a single pattern. Consider using `if let`

Surpress:
* returning an `Err(_)` with the `?` operator

### DIFF
--- a/print/src/main.rs
+++ b/print/src/main.rs
@@ -87,8 +87,7 @@ fn main() -> anyhow::Result<()> {
                 Err(defmt_decoder::DecodeError::UnexpectedEof) => break,
                 Err(defmt_decoder::DecodeError::Malformed) => {
                     log::error!("failed to decode defmt data: {:x?}", frames);
-                    #[allow(clippy::try_err)]
-                    Err(defmt_decoder::DecodeError::Malformed)?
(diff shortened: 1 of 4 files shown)


print/src/main.rs
```diff
@@ -92,6 +87,7 @@ fn main() -> anyhow::Result<()> {
                 Err(defmt_decoder::DecodeError::UnexpectedEof) => break,
                 Err(defmt_decoder::DecodeError::Malformed) => {
                     log::error!("failed to decode defmt data: {:x?}", frames);
+                    #[allow(clippy::try_err)]
```
jonas-schievink, line 90: I think I agree with Clippy here – this would be clearer if it used `return`.

jonas-schievink (approved):
Looks good! But we can fix the last warning too

print/src/main.rs
Urhengulas: Done!

PR #368, opened 2021-02-07
Raise compile error on armv6m

### LINKED ISSUE #365: defmt-itm does not compile for ARMv6-M
Those microcontrollers don't have an ITM peripheral. Instead of raising several, unintelligible errors when compiled we should use `compile_error!` to tell the user that this crate cannot be used on Cortex-M0(+) chips.

Urhengulas:
Fixes #365

### DIFF
--- a/firmware/defmt-itm/src/lib.rs
+++ b/firmware/defmt-itm/src/lib.rs
@@ -20,6 +20,9 @@ use core::{
 
 use cortex_m::{interrupt, itm, peripheral::ITM, register};
 
+#[cfg(armv6m)]
+compile_error! {"`defmt-itm` cannot be used on Cortex-M0(+) chips, because it requires an ITM peripheral"}
+
 static ENABLED: AtomicBool = AtomicBool::new(false);
 
 /// Enables defmt logging over the ITM stimulus port 0.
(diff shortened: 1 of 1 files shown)


src/lib.rs
```diff
@@ -26,6 +26,10 @@ use core::{fmt, ptr::NonNull};
 pub mod export;
 mod impls;
 mod leb;
+
+#[cfg(armv6m)]
+compile_error! {"defmt cannot be used on Cortex-M0(+) chips, because it requires an ITM peripheral"}
```
jonas-schievink, line 31: Did you mean to put this in the `defmt-itm` crate instead? defmt should work on thumbv6m without problems.

Urhengulas: Oh, yes 😓

firmware/defmt-itm/src/lib.rs
```diff
@@ -20,6 +20,9 @@ use core::{
 
 use cortex_m::{interrupt, itm, peripheral::ITM, register};
 
+#[cfg(armv6m)]
+compile_error! {"`defmt-itm` cannot be used on Cortex-M0(+) chips, because it requires an ITM peripheral"}
```
jonas-schievink, line 24:
```suggestion
compile_error!("`defmt-itm` cannot be used on Cortex-M0(+) chips, because it requires an ITM peripheral");
```

Urhengulas: My linter told me to use `{`, which I also didn't understood, but it actually also worked in some dummy code I tried.

jonas-schievink: Ah, maybe you forgot the trailing `;`? I generally use `{ }` if there's multi-line content inside, but `();` if not.

Issue #356, opened 2021-01-24
logger: don't eat all logs that don't match "probe-run"


Dirbaio:
https://github.com/knurling-rs/defmt/blob/a57f6c181d9a542c03560e45877d32a1b0f35076/logger/src/lib.rs#L83-L86

This is a problem for 2 reasons:

- in probe-run it's impossible to see probe-rs logs without patching. RUST_LOG doesn't work, --verbose doesn't work.
- If you intend defmt-logger to be usable as a library, users that are not probe-run will want to log their own stuff.

PR #376, opened 2021-02-08
Make defmt-logger more configurable, remove probe-run strings

jonas-schievink:
This refactors `defmt-logger` so that it does not make decisions about what data to include. It adds more API surface that can be used to write custom loggers, and removes any assumptions about being used by `probe-run` from the existing logger.

Closes https://github.com/knurling-rs/defmt/issues/356

### DIFF
--- a/decoder/src/logger/mod.rs
+++ b/decoder/src/logger/mod.rs
@@ -273,7 +273,8 @@ fn color_diff(text: String) -> String {
 /// printed.
 ///
 /// If `always_include_location` is `true`, a second line containing location information will be
-/// printed for *all* records, not just for defmt frames.
+/// printed for *all* records, not just for defmt frames (defmt frames always get location info
+/// included if it is available, regardless of this setting).
 pub fn init_logger(
     always_include_location: bool,
     should_log: impl Fn(&log::Metadata) -> bool + Sync + Send + 'static,
--- a/decoder/src/logger/mod.rs
+++ b/decoder/src/logger/mod.rs
@@ -12,11 +12,12 @@ use crate::decoder::Frame;
 use ansi_term::Colour;
 use colored::{Color, Colorize};
 use difference::{Changeset, Difference};
-use log::{Level, Metadata, Record};
+use log::{Level, Log, Metadata, Record};
 
 use std::{
     fmt::{self, Write as _},
-    io,
+    io::{self, Write},
+    sync::atomic::{AtomicUsize, Ordering},
 };
 
 const DEFMT_TARGET_MARKER: &str = "defmt@";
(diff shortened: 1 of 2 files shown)


decoder/src/logger/mod.rs
```diff
@@ -263,3 +264,105 @@ fn color_diff(text: String) -> String {
...
+    /// messages.
+    timing_align: AtomicUsize,
+}
+
+impl Log for Logger {
+    fn enabled(&self, metadata: &log::Metadata) -> bool {
+        (self.should_log)(metadata)
+    }
+
+    fn log(&self, record: &log::Record) {
```
Lotterleben, line 305: nit: imo it should be documented here that defmt `record`s go to stdout and non-defmt `record`s will be logged to stderr

decoder/src/logger/mod.rs
```diff
@@ -263,3 +264,105 @@ fn color_diff(text: String) -> String {
...
+
+/// Initializes a `log` sink that handles defmt frames.
+///
+/// Defmt frames will be printed to stdout, other logs to stderr.
+///
+/// The caller has to provide a `should_log` closure that determines whether a log record should be
+/// printed.
+///
+/// If `always_include_location` is `true`, a second line containing location information will be
+/// printed for *all* records, not just for defmt frames.
```
Lotterleben, line 276: nit: from these docs I would've interpreted that `always_include_location = false` deactivates locations from `defmt` logs as well but from the `log()` implementation it seems that it is always on?

decoder/src/logger/mod.rs
```diff
@@ -263,3 +264,105 @@ fn color_diff(text: String) -> String {
...
+                    .min_timestamp_width(min_timestamp_width)
+                    .print_colored(&mut sink)
+                    .ok();
+            }
+            None => {
+                // non-defmt logs go to stderr
+                let stderr = io::stderr();
+                let mut sink = stderr.lock();
+
+                let level_color = match record.level() {
```
Lotterleben, line 332: nit: the same conversion is done in `Printer.print_colored()`, could this be refactored into a helper function to prevent diverging designs in the future and also for brevity?

decoder/src/logger/mod.rs
```diff
@@ -263,3 +264,105 @@ fn color_diff(text: String) -> String {
...
+                    "{timestamp:>0$} {level:5} {args}",
+                    min_timestamp_width,
+                    timestamp = "(HOST)",
+                    level = record.level().to_string().color(level_color),
+                    args = record.args()
+                )
+                .ok();
+
+                if let Some(file) = record.file() {
+                    // NOTE will be `Some` if `file` is `Some`
```
Lotterleben, line 353:
```suggestion
                    // NOTE will always be `Some` if `file` is `Some`
```

decoder/src/logger/mod.rs
```diff
@@ -228,3 +264,105 @@ fn color_diff(text: String) -> String {
...
+                    sink,
+                    "{timestamp:>0$} {level:5} {args}",
+                    min_timestamp_width,
+                    timestamp = "(HOST)",
+                    level = record.level().to_string().color(level_color),
+                    args = record.args()
+                )
+                .ok();
+
+                if let Some(file) = record.file() {
```
Lotterleben, line 352: nit: looks pretty similar to the end of `print_colored()` too? Could this be deduplicated?

decoder/src/logger/mod.rs
```diff
@@ -63,98 +55,141 @@ pub fn log_defmt(
...
 
-    fn log(&self, record: &Record) {
-        if !self.enabled(record.metadata()) {
-            return;
+    pub fn line(&self) -> Option<u32> {
+        self.line
+    }
+
+    /// Returns a builder that can format this record for displaying it to the user.
+    pub fn print(&'a self) -> Printer<'a> {
```
Lotterleben, line 121: nit: `print()` sounds like it would print itself, maybe `printer()` would make it clearer that this is not the case?

decoder/src/logger/mod.rs
```diff
@@ -63,98 +55,141 @@ pub fn log_defmt(
...
+
+/// Printer for `DefmtRecord`s.
+pub struct Printer<'a> {
+    record: &'a DefmtRecord<'a>,
+    include_location: bool,
+    min_timestamp_width: usize,
+}
+
+impl<'a> Printer<'a> {
+    /// Whether to include location info (file, line) in the output.
```
Lotterleben, line 138:
```suggestion
    /// Configure whether to include location info (file, line) in the output.
```

decoder/src/logger/mod.rs
```diff
@@ -63,98 +55,141 @@ pub fn log_defmt(
...
+    ///
+    /// If `true`, an additional line will be included in the output that contains file and line
+    /// information of the logging statement. By default, this is `false`.
```
Lotterleben, line 141:
```suggestion
    /// information of the logging statement.
    /// By default, this is `false`.
```

Lotterleben (approved):

decoder/src/logger/mod.rs
jonas-schievink: The only public API for the logger is `init_logger`, which does document this

Lotterleben: ah nevermind me then! my bad

decoder/src/logger/mod.rs
jonas-schievink: Clarified the docs a bit

jonas-schievink:
bors r=Lotterleben

Issue #375, opened 2021-02-08
defmt-test: Result support and #[should_error]


japaric:
It would be nice if this was supported:

``` rust
#[test]
fn ok() -> Result<(), ()> {
    Ok(())
}

#[test]
#[should_error]
fn err() -> Result<(), ()> {
    Err(())
}
```

That is:
- being able to return a `Result` from unit tests. That way one can use the `?` operator instead of `unwrap`-ing all over the place. (To some degree because it may not always be possible to unify all error types in `no_std`, e.g. `anyhow` and similar don't work in `no_std` / w/o heap allocation)
- have tests that pass when they return an error. The `std` test harness has `#[should_panic]` but there's not `panic::catch_unwind` in `no_std` so catching the `Err` variant is as close as we can get to that.

PR #377, opened 2021-02-08
defmt-test: support returning `Result` from tests

jonas-schievink:
Closes https://github.com/knurling-rs/defmt/issues/375

This adds a public `TestOutcome` trait, which is implemented for `Result` and `()`. Tests are now allowed to return anything that implements `TestOutcome`.


spookyvision:
LGTM but I can only vouch for like 90% of it

firmware/defmt-test/README.md
```diff
@@ -110,16 +108,23 @@ mod tests {
...
 └─ test::tests::__defmt_test_entry @ tests/test.rs:11
-0.000002 INFO  running assert_flag ..
+0.000001 INFO  (2/2) running `assert_flag`...
 └─ test::tests::__defmt_test_entry @ tests/test.rs:11
-0.000003 INFO  .. assert_flag ok
+0.000002 INFO  all tests passed!
 └─ test::tests::__defmt_test_entry @ tests/test.rs:11
 ```
 
+## Test Outcome
```
japaric, line 119: :100: 
perhaps we should an example of this to app-template? it's a bit of 'duplicated documentation' but I tend to favor 'put docs right in front of the user'; that way 'didn't read the docs' issues become less common (hopefully)

firmware/qemu/src/bin/defmt-test.rs
```diff
@@ -1,25 +1,38 @@
...
 #[defmt_test::tests]
 mod tests {
-    use super::MAX;
+    use core::u8::MAX;
+    use defmt::{assert, assert_eq};
+
     #[test]
-    fn assert_true() {
-        defmt::assert!(true)
+    fn assert_true() -> () {
```
japaric, line 12: :+1:

firmware/defmt-test/src/lib.rs
```diff
@@ -8,6 +8,30 @@
...
+/// Private implementation details used by the proc macro.
+#[doc(hidden)]
 pub mod export;
+
+/// Types indicating whether a test succeeded or failed.
+///
+/// This is comparable to the `Termination` trait in libstd, except stable and tailored towards the
+/// needs of defmt-test. It can be implemented by hand if desired, though normal usage shouldn't
+/// require that.
+pub trait TestOutcome: Format {
```
japaric, line 23: unsure if we want to make this public right off the bat I would error on making it a sealed trait first.
if / when we want to make it public perhaps we should put the trait in the `defmt` crate because `defmt-test` depends on the `cortex-m` crate which is going to limit which crates can implement the trait.

japaric (approved):
Looks great! Left a question about the trait and some comments about follow up work.

firmware/defmt-test/Cargo.toml
```diff
@@ -18,3 +18,12 @@ defmt-test-macros = { version = "=0.1.1", path = "macros" }
...
 version = "0.1.0"
+
+[features]
+defmt-default = []
+defmt-trace = []
+defmt-debug = []
+defmt-info = []
+defmt-warn = []
+defmt-error = []
+default = ["defmt-trace"] # enable everything
```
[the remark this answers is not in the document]
jonas-schievink: That would turn off defmt in the testsuite itself, I think we should keep it

firmware/defmt-test/src/lib.rs
jonas-schievink: Yeah, sealed it for now

firmware/defmt-test/Cargo.toml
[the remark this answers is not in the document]
japaric: you are correct. never mind then

PR #380, opened 2021-02-09
Merge crates "elf2table" and "logger" into "decoder"

spookyvision:


jonas-schievink:
This looks good. We should probably revisit the API we expose though.

bors r+

Issue #381, opened 2021-02-10
defmt-test: support the #[ignore] attribute & raise error on unknown attributes


japaric:
The standard test runner uses this attribute to ignore (omit running) a unit test. We implement the same functionality or, at least, error when an unknown attribute is used within `defmt_test::tests` because right now `cargo test`-ing this:
``` rust
#[ignore]
#[test]
fn loopback() { /* .. */ }
```
does run the test marked as `#[ignore]`

japaric:
I did `cargo expand` and it looks like we are discarding all the attributes that appear in the input code.

japaric:
This doesn't work either:
``` rust
#[cfg(ignore)]
#[test]
fn loopback() { /* .. */ }
```
in the sense that the test is compiled and executed when `--cfg ignore` is *not* passed to rustc.

jonas-schievink:
This now behaves as expected, do we want to close it, or leave it open to track `#[ignore]` support?

japaric:
I'll open a new issue about the `#[ignore]` attribute.

PR #383, opened 2021-02-10
defmt-test: Modify attributes in place and handle #[cfg]

jonas-schievink:
Closes (?) https://github.com/knurling-rs/defmt/issues/381 (but doesn't add `#[ignore]` support)

This changes the defmt-test macro to retain the `ItemFn`s and selectively delete defmt-test-specific attributes from them. The result is that the code is now simpler and arbitrary Rust attributes can be attached to tests. Doing so with `#[ignore]` gives the expected "unused attribute" warning.

### DIFF
--- a/firmware/defmt-test/macros/src/lib.rs
+++ b/firmware/defmt-test/macros/src/lib.rs
@@ -167,7 +167,6 @@ fn tests_impl(args: TokenStream, input: TokenStream) -> parse::Result<TokenStrea
         let should_error = test.should_error;
         let ident = &test.func.sig.ident;
         let span = test.func.sig.ident.span();
-        let cfgs = &test.cfgs;
         let call = if let Some(input) = test.input.as_ref() {
             if let Some(state) = &state_ty {
                 if input.ty != **state {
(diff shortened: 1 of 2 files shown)


firmware/defmt-test/macros/src/lib.rs
```diff
@@ -200,19 +188,23 @@ fn tests_impl(args: TokenStream, input: TokenStream) -> parse::Result<TokenStrea
...
+    let test_functions = tests.iter().map(|test| &test.func);
     let unit_test_running = tests
         .iter()
         .enumerate()
-        .map(|(i, test)| format!("({}/{}) running `{}`...", i + 1, tests.len(), test.ident))
+        .map(|(i, test)| {
+            format!(
+                "({}/{}) running `{}`...",
+                i + 1,
+                tests.len(),
```
japaric, line 204: the number of tests that run are not known in this context because cfg are evaluated after expansion

firmware/qemu/src/bin/defmt-test.rs
```diff
@@ -18,11 +21,18 @@ mod tests {
...
 
+    #[cfg(not(never))]
     #[test]
     fn result() -> Result<(), ()> {
         Ok(())
     }
 
+    #[cfg(never)]
+    #[test]
+    fn doesnt_compile() {
```
japaric, line 32: the QEMU output includes `INFO (4/6) running `doesnt_compile`...` but shouldn't

japaric (approved):
The overall approach looks good to me but the log messages that report which test is run next should also be `cfg`-ed

firmware/defmt-test/macros/src/lib.rs
jonas-schievink: Yeah, I changed it so it gets computed at runtime (or in a `const`) in adce0f3

PR #384, opened 2021-02-11
Pin unstable path dependencies

### LINKED ISSUE #378: pin dependency versions
all dependencies that are in this repo ('path' dependencies) and have the "unstable" feature (e.g. defmt-decoder) should be specified in Cargo.toml using an exact version, e.g. `=0.1.2`.
that way we'll be able to make breaking changes in the "unstable" API in patch versions

**NOTE** this also needs to be in `probe-run`

Urhengulas:
Fixes #378


jonas-schievink (approved):
bors r+

PR #386, opened 2021-02-12
Tidy defmt-decoder

Urhengulas:
This PR
* moves the `cfg`-attributes from modules to crate root
* rename `mod logger` to `log`


decoder/src/decoder.rs
```diff
@@ -4,10 +4,6 @@
 //! not meant to be consumed by other tools at the moment so all the API is unstable.
 
 // NOTE: always runs on the host
```
jonas-schievink, line 6: This note can also be moved to lib.rs

Urhengulas:
@jonas-schievink What do you think of the dissolving? I hope it reduces a bit of complexity in the imports and file structure.

decoder/src/elf2table/mod.rs
```diff
@@ -12,8 +12,8 @@ use std::{
     path::{Path, PathBuf},
 };
 
-pub use crate::decoder::Table;
-use crate::decoder::{StringEntry, TableEntry, Tag};
+pub use crate::Table;
```
jonas-schievink, line 15: This can be made private, there's no longer any reason to reexport it

```suggestion
use crate::Table;
```

PR #389, opened 2021-02-15
Update decoder dependencies

jonas-schievink:
This deduplicates `object` and `gimli`, once probe-run has been
updated after this


japaric (approved):
:+1:

PR #391, opened 2021-02-15
348 format for duration

### LINKED ISSUE #348: implement `Format` for `core::time::Duration`
it would be nice if this type was supported

Urhengulas:
This implements the `trait defmt::Format for core::time::Duration` and also adapts `defmt_macros::write!` to achieve this.

Fixes #348


jonas-schievink (commented):
bors r+

PR #392, opened 2021-02-16
Tidy up decoder 2

Urhengulas:
This PR does actually quite some restructuring of the `defmt_decoder`. The main ones are:
* make `fn parse_*` and `fn get_locations` methods of `struct Table`
* move `struct Frame` and `struct Decoder` into own private module each to make code easier digestable
* various simplifications

## open questions
* First question: What do you thing of splitting it up like this?
* Second question: Should `fn decode` maybe also become a method of `Table`? 

The tests can likely also get split up more into each submodule, but I'd do this in another PR.


decoder/src/elf2table/mod.rs
```diff
@@ -1,4 +1,4 @@
-//! Reads ELF metadata and builds a [`defmt`](https://github.com/knurling-rs/defmt) interner table.
+//! Reads ELF metadata and builds a [`defmt`](https://github.com/knurling-rs/defmt) internal table.
```
jonas-schievink, line 1: "interner" was intentional here, but maybe we can clarify this:

```suggestion
//! Reads ELF metadata and builds a table containing [`defmt`](https://github.com/knurling-rs/defmt) format strings.
```

jonas-schievink (commented):
Looks good!

jonas-schievink:
> Second question: Should `fn decode` maybe also become a method of `Table`?

Yeah, this definitely makes sense to me

decoder/src/elf2table/mod.rs
Urhengulas: Ups. Though this was a typo 🙈

Urhengulas:
I've added `decode` to `Table` and included your suggestion. Are we good to go?

jonas-schievink:
Yeah, looks good!

bors r+

PR #393, opened 2021-02-17
implement CI tests as xtask

### LINKED ISSUE #366: turn CI logic into xtasks
would be nice if there were xtasks to run the CI tests locally.
right now CI logic is in a yaml file that contains bash-like stuff and the tests.sh bash file (which won't run on Windows)
Ideally I would like the GHA yaml be more or less just a call of different xtasks
``` yaml
# apt-get install stuff has been omitted
jobs:
  host:
    # omitted: OS matrix
    name: run unit tests on the host
    run: cargo xtask test host

  cross:
    name: check cross compilation
    run: cargo xtask test cross

  snapshot:
    name: run snapshot tests (QEMU)
    run: cargo xtask test snapshot

  book:
    name: run mdbook tests
    run: cargo xtask test book

  lint:
    name: run rustfmt and clippy
    run: cargo xtask test lint
```

and then a single xtask, e.g. `cargo xtask test all`, that sort of runs all the above jobs which is what contributors would run locally to check their branch / PR is "green".
the xtasks should auto-install the cross compilation targets (`rustup target add`) so that contributors don't need to do "setup" steps
(maybe the xtasks should uninstall the cross compilation targets that were not there when they exit, regardless of the exit code)

spookyvision:
this xtask implementation fixes #366 and
- collects all test errors in one grand result vec and displays them at the end
- has commands for `test-all`, `test-host`, `test-cross` etc.
- has `-k` switch to keep targets that were installed as required dependency
- displays colored diff for failed snapshot tests
- puts "macos book build fixes" in separate commit ( c847772 ) - please review
- updates the ci yaml to use this xtask

not done:
- install os specific packages like `libusb` (out of scope, IMO)


xtask/src/main.rs
```diff
@@ -0,0 +1,607 @@
...
+
+    match rustc_is_nightly() {
+        Ok(is_nightly) => {
+            if is_nightly {
+                tests.push("alloc");
+            }
+        }
+        Err(e) => {
+            eprintln!(
+                "could not determine whether rust compiler is nightly - assuming it's not ({})",
```
jonas-schievink, line 533: You can unwrap this result – if we cannot run `rustc` something is very wrong

xtask/src/main.rs
```diff
@@ -0,0 +1,607 @@
...
+        TestCommand::TestBook => {
+            test_book(&mut all_errors);
+        }
+        TestCommand::TestLint => {
+            test_lint(&mut all_errors);
+        }
+    }
+
+    if !opt.keep_targets {
+        uninstall_targets(added_targets);
```
jonas-schievink, line 597: Ah, I see, this is why you query the targets. Seems fine then, but I don't think this functionality is necessary.

xtask/src/main.rs
spookyvision: aye, I wasn't sure about this one either. Mainly kept my approach to *always* be able to uninstall the previously installed targets. But it's such a strange edge case that people will probably not complain about lingering targets ever.

xtask/src/main.rs
spookyvision: I can still `" "`join them for installing though :D Rust is *all about efficiency* after all

spookyvision: (Optional uninstall was suggested by Jorge and from my POV it was a quick+easy win, so I added it)

xtask/src/main.rs
```diff
@@ -0,0 +1,607 @@
...
+}
+
+fn test_host(errors: &mut Vec<String>) {
+    println!("🧪 host");
+    do_test(
+        || {
+            run_command::<&str>(
+                &["cargo", "check", "--workspace"],
+                None,
+                &[("RUSTFLAGS", "--deny warnings")],
```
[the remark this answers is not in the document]
spookyvision: is the ultimate goal to use this xtask approach in CI as well? in that case I'd add a cmdline switch, something like `--strict-clippy`?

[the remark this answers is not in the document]
jonas-schievink: The main goal is to rewrite the CI configuration as an xtask, yes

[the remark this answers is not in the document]
spookyvision: OK - I'll switch this PR to draft mode then and ping again when I've migrated the CI yaml as well.

xtask/src/main.rs
```diff
@@ -0,0 +1,605 @@
...
+            || test_single_snapshot(test, features, true),
+            "qemu/snapshot",
+            errors,
+        );
+    }
+}
+
+fn main() -> Result<(), Vec<String>> {
+    let opt: Options = Options::from_args();
+    let mut all_errors: Vec<String> = vec![];
```
Urhengulas, line 559: Since all the code has access to `all_errors` anyways: What do you think about making it a global?
```rust
static mut all_errors: Vec<String> = vec![];
```

xtask/src/main.rs
```diff
@@ -0,0 +1,605 @@
...
+        }
+    }
+
+    if !opt.keep_targets {
+        uninstall_targets(added_targets);
+    }
+
+    if !all_errors.is_empty() {
+        eprintln!();
+        eprintln!("😔 some tests failed");
```
Urhengulas, line 600:
```suggestion
        eprintln!("\n😔 some tests failed");
```

xtask/src/main.rs
```diff
@@ -0,0 +1,605 @@
...
+}
+
+fn install_targets() -> Result<Vec<String>> {
+    let required_targets = vec![
+        "thumbv6m-none-eabi",
+        "thumbv7m-none-eabi",
+        "thumbv7em-none-eabi",
+        "thumbv8m.base-none-eabi",
+        "riscv32i-unknown-none-elf",
+    ];
```
Urhengulas, line 218: make `required_targets` a `const [&str; 5]`?

xtask/src/main.rs
```diff
@@ -0,0 +1,605 @@
...
+    let all_targets = required_targets
+        .iter()
+        .map(|item| item.to_string())
+        .collect::<HashSet<_>>();
```
Urhengulas, line 222: Is this conversion really needed? I assume this is in order to be able to use `.difference`, right?

Either we can keep `required_targets` a `let` and call `.to_string()` during initialization, or we figure out if `.difference(...)` can be called on `&str` too.

xtask/src/main.rs
```diff
@@ -0,0 +1,605 @@
...
+    if actual_matches_expected {
+        Ok(())
+    } else {
+        Err(anyhow!("{}", display_name))
+    }
+}
+
+fn get_installed_targets() -> Result<HashSet<String>> {
+    let out = run_capturing_stdout(Command::new("rustup").args(&["target", "list"]))?;
+    let installed_marker = " (installed)";
```
Urhengulas, line 201: make `installed_marker` a `const`?

xtask/src/main.rs
Urhengulas: on another note: iirc there is some project providing macros to create hashmaps and hashsets ala `vec!`. maybe an idea for future readability improvement :)

xtask/src/main.rs
```diff
@@ -0,0 +1,605 @@
...
+    do_test(
+        || run_command::<&str>(&["cargo", "check", "--workspace"], None, &env),
+        "host",
+        errors,
+    );
+
+    do_test(
+        || {
+            run_command::<&str>(
+                &["cargo", "check", "--all", "--features", "unstable-test"],
```
Urhengulas, line 340: `cargo --all` and `cargo --workspace` are alias, right? let's stick to one of them? (you used `--all` above).

xtask/src/main.rs
```diff
@@ -0,0 +1,605 @@
...
+        || {
+            run_command::<&str>(
+                &["cargo", "check", "--all", "--features", "alloc"],
+                None,
+                &env,
+            )
+        },
+        "host",
+        errors,
+    );
```
Urhengulas, line 359: This pattern seems to take up quite some space.. Can we make it a bit less spacious?

Maybe:
```rust
    let cmd_and_args = &["cargo", "check", "--all", "--features", "alloc"];
    do_test(
        || run_command::<&str>(cmd_and_args, None, &env),
        "host",
        errors,
    );
```

xtask/src/main.rs
```diff
@@ -0,0 +1,605 @@
...
+}
+
+fn test_cross(errors: &mut Vec<String>) {
+    println!("🧪 cross");
+    let targets = vec![
```
Urhengulas, line 400: make `target` a `const`?

Urhengulas (changes requested):
Hii, looks good to me overall. Just a couple of nits, which also could be only personal preferences:
* i generally like to have things in order, either logically or alphabetically, which in this case applies to dependencies, enum variants etc.; i added quite some suggestions to have that
* for my understanding of foreign code it helps me to start from the top-level and then go deeper into the details where neccessary. therefore i like to have e.g. the `fn main` first (in this case after `Options` and `TestComand`). What do you think?

.github/workflows/ci.yml
```diff
@@ -26,89 +26,53 @@ jobs:
...
-  rustfmt:
-    runs-on: ubuntu-latest
+  lint:
+    strategy:
+      matrix:
+        rust:
+          - stable
+        os:
+          - ubuntu-latest
+          - macOS-latest
```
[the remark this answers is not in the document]
Urhengulas: If you want to run on all platforms, but only stable you can probably replace the whole strategy matrix with
```yml
if: matrix.rust == stable
```

Else if you only want one platform
```yml
if: matrix.rust == stable && matrix.os == ubuntu
```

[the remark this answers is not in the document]
Urhengulas: Using this you can avoid having multiple strategy matrices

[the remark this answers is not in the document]
Urhengulas: `jobs.<job_id>.if`: https://docs.github.com/en/actions/reference/workflow-syntax-for-github-actions#jobsjob_idif

xtask/src/main.rs
```diff
@@ -0,0 +1,605 @@
...
+        }
+        TestCommand::TestSnapshot => {
+            test_snapshot(&mut all_errors);
+        }
+        TestCommand::TestBook => {
+            test_book(&mut all_errors);
+        }
+        TestCommand::TestLint => {
+            test_lint(&mut all_errors);
+        }
```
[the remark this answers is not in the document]
spookyvision: rustfmt (which I run on save) will break this into multiple lines again

xtask/src/main.rs
spookyvision: it *does* make the code a lot shorter, but also requires `unsafe {}` :D 
I think it's one of those rare cases where global mutable state could be OK, so I've gone ahead and wrapped it in a `once_cell::sync::Lazy`

xtask/src/main.rs
spookyvision: hm, personal style I guess, I generally prefer one line of output per `whatever_ln()` function. But I don't feel very strongly about that either.

xtask/src/main.rs
spookyvision: I prefer the flexibility of a `vec` (in case new targets come in) and consider the runtime overhead negligible. Again no super strong preference though.

xtask/src/main.rs
spookyvision: ultimately this function needs to return a bunch of owned strings, feel free to suggest an alternative for making them. I'd aim for "less lines of code and less code in those lines", whatever satisfies that is fine with me

xtask/src/main.rs
spookyvision: I'm not A BIG FAN OF CAPS IN CODE but I guess that can be done

xtask/src/main.rs
spookyvision: I've (hopefully! please double check!) copied all the commands from the old `ci.yml` verbatim

xtask/src/main.rs
spookyvision: `rustfmt` won't

spookyvision: hm. I like consistency, which means here to introduce a temp variable for every `do_test` call. That's also verbose in its own way (and harder to type when adding a test).

xtask/src/main.rs
spookyvision: can't allocate in constants

spookyvision:
> Hii, looks good to me overall. Just a couple of nits, which also could be only personal preferences:
> 
>     * i generally like to have things in order, either logically or alphabetically, which in this case applies to dependencies, enum variants etc.; i added quite some suggestions to have that

sure, accepted all of those

> 
>     * for my understanding of foreign code it helps me to start from the top-level and then go deeper into the details where neccessary. therefore i like to have e.g. the `fn main` first (in this case after `Options` and `TestComand`). What do you think?

hah, I'm exactly the opposite, I prefer to have stuff ordered like a dependency tree, so for me `main()` comes after all the stuff it needs to function. I don't see a possible consensus here 😆 but also I'm kinda "whatever" regarding this. Is there maybe a best practive in the Rust community? (my preference might after all simply come from C, where you *have* to put things in dependency tree order)

xtask/src/main.rs
Urhengulas: Even though I brought it up, I am actually very unsure about this. Globals always feel like a bad design, but I'd agree that it seems to make sense in this case 🤔

xtask/src/main.rs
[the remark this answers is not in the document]
Urhengulas: I don't think so. At least for me `rustfmt` accepts this. The crucial change is removing the `;`..

xtask/src/main.rs
Urhengulas: just stick with it then 😄

xtask/src/main.rs
Urhengulas: actually `let` is also fine in rust, since it is immutable, but somehow `const` still communicates this better for me 🤷🏾‍♀️

xtask/src/main.rs
Urhengulas: I agree that the separate variable is also bad. But there has to be a better way! :P

xtask/src/main.rs
Urhengulas: I don't really see why we need the runtime-flexibility of `Vec`. When adding targets we can just add it to the array, but you are right that this then would require to adapt the length in the type, when making it a `const`.

What do you think of:
```rust
let required_targets = [
    "thumbv6m-none-eabi",
    "thumbv7m-none-eabi",
    "thumbv7em-none-eabi",
    "thumbv8m.base-none-eabi",
    "riscv32i-unknown-none-elf",
]
.iter()
.map(|item| item.to_string())
.collect::<HashSet<_>>();

let installed_targets = get_installed_targets()?;
let added_targets = all_targets
    .difference(&installed_targets)
    .cloned()
    .collect::<Vec<_>>();
```
I realized that the intermediate `required_targets` isn't really needed, so we can merge the `all_targets` into it.

xtask/src/main.rs
```diff
@@ -0,0 +1,599 @@
...
+                &[],
+            )
+        },
+        "cross",
+        errors,
+    );
+
+    do_test(
+        || {
+            run_command::<&str>(
```
Urhengulas, line 500: Not needed since it can be inferred
```suggestion
            run_command(
```

Urhengulas (changes requested):

xtask/src/main.rs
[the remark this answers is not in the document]
spookyvision: ah, yes, ok

xtask/src/main.rs
spookyvision: ... `#[rustfmt::skip]`? 😬 (I'm only half joking)

Urhengulas: Naah 🤣

xtask/src/main.rs
```diff
@@ -0,0 +1,589 @@
...
+    TestHost,
+    TestLint,
+    TestSnapshot,
+}
+
+fn run_command<P: AsRef<Path>>(
+    cmd_and_args: &[&str],
+    cwd: Option<P>,
+    env: &[(&str, &str)],
+) -> Result<()> {
```
Urhengulas, line 49: As far as I can see the generic parameter is actually not needed. (You need to adapt all calls to `run_command` and remove `::<&str>`).

```suggestion
fn run_command(cmd_and_args: &[&str], cwd: Option<&str>, env: &[(&str, &str)]) -> Result<()> {
```

xtask/src/main.rs
spookyvision: I mean, you can annotate on block level, so ... it's not a _terrible_ idea.

xtask/src/main.rs
spookyvision: yeah, I mean I wanted to have the flexibility to pass `cwd`s in any useful form, but when I made `env` generic too that became too cumbersome, so I took *that* out again ... in the end we're probably fine with no generics here.

xtask/src/main.rs
Urhengulas: See suggestion here: https://github.com/knurling-rs/defmt/pull/393#discussion_r596423372

xtask/src/main.rs
Urhengulas: and we can add that if we actually need it. and if we don't add it we probably don't need it 🤷🏾‍♂️

xtask/src/main.rs
```diff
@@ -0,0 +1,559 @@
...
+    let cwd_s: String;
+    match cwd {
+        Some(path_ref) => {
+            cmd = cmd.current_dir(path_ref);
+            cwd_s = format!("CWD:{} ", path_ref);
+        }
+        None => {
+            cwd_s = "".to_string();
+        }
+    }
```
Urhengulas, line 69: `if let` this?

Urhengulas (changes requested):

xtask/src/main.rs
spookyvision: `Vec` flexibility: I was strictly thinking in terms of code length (and hassle in case of changes), but it turns out I was having a brain fart moment when I thought a slice must be of compile time fixed size here. So yeah, let's get rid of superfluous `Vecs`!

spookyvision: also, we're now one line shorter with your solution, you win 😆

xtask/src/main.rs
spookyvision: `const` *does* indicate that it's fixed at compile time, so yeah

xtask/src/main.rs
Urhengulas: This actually isnt a blocking issue. Lets just keep it like it is, and fix it if we find a good solution and not the hacka we were thinking off.
(btw: increasing the rustfmt max_width to 115 flattens many of the tests :D)

spookyvision: I actually like this approach a lot. Have committed a corresponding `rustfmt.toml` and the reformatted code

Urhengulas: people with small screens will hate us, but yes, seems like the best workaround for now 😆

xtask/src/main.rs
```diff
@@ -0,0 +1,479 @@
...
+        },
+        "book",
+    );
+}
+
+fn test_lint() {
+    println!("🧪 lint");
+    do_test(|| run_command(&["cargo", "clean"], None, &[]), "lint");
+    do_test(
+        || run_command(&["cargo", "fmt", "--all", "--", "--check"], None, &[]),
```
Urhengulas, line 262:
```suggestion
        || run_command(&["cargo", "fmt", "--workspace", "--", "--check"], None, &[]),
```

xtask/src/main.rs
```diff
@@ -0,0 +1,479 @@
...
+    let env = if deny_warnings {
+        vec![("RUSTFLAGS", "--deny warnings")]
+    } else {
+        vec![]
+    };
+
+    do_test(|| run_command(&["cargo", "check", "--workspace"], None, &env), "host");
+
+    do_test(
+        || run_command(&["cargo", "check", "--all", "--features", "unstable-test"], None, &env),
```
Urhengulas, line 281:
```suggestion
        || run_command(&["cargo", "check", "--workspace", "--features", "unstable-test"], None, &env),
```

xtask/src/main.rs
```diff
@@ -0,0 +1,479 @@
...
+        "host",
+    );
+
+    do_test(
+        || run_command(&["cargo", "check", "--all", "--features", "alloc"], None, &env),
```
Urhengulas, line 286:
```suggestion
        || run_command(&["cargo", "check", "--workspace", "--features", "alloc"], None, &env),
```

Urhengulas (changes requested):

xtask/src/main.rs
spookyvision: play small screen games, win small font prizes

Urhengulas (approved):

xtask/src/main.rs
```diff
@@ -0,0 +1,477 @@
...
+        "hints",
+    ];
+
+    if rustc_is_nightly() {
+        tests.push("alloc");
+    }
+
+    let mut features_map = HashMap::new();
+    features_map.insert("alloc", "alloc");
+    let no_features = "";
```
Urhengulas, line 434: `const` 🎉

xtask/src/main.rs
```diff
@@ -0,0 +1,477 @@
...
+
+    for test in &tests {
```
Urhengulas, line 436:
```suggestion
    for test in tests {
```

Urhengulas (commented):
I checked if you missed to transfer any tests, but seems good to me.

`fn test_single_snapshot` looks a bit messy to me, but from what I understand it does the same as `firmware/qemu/test.sh`.

One question I am unsure about: Why is the `--deny-warnings` only looked at in `test_host` and not the other tests as well?

spookyvision:
> I checked if you missed to transfer any tests, but seems good to me.
> 
> `fn test_single_snapshot` looks a bit messy to me, but from what I understand it does the same as `firmware/qemu/test.sh`.
> 
> One question I am unsure about: Why is the `--deny-warnings` only looked at in `test_host` and not the other tests as well?

@jonas-schievink what do you think?

xtask/src/main.rs
```diff
@@ -0,0 +1,477 @@
...
+
+    let content = fs::read_to_string(path)
+        .with_context(|| format!("Failed to load expected output data from {}", path_str))?;
+    Ok(content)
+}
+
+fn test_single_snapshot(name: &str, features: &str, release_mode: bool) -> Result<()> {
+    let display_name = format!("{} ({})", name, if release_mode { "release" } else { "dev" });
+    println!("{}", display_name);
+    let cwd_name = "firmware/qemu".to_string();
```
jonas-schievink, line 134: `.to_string()` shouldn't be needed here

```suggestion
    let cwd_name = "firmware/qemu";
```

jonas-schievink:
`test_single_snapshot` seems okay for now, we can clean it up later though.

> One question I am unsure about: Why is the `--deny-warnings` only looked at in `test_host` and not the other tests as well?

I don't think there was a particular reason we did this, but if there's a `.cargo/config.toml` then `RUSTFLAGS` overrides that, which would break the build, so we can't always set it.

Urhengulas:
> `test_single_snapshot` seems okay for now, we can clean it up later though.
> 
> > One question I am unsure about: Why is the `--deny-warnings` only looked at in `test_host` and not the other tests as well?
> 
> I don't think there was a particular reason we did this, but if there's a `.cargo/config.toml` then `RUSTFLAGS` overrides that, which would break the build, so we can't always set it.

I don't really question the existence of the cmdline-arg, but rather why it is only taken into account for some tests and not also e.g. the `test_cross`.

xtask/src/main.rs
```diff
@@ -0,0 +1,477 @@
...
+    let child = cmd.stdout(Stdio::piped()).spawn()?;
+    let mut stdout = child
+        .stdout
+        .ok_or_else(|| anyhow!("could not access standard output"))?;
+    let mut out = String::new();
+    stdout
+        .read_to_string(&mut out)
+        .with_context(|| "could not read standard output")?;
+    Ok(out)
+}
```
Urhengulas, line 93:
```suggestion
fn run_capturing_stdout(cmd: &mut Command) -> Result<String> {
    let o = cmd.output()?.stdout;
    Ok(str::from_utf8(&o)?.to_string())
}
```

Urhengulas: (need to update imports)

Urhengulas (changes requested):
I checked the `test.sh` and as far as I could understand you translated it properly. Can we delete the old file then?

Also I discovered another small nit while looking through the code xD

Urhengulas (approved):

Urhengulas:
🎉

PR #395, opened 2021-02-19
use rust-analyzer.linkedProjects

japaric:
with this VS code configuration RA works transparently across the two Cargo workspaces

### DIFF
--- a/.vscode/settings.json
+++ b/.vscode/settings.json
@@ -1,0 +1,6 @@
+{
+    "rust-analyzer.linkedProjects": [
+        "Cargo.toml",
+        "firmware/Cargo.toml",
+    ]
+}


jonas-schievink:
This looks good to me, but the file conflicts with local configuration (any per-workspace configuration will end up making the file dirty). I'm not sure there's another way to do this though...

Urhengulas:
I see two options to resolve this:
1. Only documenting that developers should add this config, but actually not have it in the repo
2. Force users to have their configuration outside the repo, which seems to be doable as described [on this stackoverflow post](https://stackoverflow.com/a/55533004) (though I didn't fully try it).

japaric:
> Only documenting that developers should add this config, but actually not have it in the repo

that would be useful to have in the `app-template` README (probably easy to miss the new info if you already know how to init an app-template project).

as there's no good solution to incorporate this right now I'll close the PR until we come up with a solution.

PR #396, opened 2021-02-19
defmt_decoder: Bump dep `object` to `0.23.0`

Urhengulas:


jonas-schievink:
This will duplicate the dependency in probe-run since 0.22 is still pulled in by other dependencies

Urhengulas:
> This will duplicate the dependency in probe-run since 0.22 is still pulled in by other dependencies

Oh. Should we then push the dependency to update, or downgrade ourselves?

jonas-schievink:
Let's keep using 0.22.0 for now to get this out the door

PR #398, opened 2021-02-19
defmt 0.2.0

Urhengulas:


oddstr13:
panic-probe release does not seem to have been successfully uploaded to [crates.io](https://crates.io/crates/panic-probe)

Issue #409, opened 2021-02-24
defmt 0.2.0 - docs.rs build failed 


japaric:
https://docs.rs/crate/defmt/0.2.0

jonas-schievink:
```
[INFO] [stderr]  Documenting defmt v0.2.0 (/opt/rustwide/workdir)
[INFO] [stderr] error[E0463]: can't find crate for `std`
[INFO] [stderr]   |
[INFO] [stderr]   = note: the `thumbv6m-none-eabi` target may not be installed
```
weird

jonas-schievink:
Ah, maybe we need to tell it to not pass `--all-features`?

Urhengulas:
This seems to be fixed, but it's not available on docs.rs. How can we update it there?

Urhengulas:
Reading the website it seems opening an issue is the way to request a rebuild (https://docs.rs/about/builds#failures-and-rebuilds). Should we do this?

jonas-schievink:
We can just publish 0.2.1 with this change included

Urhengulas:
Like now, or later?

jonas-schievink:
Feel free to do it now if you want. No other crates have to be published, just `defmt`, so it should be fairly straightforward compared to the 0.2.0 release.

Urhengulas:
Fixed by #413 and published by releasing `defmt v0.2.1` (#428).

PR #413, opened 2021-02-24
Disable `unstable-test` on docs.rs

jonas-schievink:
https://github.com/knurling-rs/defmt/issues/409

### DIFF
--- a/Cargo.toml
+++ b/Cargo.toml
@@ -51,6 +51,6 @@ overflow-checks = false
 semver = "0.11.0"
 
 [package.metadata.docs.rs]
-all-features = true
+features = ["alloc"]
 rustdoc-args = [ "--cfg=docsrs" ]
 targets = ["thumbv6m-none-eabi", "thumbv7em-none-eabihf"]


Urhengulas:
How can one test if this configuration works?

jonas-schievink:
I tested this locally. I'm not aware of any convenient ways to test it in CI.

jonas-schievink:
I ran:

```console
env RUSTDOCFLAGS=--cfg=docsrs cargo +nightly doc --features alloc --target thumbv6m-none-eabi --open
```

Replacing the `--features alloc` with `--all-features` reproduces the error on docs.rs.

japaric (approved):
Looks good to me!
By the way, was what the reason for specifying thumbv6 and thumbv7 as the docs.rs targets? we don't cfg public API based on the target (architecture), right? (`impl Format for Arc` doesn't exist on thumbv6 but `Arc` doesn't exist there either)

jonas-schievink:
> `impl Format for Arc` doesn't exist on thumbv6 but `Arc` doesn't exist there either

It does exist on thumbv7, so it will show up in the docs for thumbv7

PR #415, opened 2021-02-25
Make elf2table::Location public

mattico:
`Table::get_locations` returns `Locations` which is not public. The `Location` type seems pretty safe to make public. There should at least be a public opaque type so the result of `get_locations` can be passed between functions.


Urhengulas:
Sounds good to me! Though be aware that `defmt_decoder` is still unstable and is subject to change.

bors r+

Urhengulas (approved):

PR #421, opened 2021-03-03
defmt-test: update README

Lotterleben:
- update dependency versions
- use `defmt::assert()` and `defmt::assert_eq()` in our example while we're at it


jonas-schievink (approved):
bors r+

jonas-schievink:
bors retry

PR #424, opened 2021-03-05
CI: add nightly run

Lotterleben:
Add nightly run so we can detect CI whoopsies early on

note: maybe we can set up some kind of forwarding of nightly fail e-mails to a zulip stream?


jonas-schievink (approved):

jonas-schievink:
Oh, nice!

Lotterleben:
huh?

jonas-schievink:
I feared that the CI failure would take more time to fix

PR #431, opened 2021-03-15
Simplify `defmt_macros`

Urhengulas:


macros/src/lib.rs
```diff
@@ -1131,7 +1131,7 @@ impl Codegen {
             .iter()
             .filter_map(|frag| match frag {
                 Fragment::Parameter(param) => Some(param.clone()),
-                Fragment::Literal(_) => None,
+                _ => None,
```
jonas-schievink, line 1081: This was exhaustive to cause a compiler error when new variants are added

jonas-schievink (approved):
Looks good, just one minor issue

macros/src/lib.rs
Urhengulas: This makes sense and I didn't think about that. Iirc there was a second place where I did this.

Urhengulas: Seems to be only this one. Should be good now!

PR #432, opened 2021-03-15
Simplify `defmt_parser`

Urhengulas:


parser/src/lib.rs
```diff
@@ -41,6 +40,24 @@ pub enum DisplayHint {
...
+        Ok(match s {
+            "µs" => DisplayHint::Microseconds,
+            "a" => DisplayHint::Ascii,
+            "b" => DisplayHint::Binary,
+            "x" => DisplayHint::Hexadecimal {
+                is_uppercase: false,
+            },
+            "X" => DisplayHint::Hexadecimal { is_uppercase: true },
+            "?" => DisplayHint::Debug,
+            _ => return Err("Error parsing `DisplayHint`".to_string()),
```
jonas-schievink, line 56: Since we don't use this string for anything, might as well make it `()`.

parser/src/lib.rs
```diff
@@ -128,50 +155,55 @@ impl Default for Type {
...
     let end_digits = s
         .as_bytes()
         .iter()
-        .take_while(|b| is_digit(Some(**b as char)))
+        .take_while(|b| (**b as char).is_ascii_digit())
         .count();
     let end = s[..end_digits].parse().ok()?;
 
+    // check for faulty state
+    #[allow(clippy::if_same_then_else)]
```
jonas-schievink, line 182: Clippy is right, this is a bit difficult to understand. Why not use something like:

```rust
if end <= start || start >= 128 {
    return None;
}
```

parser/src/lib.rs
Urhengulas: I didn't think about the option of `()` as an `type Err`. Good idea!

parser/src/lib.rs
Urhengulas: 👍🏾

jonas-schievink:
bors retry

PR #435, opened 2021-03-23
Move `defmt_macros` doc-comments from `defmt` to `defmt_macros`

Urhengulas:
This is because I'd like the documentation live next to the code. `rustdoc` will still use the doc-comments for the re-exports.


Urhengulas:
Closed due to rust doc-tests complications

PR #438, opened 2021-03-26
`defmt_decoder`: Extend doc-comments and make `Frame` public

### LINKED ISSUE #430: Reasoning behind private Frame
What is the reasoning behind making `Frame` private? It seems very restricting to implementations wanting to make custom decoders?

Before it was possible to do something like this
```rust
impl<'a> Stream for FrameStream<'a> {
    type Item = defmt_decoder::Frame<'a>;

    fn poll_next(mut self: Pin<&mut Self>, cx: &mut Context<'_>) -> Poll<Option<Self::Item>> {
        if let Some(logging_channel) = &mut self.logging_channel {
            let mut read_buf = [0; 1024];

            match logging_channel.read(&mut read_buf) {
                Ok(n) if n != 0 => self.frames.extend_from_slice(&read_buf[..n]),
                Ok(_) => {}
                Err(_e) => return Poll::Ready(None),
            };

            match self.table.decode(&self.frames) {
                Ok((frame, consumed)) => {
                    let num_frames = self.frames.len();
                    self.frames.rotate_left(consumed);
                    self.frames.truncate(num_frames - consumed);

                    Poll::Ready(Some(frame))
                }
                Err(_) => {
                    // FIXME: This waker just keeps waking, thus it is extremely
                    // inefficient. Should probably be optimized eventually
                    cx.waker().wake_by_ref();
                    Poll::Pending
                }
            }
        } else {
            Poll::Ready(None)
        }
    }
}
```

But that does not seem possible after making `Frame` private?

Urhengulas:
Fixes #430

### DIFF
--- a/decoder/src/elf2table/mod.rs
+++ b/decoder/src/elf2table/mod.rs
@@ -178,6 +178,7 @@ panic-probe = {{ git = \"https://github.com/knurling-rs/defmt\", features = [\"p
     Ok(())
 }
 
+/// Location of a defmt log statement in the elf-file
 #[derive(Clone)]
 pub struct Location {
     pub file: PathBuf,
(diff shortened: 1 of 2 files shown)


Urhengulas:
@jonas-schievink I actually wasn't sure if some of the doc-comments are exact enough, but glad they were 😅

PR #440, opened 2021-03-27
Update dependencies

Sh3Rm4n:
More specifically update cortex-m from `0.6.x` to `0.7.x` to reduce
the dependency tree a little, see https://github.com/rust-embedded/cortex-m/pull/190

Another interesting update is rtt-target (`0.2.x` -> `0.3.x`)

I've lazily used `cargo upgrade` to update the dependencies, which 
means patch release versions are also changed. I can revert these lines, if not desired :)


Urhengulas:
Looks good to me! Thank you @Sh3Rm4n :)

bors r+

PR #441, opened 2021-03-31
Support more items in `#[tests]` module

### LINKED ISSUE #416: defmt-test: support other items inside the `#[tests]` module
This is currently rejected:

``` rust
#[defmt_test::tests]
mod tests {
    struct State {} //~ error

    #[init]
    fn setup() -> State {
        State {}
    }
}
```

But would be nice if structs, functions, etc. were accepted inside the module.

Right now you can work around the issue by putting helper structs and functions *outside* the `#[tests]` module and then importing them but this is cumbersome and quite different from how the standard `test` harness works.

``` rust
struct State {}

#[defmt_test::tests]
mod tests {
    use super::State;

    #[init]
    fn setup() -> State {
        State {}
    }
}
```

Sh3Rm4n:
This allows every item to be defined inside the `#[tests]` module,
e.g. `struct`s `const`s etc. except for function.

Closes #416


firmware/defmt-test/macros/src/lib.rs
```diff
@@ -61,10 +61,8 @@ fn tests_impl(args: TokenStream, input: TokenStream) -> parse::Result<TokenStrea
                 let attr = match test_kind {
                     Some(it) => it,
                     None => {
-                        return Err(parse::Error::new(
-                            f.span(),
-                            "function requires `#[init]` or `#[test]` attribute",
-                        ));
+                        untouched_tokens.push(Box::new(f));
+                        continue;
```
jonas-schievink, line 65: This might paint us into a corner, since it means we can't add other attributes in the future

firmware/defmt-test/macros/src/lib.rs
```diff
@@ -135,13 +133,29 @@ fn tests_impl(args: TokenStream, input: TokenStream) -> parse::Result<TokenStrea
...
+            Item::Mod(t) => untouched_tokens.push(Box::new(t)),
+            Item::Static(t) => untouched_tokens.push(Box::new(t)),
+            Item::Struct(t) => untouched_tokens.push(Box::new(t)),
+            Item::Trait(t) => untouched_tokens.push(Box::new(t)),
+            Item::TraitAlias(t) => untouched_tokens.push(Box::new(t)),
+            Item::Type(t) => untouched_tokens.push(Box::new(t)),
+            Item::Union(t) => untouched_tokens.push(Box::new(t)),
+            Item::Use(t) => untouched_tokens.push(Box::new(t)),
+            Item::Verbatim(t) => {
+                untouched_tokens.push(Box::new(t));
```
jonas-schievink, line 152: I think you could use a wildcard match here and accumulate `syn::Item`s instead of `Box<dyn ToTokens>`

Urhengulas:
Hi @Sh3Rm4n, Thank you for submitting the PR! Can you please quickly describe how your solution solves the issue :)

firmware/defmt-test/macros/src/lib.rs
Sh3Rm4n: Haha, this is a much simpler solution. :grinning: 
Was underestimating the cleverness of the `quote!` macro.

firmware/defmt-test/macros/src/lib.rs
Sh3Rm4n: I thought about this case as well. 
Keeping the attribute requirement for functions, but allowing everything else is ok?

With the other code change, it would mean, that we can't add any attribute to other items in the future as well, right?

Sh3Rm4n: We can't add a newer attribute in a backwards-compatible way under the assumption that every new attribute(-name) is already used / introduced by another crate. 

Do I understand that correctly? 

What about namespacing? Every attribute from defmt-test lives under `defmt_test::init` or `defmt_test::test` and we generate a `use defmt_test::test` implicitly. Would this be possible or even useful?

jonas-schievink: > We can't add a newer attribute in a backwards-compatible way under the assumption that every new attribute(-name) is already used / introduced by another crate.

Yeah that's correct. It wouldn't be too hard to avoid builtin attributes, but you can use arbitrary procedural macros too, which might define their own stuff.

> Every attribute from defmt-test lives under `defmt_test::init` or `defmt_test::test` and we generate a `use defmt_test::test` implicitly.

There's no name resolution inside the test module, so we can't insert a `use`, but namespacing does sound like a possible solution for attributes we add later (unfortunately requiring the user to write out a rather long path).

japaric:
This looks like a good start to me but I'll let @jonas-schievink  sent this to bors.

I think we could support (auxiliary) functions in a new patch version and still have some degree of extensibility by reserving some attribute names, e.g. `ignore`. I don't see ourselves supporting a wide gamut of attribute within `defmt_test::tests` in any case so if we do fail to reserve an attribute name and require a semver bump to add support a new attribute I don't think that'd be too bad since it won't happen too many times. But nothing of that needs to be addressed in this PR.

jonas-schievink:
It looks like as written this will reject bare `fn`s in the test module, which should give us enough flexibility to add more custom function attributes. This seems good!

bors r+

PR #447, opened 2021-04-13
`README.md`: Document MSRV

### LINKED ISSUE #443: MSRV
What's defmt's MSRV (Minimum Supported Rust Version)?

I've found defmt uses the following features that restrict the MSRV:

`matches!` macro: 1.42+
`u64::MAX` const: 1.43+
function-like proc macros in expression context: 1.45+

However with 1.45 it won't build, I'm not sure why. [code here](https://github.com/smoltcp-rs/smoltcp/pull/455)

```
error[E0425]: cannot find value `dhcp_repr` in this scope
   --> src/macros.rs:10:50
    |
10  |     (trace, $($arg:expr),*) => { defmt::trace!($($arg),*); };
    |                                                  ^^^^ not found in this scope
    | 
   ::: src/dhcp/clientv4.rs:343:17
    |
343 |                 net_trace!("DHCP send renew to {}: {:?}", endpoint, dhcp_repr);
    |                 --------------------------------------------------------------- in this macro invocation
```

1.46 works fine.

I thought I'd ask first, so if 1.45 is unsupported I can skip debugging it...

Urhengulas:
Fix #443

### DIFF
--- a/README.md
+++ b/README.md
@@ -20,6 +20,10 @@ To include `defmt` in your existing project, follow our [Application Setup guide
 
 [Application Setup guide]: https://defmt.ferrous-systems.com/setup-app.html
 
+## MSRV
+`defmt` always compiles on the [latest `stable` rust release](https://github.com/rust-lang/rust/releases/latest). This is enforced by our CI building and testing against this version. <br>
+It still might work on older rust versions, but this isn't ensured.
+
 ## Support
 
 `defmt` is part of the [Knurling] project, [Ferrous Systems]' effort at
--- a/README.md
+++ b/README.md
@@ -21,7 +21,8 @@ To include `defmt` in your existing project, follow our [Application Setup guide
 [Application Setup guide]: https://defmt.ferrous-systems.com/setup-app.html
 
 ## MSRV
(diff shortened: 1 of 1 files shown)


README.md
```diff
@@ -20,6 +20,10 @@ To include `defmt` in your existing project, follow our [Application Setup guide
 
 [Application Setup guide]: https://defmt.ferrous-systems.com/setup-app.html
 
+## MSRV
+`defmt` always compiles on the [latest `stable` rust release](https://github.com/rust-lang/rust/releases/latest). This is enforced by our CI building and testing against this version. <br>
```
japaric, line 24: nit: can we avoid the HTML element here?

japaric (approved):
Looks good!

japaric:
feel free to address nit and send to bors or just send to bors directly
bors d+

README.md
Urhengulas: Yees, but then we either have two spaces, which is bad because it's hard to see, or a newline, but this actually appears differently in the rendered version. But in the end it probably doesn't matter.

PR #453, opened 2021-04-20
Remove minimum timestamp alignment

jonas-schievink:
probe-run currently inserts blank spaces in front of logging messages if no custom timestamp is used. This is pretty useless and looks weird, so this removes that.


Urhengulas (approved):

Urhengulas:
bors retry

PR #464, opened 2021-05-10
`impl<T> Format for {*const, *mut} T where T: Format + ?Sized`

### LINKED ISSUE #442: implement Format for raw pointers
namely for `*const T` and `*mut T`
the same format as std's Debug trait (`{:?}`) should be used: the address in hexadecimal format

Urhengulas:
* This behaves like `std::fmt::Debug` by printing the address as hexadecimal.
* The `*self as *const () as usize` is also used at [rust/library/core/src/fmt/mod.rs#L2130](https://github.com/rust-lang/rust/blob/master/library/core/src/fmt/mod.rs#L2130).
* The `as *const ()`, seems to be required cause else `rustc` complains with "casting `*const T` as `usize` is invalid, cast through a thin pointer first".

Fixes #442

### DIFF
--- a/firmware/qemu/src/bin/log.out
+++ b/firmware/qemu/src/bin/log.out
@@ -113,4 +113,6 @@
 0.000112 INFO [45054, 49406]
 0.000113 INFO [Data { name: b"Hi", value: true }]
 0.000114 INFO true true
-0.000115 INFO QEMU test finished!
+0.000115 INFO 0xaabbccdd
+0.000116 INFO 0xddccbbaa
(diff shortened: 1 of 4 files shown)


Urhengulas:
> Nice! Looks good to me. Could you also add a snapshot test for this? (you can create a raw pointer with `0xaabbccdd as *const u8`). You can send this to bors afterwards.

Will do :) Do you know why the snapshot tests are failing btw?

japaric:
> Do you know why the snapshot tests are failing btw?

#467

PR #465, opened 2021-05-10
Add leading zeros support.

derekdreery:
This adds support for things like `info!("{=u16:08X}", 42")`. It is tested in the mdbook.

I could have added precision support as well, but I figure floats aren't so often used on embedded, whereas printing out a `u32` nicely aligned is a common task.

### DIFF
--- a/decoder/src/frame.rs
+++ b/decoder/src/frame.rs
@@ -137,9 +137,15 @@ fn format_args_real(
         buf: &mut String,
     ) -> Result<(), fmt::Error> {
         match hint {
-            Some(DisplayHint::Binary) => write!(buf, "{:#b}", x)?,
-            Some(DisplayHint::Hexadecimal { uppercase: false }) => write!(buf, "{:#x}", x)?,
-            Some(DisplayHint::Hexadecimal { uppercase: true }) => write!(buf, "{:#X}", x)?,
+            Some(DisplayHint::NoHint { zero_pad }) => write!(buf, "{:#01$}", x, zero_pad)?,
+            Some(DisplayHint::Binary { zero_pad }) => write!(buf, "{:#01$b}", x, zero_pad)?,
+            Some(DisplayHint::Hexadecimal {
+                uppercase,
+                zero_pad,
+            }) => match uppercase {
+                false => write!(buf, "{:#01$x}", x, zero_pad)?,
+                true => write!(buf, "{:#01$X}", x, zero_pad)?,
+            },
             Some(DisplayHint::Microseconds) => {
                 let seconds = x / 1_000_000;
                 let micros = x % 1_000_000;
@@ -156,9 +162,15 @@ fn format_args_real(
         buf: &mut String,
     ) -> Result<(), fmt::Error> {
         match hint {
-            Some(DisplayHint::Binary) => write!(buf, "{:#b}", x)?,
-            Some(DisplayHint::Hexadecimal { uppercase: false }) => write!(buf, "{:#x}", x)?,
-            Some(DisplayHint::Hexadecimal { uppercase: true }) => write!(buf, "{:#X}", x)?,
+            Some(DisplayHint::NoHint { zero_pad }) => write!(buf, "{:#01$}", x, zero_pad)?,
+            Some(DisplayHint::Binary { zero_pad }) => write!(buf, "{:#01$b}", x, zero_pad)?,
+            Some(DisplayHint::Hexadecimal {
+                uppercase,
+                zero_pad,
+            }) => match uppercase {
+                false => write!(buf, "{:#01$x}", x, zero_pad)?,
+                true => write!(buf, "{:#01$X}", x, zero_pad)?,
+            },
             _ => write!(buf, "{}", x)?,
         }
         Ok(())
--- a/parser/src/lib.rs
+++ b/parser/src/lib.rs
@@ -30,10 +30,18 @@ pub struct Parameter {
 /// All display hints
 #[derive(Clone, Debug, PartialEq)]
 pub enum DisplayHint {
+    NoHint {
+        zero_pad: usize,
+    },
     /// `:x` OR `:X`
-    Hexadecimal { uppercase: bool },
+    Hexadecimal {
+        uppercase: bool,
+        zero_pad: usize,
+    },
     /// `:b`
-    Binary,
+    Binary {
+        zero_pad: usize,
+    },
     /// `:a`
     Ascii,
     /// `:?`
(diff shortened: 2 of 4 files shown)


derekdreery:
Ok need to fix up some tests

derekdreery:
Would you mind setting off the CI again sir?

parser/src/lib.rs
```diff
@@ -44,20 +52,39 @@ pub enum DisplayHint {
...
-            "a" => DisplayHint::Ascii,
-            "b" => DisplayHint::Binary,
-            "x" => DisplayHint::Hexadecimal { uppercase: false },
-            "X" => DisplayHint::Hexadecimal { uppercase: true },
-            "?" => DisplayHint::Debug,
-            _ => return Err(()),
-        })
-    }
+fn parse_display_hint(mut s: &str) -> Option<DisplayHint> {
+    let zero_pad = if s.starts_with("0") {
```
Urhengulas, line 56: Maybe use `.strip_prefix`? _(see https://rust-lang.github.io/rust-clippy/master/index.html#manual_strip)_

parser/src/lib.rs
```diff
@@ -30,10 +30,18 @@ pub struct Parameter {
 /// All display hints
 #[derive(Clone, Debug, PartialEq)]
 pub enum DisplayHint {
+    NoHint {
+        zero_pad: Option<ZeroPad>,
```
jonas-schievink, line 34: I think these `zero_pad` fields could just be of type `usize` and fall back to 0 if no zero-padding is requested. That should simplify the `format_args` code.

parser/src/lib.rs
```diff
@@ -118,6 +138,21 @@ impl Level {
...
+///
+/// Returns the integer and remaining text, if `s` started with an integer. Any errors parsing the
+/// number (which we already know only contains digits) are silently ignored.
+fn parse_integer<T: FromStr>(s: &str) -> Option<(&str, usize)> {
+    let start_digits = s
+        .as_bytes()
+        .iter()
+        .copied()
+        .take_while(|b| b.is_ascii_digit())
+        .count();
```
Urhengulas, line 151: Use https://doc.rust-lang.org/std/primitive.str.html#method.chars?

derekdreery: Using a character iterator has the potential to be slightly slower, because each 1-4 bytes will be converted into a 4-byte char. Of course who knows what the optimizer will end up converting your code to, you'd have to benchmark to see. However I don't think there is a downside to using a byte iterator, since we only support looking at ascii characters anyway.

parser/src/lib.rs
derekdreery: Ok converted to using a flat usize, with `0` being equivalent to `None`.

parser/src/lib.rs
```diff
@@ -44,20 +52,39 @@ pub enum DisplayHint {
...
-            "µs" => DisplayHint::Microseconds,
-            "a" => DisplayHint::Ascii,
-            "b" => DisplayHint::Binary,
-            "x" => DisplayHint::Hexadecimal { uppercase: false },
-            "X" => DisplayHint::Hexadecimal { uppercase: true },
-            "?" => DisplayHint::Debug,
-            _ => return Err(()),
-        })
-    }
+fn parse_display_hint(mut s: &str) -> Option<DisplayHint> {
```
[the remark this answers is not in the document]
derekdreery: Because it's an internal implementation detail, I don't see a benefit to using `FromStr`, because a) it has the Err type parameter that we don't need or use, and it means that we special case parsing the display hint compared to other stuff. Happy to hear other people's opinions.

Personally I think the best format for these functions is to follow the `nom` pattern, where you have functions like `fn parse_*(input: &[u8]) -> Option<(&[u8], Type)>`, where the unconsumed input is returned along with the parsed value. This makes it easy to chain parsing functions together. (The DSL here is simple enough that these are not things that need much thought IMO).

derekdreery:
Could anyone help me understand the output from the qemu test?

parser/src/lib.rs
Urhengulas: > Using a character iterator has the potential to be slightly slower, because each 1-4 bytes will be converted into a 4-byte char. Of course who knows what the optimizer will end up converting your code to, you'd have to benchmark to see. However I don't think there is a downside to using a byte iterator, since we only support looking at ascii digits anyway.

Since the behavior is the same in the end I was in favor of `.chars`, since it simplifies the code.  But I wasn't aware of the potential performance overhead, so it sounds good to me.

Urhengulas:
> Could anyone help me understand the output from the qemu test?

What do you mean? `qemu-snapshot`?

decoder/src/frame.rs
```diff
@@ -156,9 +163,16 @@ fn format_args_real(
...
+            Some(DisplayHint::NoHint { zero_pad }) => write!(buf, "{:#01$}", x, zero_pad)?,
+            Some(DisplayHint::Binary { zero_pad }) => write!(buf, "{:#01$b}", x, zero_pad)?,
+            Some(DisplayHint::Hexadecimal {
+                uppercase: false,
+                zero_pad,
+            }) => write!(buf, "{:#01$x}", x, zero_pad)?,
+            Some(DisplayHint::Hexadecimal {
+                uppercase: true,
+                zero_pad,
+            }) => write!(buf, "{:#01$X}", x, zero_pad)?,
```
Urhengulas, line 175: I am slightly in favor of this pattern, cause I think it is more clear:
```suggestion
            Some(DisplayHint::Hexadecimal {
                uppercase,
                zero_pad,
            }) => match uppercase {
                false => write!(buf, "{:#01$X}", x, zero_pad)?,
                true => write!(buf, "{:#01$x}", x, zero_pad)?,
            },
```
_(same for `fn format_u128`)_

Urhengulas (changes requested):
Overall I am good now, only two topics:
1. a small nit about formatting some match arms 😁  (see suggestions)
2. please remove the changes to `xtask` from this pr and send it as separate one :)

derekdreery:
Made requested changes.

Urhengulas:
Awesome! Please rebase and then we can merge (if you "allow edit from maintainers" i can also do this for you, see https://docs.github.com/en/github/collaborating-with-issues-and-pull-requests/allowing-changes-to-a-pull-request-branch-created-from-a-fork)

derekdreery:
Squashed and rebased.

Urhengulas:
Thank you for the PR @derekdreery! Feel free to send your PR for `xtask` :D

Issue #452, opened 2021-04-20
Snapshot tests are failing without an error message


jonas-schievink:
Running `cargo xtask test-snapshot` locally results in this output:

```diff
🧪 qemu/snapshot
log (dev)
-0.000000 INFO Hello!
-0.000001 INFO World!
-0.000002 INFO The answer is 42
-0.000003 INFO Hello 42 42!
-0.000004 INFO Hello 256 42 false
-0.000005 INFO 🍕 slice [3, 14]
-0.000006 INFO 🍕 array [3, 14, 1]
-0.000007 INFO float like a butterfly 5.67 5.67
...
```

and so on, for every snapshot test. xtask should display the Cargo/rustc output here instead of just outputting this unhelpful diff.

Urhengulas:
@spookyvision Do you want to take this on?

spookyvision:
no time, sorry

Urhengulas:
@jonas-schievink Do you prefer only the rustc output, or both rustc and the diff?

jonas-schievink:
Just rustc output is enough, the diff is always going to be "everything deleted" if compilation fails

spookyvision:
(I'd keep the diff around for other cases though?)

Urhengulas:
> (I'd keep the diff around for other cases though?)

In which cases do you imagine it to be useful?

spookyvision:
the diffs are detailed info about "expected" vs "actual" output, so you'd use that to narrow a problem down

Urhengulas:
I will go forward with removing the diff in the first step and replace it with the rustc output. Later we can add the diff back, but probably guarded by a cli-flag.

Urhengulas:
> The diff should be displayed if the compilation succeeds, it just isn't of much use if it fails

akay

PR #469, opened 2021-05-12
Xtask refactor

Urhengulas:
When starting to look into `#452` I got a bit hung up on refactoring 😬 

_Review note: The functions relating to the target also got refactored, not only moved out._

### DIFF
--- a/xtask/src/main.rs
+++ b/xtask/src/main.rs
@@ -98,9 +98,10 @@ fn run_command(cmd_and_args: &[&str], cwd: Option<&str>, env: &[(&str, &str)]) -
         })
 }
 
+/// Execute the [`Command`] and return the text emitted to `stdout`, if valid UTF-8
 fn run_capturing_stdout(cmd: &mut Command) -> anyhow::Result<String> {
-    let o = cmd.output()?.stdout;
-    Ok(str::from_utf8(&o)?.to_string())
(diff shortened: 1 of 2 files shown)


xtask/src/main.rs
```diff
@@ -1,37 +1,28 @@
...
-        long,
-        short,
-        help = "treat compiler warnings as errors (RUSTFLAGS=\"--deny warnings\")"
-    )]
-    deny_warnings: bool,
-
     #[structopt(subcommand)]
     cmd: TestCommand,
+    #[structopt(long, short)]
+    /// Treat compiler warnings as errors (`RUSTFLAGS="--deny warnings"`)
```
jonas-schievink, line 18: I prefer to place doc comments before any attributes
```suggestion
    /// Treat compiler warnings as errors (`RUSTFLAGS="--deny warnings"`)
    #[structopt(long, short)]
```

xtask/src/main.rs
```diff
@@ -1,37 +1,28 @@
...
+    deny_warnings: bool,
+    #[structopt(long, short)]
+    /// Keep target toolchains that were installed as dependency
```
jonas-schievink, line 21:
```suggestion
    /// Keep target toolchains that were installed as dependency
    #[structopt(long, short)]
```

spookyvision (approved):

PR #471, opened 2021-05-14
`xtask`: Simplify `fn utils::run_command`

Urhengulas:
The changes in `main.rs` are the adaptions to the function signature change.

### DIFF
--- a/xtask/src/utils.rs
+++ b/xtask/src/utils.rs
@@ -24,34 +24,25 @@ pub fn run_capturing_stdout(cmd: &mut Command) -> anyhow::Result<String> {
     Ok(str::from_utf8(&stdout)?.to_string())
 }
 
-pub fn run_command(cmd_and_args: &[&str], cwd: Option<&str>, env: &[(&str, &str)]) -> anyhow::Result<()> {
-    let cmd_and_args = Vec::from(cmd_and_args);
-    let mut cmd = &mut Command::new(cmd_and_args[0]);
-    if cmd_and_args.len() > 1 {
-        cmd.args(&cmd_and_args[1..]);
-    }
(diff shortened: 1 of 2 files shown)


xtask/src/utils.rs
```diff
@@ -24,34 +24,30 @@ pub fn run_capturing_stdout(cmd: &mut Command) -> anyhow::Result<String> {
...
-    }
-
-    let cwd_s = if let Some(path_ref) = cwd {
-        cmd = cmd.current_dir(path_ref);
-        format!("CWD:{} ", path_ref)
+pub fn run_command(
+    program: &str,
+    args: &[&str],
+    cwd: Option<&str>,
+    envs: Vec<(&str, &str)>,
```
jonas-schievink, line 31: You shouldn't need to pass a `Vec` here, a slice should work fine (perhaps with a `.copied()` somewhere in this function), that should be a bit simpler

Urhengulas: I tried it with the `envs: &[(&str, &str)]` first, but this didn't work and the only solution I figured out is `Vec::from(envs)`. Doing the vector creation in the functions makes the api easier, but is a lie 😬

Urhengulas: With just cmd.envs(&[(&str, &str]);` I get:
```
error[E0271]: type mismatch resolving `<&[(&str, &str)] as IntoIterator>::Item == (_, _)`
  --> xtask/src/utils.rs:34:9
   |
34 |     cmd.envs(env);
   |         ^^^^ expected reference, found tuple
   |
   = note: expected reference `&(&str, &str)`
                  found tuple `(_, _)`
```

jonas-schievink: What happens if you pass `env.iter().copied()`?

Urhengulas: Noooooo, it works 😭 ... I mean yay! 🙂

Issue #411, opened 2021-02-24
globally disabled logs still require symbols on x86 = linker error


japaric:
This program:
``` rust
fn main() {
    defmt::info!("hello");
    println!("hello");
}
```
fails to link if you use `defmt` v0.2.0
``` console
$ cargo r
(..)
  = note: /usr/bin/ld: (..)/target/debug/deps/libdefmt-00ba9b2b47686b1f.rlib(defmt-00ba9b2b47686b1f.defmt.bhjss6u8-cgu.12.rcgu.o): in function `defmt::export::acquire':
          (..)/defmt-0.2.0/src/export.rs:36: undefined reference to `_defmt_acquire'
          /usr/bin/ld: (..)/target/debug/deps/libdefmt-00ba9b2b47686b1f.rlib(defmt-00ba9b2b47686b1f.defmt.bhjss6u8-cgu.12.rcgu.o): in function `defmt::export::release':
          (..)/defmt-0.2.0/src/export.rs:48: undefined reference to `_defmt_release'
          /usr/bin/ld: (..)/target/debug/deps/libdefmt-00ba9b2b47686b1f.rlib(defmt-00ba9b2b47686b1f.defmt.bhjss6u8-cgu.12.rcgu.o): in function `defmt::export::timestamp':
          (..)/defmt-0.2.0/src/export.rs:60: undefined reference to `_defmt_timestamp'
          collect2: error: ld returned 1 exit status
```

but links and runs fine if you use `defmt` v0.1.3
``` console
$ cargo r
(..)
     Running `target/debug/hello`
hello
```

note that all logs are disabled in both cases (no Cargo features were added to the crate's Cargo.toml)

japaric:
git bisect points to PR #291 as the cause of this issue

japaric:
a user has reported the `_defmt_panic` symbol is also required when host = x86 windows

WasabiFan:
Unfortunately (in addition to having inadvertently introduced breaking changes), this doesn't seem to solve the problem in all cases. To give a sampling:
- My project's normal release build works fine, but building the tests on my Windows box returns the original linker errors
  - Depending on defmt via a local path dependency avoids any error, but depending on it via git or crates.io fails to compile as described above
  - Deleting a few arbitrary modules in my project fixes the error (probably something to do with Microsoft's linker giving up on optimizations if the project is too big? Not clear.)
- Other team members on their own Windows machines cannot reproduce this error (it builds and tests run cleanly for them)
[...]

PR #477, opened 2021-05-17
Disable logging calls via conditional compilation when all defmt features are disabled

WasabiFan:
This is my take on fixing https://github.com/knurling-rs/defmt/issues/411. It's not clear whether this is the "right" fix, as that issue was marked with a "medium" difficulty tag which would seem to imply there was more work expected here. Either way, it allows the sample in the linked issue to build for me, and fixes the usage in my own project.

The change here is to replace an "if" emitted by the logging macros (albeit with constant boolean conditions evaluation time) with an explicit `#[cfg(...)]` annotation. With the new form, the compiler seems to successfully strip out the unreachable macro body and as such there are no broken references to the nonexistent symbols.

### DIFF
--- a/macros/src/lib.rs
+++ b/macros/src/lib.rs
@@ -475,13 +475,17 @@ fn as_native_type(ty: &Type) -> Option<String> {
     }
 }
 
-fn is_logging_enabled(level: Level) -> TokenStream2 {
+fn cfg_if_logging_enabled(level: Level) -> TokenStream2 {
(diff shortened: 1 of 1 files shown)


WasabiFan:
Oh, yeah, and this has only gone through cursory testing and only on Windows. If there's a sensible place for me to introduce a test for this behavior, I'd be happy to do that; I don't see any tests for these macros as of now.

jonas-schievink:
Thanks, this looks good!

bors r+

PR #478, opened 2021-05-18
add dbg! macro

### LINKED ISSUE #394: defmt::dbg! macro
it would be nice to have a defmt version of the [`std::dbg!`](https://doc.rust-lang.org/std/macro.dbg.html) macro; possibly logging the value at the *trace* logging level.

japaric:
behaves like the one in the `std` library but logs the data at TRACE level

closes #394 

TODO:
- [x] support more than one expression

### DIFF
--- a/firmware/qemu/src/bin/dbg.rs
+++ b/firmware/qemu/src/bin/dbg.rs
@@ -16,6 +16,12 @@ fn main() -> ! {
     // dbg! in log statement
     defmt::info!("the answer is {}", dbg!(x - 1));
 
+    // dbg! with multiple arguments
+    let _: (i32, i32) = dbg!(x - 2, x + 2);
+
+    // dbg! with zero arguments
+    let _: () = dbg!();
+
     loop {
         debug::exit(debug::EXIT_SUCCESS)
     }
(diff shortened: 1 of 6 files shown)


japaric:
now with support for multiple expressions :octopus:

firmware/qemu/src/bin/dbg.rs
```diff
@@ -16,6 +16,12 @@ fn main() -> ! {
     // dbg! in log statement
     defmt::info!("the answer is {}", dbg!(x - 1));
 
+    // dbg! with multiple arguments
+    let _: (i32, i32) = dbg!(x - 2, x + 2);
+
+    // dbg! with zero arguments
+    let _: () = dbg!();
```
jonas-schievink, line 23: Might also make sense to test `dbg!(1,)`

src/lib.rs
```diff
@@ -219,6 +219,11 @@ pub use defmt_macros::trace;
 /// [the manual]: https://defmt.ferrous-systems.com/macros.html
 pub use defmt_macros::warn;
 
+/// Just like the [`std::dbg!`] macro but `defmt` is used to log the message at TRACE level
```
jonas-schievink, line 222:
```suggestion
/// Just like the [`std::dbg!`] macro but `defmt` is used to log the message at `TRACE` level.
```

jonas-schievink (approved):

PR #488, opened 2021-05-23
Structure `impl Format`s into multiple files

Urhengulas:


src/impls/mod.rs
```diff
@@ -0,0 +1,10 @@
+#[cfg(feature = "alloc")]
+mod allocs;
```
jonas-schievink, line 2: Since this refers to the `alloc` crate, it should probably be called `alloc`
```suggestion
mod alloc;
```

src/impls/mod.rs
```diff
@@ -0,0 +1,10 @@
...
+mod arrays;
+mod cores;
```
jonas-schievink, line 4: (same here)

```suggestion
mod core;
```

src/impls/mod.rs
Urhengulas: this would name-clash with the `extern crate alloc`, if the `alloc` feature is activated: https://github.com/knurling-rs/defmt/blob/fa902a95c5748fbd55f3ad173fe3d9d5873c68a8/src/lib.rs#L18-L19

jonas-schievink: Hmm, but this is in `impls`, not in the root, so shouldn't it be fine?

Urhengulas: running `cargo xtask test-host` will error with multiple `error[E0433]: failed to resolve: could not find 'borrow' in 'alloc'`

src/impls/mod.rs
Urhengulas: also this would name-clash with the reserved `core`.

`cargo check` errors with `error[E0412]: cannot find type 'Iter' in module 'core::slice'`

src/impls/mod.rs
jonas-schievink: Huh, okay, then let's go with `alloc_` maybe?

Urhengulas: I tried to match the plural I used for `primitives`, `tuples`, `arrays`, but I am also good with this naming 😄

Urhengulas:
@jonas-schievink any blockers for merging this?

PR #489, opened 2021-05-23
Structure lib

Urhengulas:
The second commit should not work speaking from one comment in the code, but it seems to do anyway 🤔


jonas-schievink:
Ah, the `core::unreachable!()` was changed at some point to always use the libcore `panic!()` macro. Since we track stable Rust, relying on that should be fine.

jonas-schievink (approved):
Looks good!

It might make sense to `#![deny(unreachable_pub)]` to make sure that everything marked `pub` is reexported from the root module

bors r+

PR #493, opened 2021-05-28
Support test of single snapshot

justahero:
This change adds a CLI argument to the `test-snapshot` subcommand to run a single snapshot test. Without the parameter all snapshot tests are run. With the change it's easier to debug a single snapshot using `xtask`.

To run the `log` snapshot test: `cargo xtask test-snapshot -n log`.

Omitting a name runs all snapshot tests.

### DIFF
--- a/xtask/src/main.rs
+++ b/xtask/src/main.rs
@@ -33,7 +33,10 @@ enum TestCommand {
     TestCross,
     TestHost,
     TestLint,
-    TestSnapshot,
+    TestSnapshot {
+        #[structopt(short, help = "Name of the snapshot")]
+        name: Option<String>,
+    },
 }
 
 fn main() -> anyhow::Result<()> {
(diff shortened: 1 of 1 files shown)


xtask/src/main.rs
```diff
@@ -52,7 +55,8 @@ fn main() -> anyhow::Result<()> {
         }
         TestCommand::TestHost => test_host(opt.deny_warnings),
         TestCommand::TestCross => test_cross(),
-        TestCommand::TestSnapshot => test_snapshot(),
+        TestCommand::TestSnapshot { name: None } => test_snapshot(),
+        TestCommand::TestSnapshot { name: Some(name) } => test_single_snapshot(&name, "", false)?,
```
Urhengulas, line 59: Do we want to use `do_test(|| test_single_snapshot(test, features, false), "qemu/snapshot");` ? And do we want to only run in debug mode, or debug and release?

Urhengulas, line 59:
```suggestion
        TestCommand::TestSnapshot { name: Some(name) } => {
            do_test(|| test_single_snapshot(&name, "", false), "qemu/snapshot")
        }
```

Urhengulas (changes requested):
I really like this feature and often though that it would be helpful to have it 😉

xtask/src/main.rs
justahero: Good question, I wanted to run a single snapshot program to minimize test time.

justahero:
After testing this change a bit more today, I must say it may not work for what I had in mind. At first I wanted to be able to debug a single snapshot program via Visual Studio Code, but I could not make it work unfortunately.

Urhengulas:
> I wanted to be able to debug a single snapshot program via Visual Studio Code

What do you mean. Running it with the "run test" button?

Urhengulas:
Let's close this PR for now, but feel free to reopen, if desired.

PR #495, opened 2021-05-30
`macros`: match unused vars if logging is disabled

### LINKED ISSUE #494: semver-breaking: "Unused variable" errors with deny(unused).
The following used to build fine with defmt 0.2.1, and now fails with defmt 0.2.2. This is due to #477 

```rust
#![deny(unused)]
let x = 1; // error: unused variable: `x` if no "defmt-xxx" features are enabled
info!("x={}", x);
```

Urhengulas:
Add `cfg_if 1.0` to `defmt`.

Fixes #494

### DIFF
--- a/macros/src/lib.rs
+++ b/macros/src/lib.rs
@@ -513,15 +513,21 @@ fn log(level: Level, log: FormatArgs) -> TokenStream2 {
     let sym = mksym(&ls, level.as_str(), true);
     let logging_enabled = cfg_if_logging_enabled(level);
     quote!({
-        #[#logging_enabled] {
-            match (#(&(#args)),*) {
-                (#(#pats),*) => {
-                    if let Some(mut _fmt_) = defmt::export::acquire() {
-                        _fmt_.header(&defmt::export::istr(#sym));
-                        #(#exprs;)*
-                        _fmt_.finalize();
-                        defmt::export::release(_fmt_)
+        cfg_if::cfg_if! {
+            if #[#logging_enabled] {
+                match (#(&(#args)),*) {
+                    (#(#pats),*) => {
+                        if let Some(mut _fmt_) = defmt::export::acquire() {
+                            _fmt_.header(&defmt::export::istr(#sym));
+                            #(#exprs;)*
+                            _fmt_.finalize();
+                            defmt::export::release(_fmt_)
+                        }
                     }
+                }
+            } else {
+                match (#(&(#args)),*) {
+                    _ => {}
                 }
             }
         }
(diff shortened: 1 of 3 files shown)


macros/src/lib.rs
```diff
@@ -513,16 +513,22 @@ fn log(level: Level, log: FormatArgs) -> TokenStream2 {
...
     quote!({
-        #[#logging_enabled] {
-            match (#(&(#args)),*) {
-                (#(#pats),*) => {
-                    if let Some(mut _fmt_) = defmt::export::acquire() {
-                        _fmt_.header(&defmt::export::istr(#sym));
-                        #(#exprs;)*
-                        _fmt_.finalize();
-                        defmt::export::release(_fmt_)
+        cfg_if::cfg_if! {
```
Dirbaio, line 516: This breaks if the end-user's crate doesn't have cfg_if as a dependency

Urhengulas: Should't it be okay, since I've added it to `defmt`s dependencies?

Urhengulas: Or do you mean if someone uses `defmt-macros` directly?

Urhengulas: I see 🤔

Urhengulas:
Re-exporting solves the problem @Dirbaio brought up, but I am very unsure if we really want to do this. One the one hand it works and also `cfg_if` seems quite stable. On the other hand it feels like a weird solution, but honestly all macro-related things feel a bit weird to me.
Opinions?

On another note: Does anybody know what the dependencies `core` and `compiler_builtins` are used for? Can we just disable them?

jonas-schievink:
Wouldn't it be easier to just avoid `cfg_if` and use the negated `#[cfg]` expression?

Urhengulas:
> Wouldn't it be easier to just avoid `cfg_if` and use the negated `#[cfg]` expression?

Yes, but then we need to maintain two config-expressions and change the other one, in case the first one does. `cfg_if` seemed like the cleaner and more readable one to me.

On the other hand we probably won't change it too often, so it is okay if we add a comment warning about this relationship.

In the end I am fine with both. What do you prefer?

jonas-schievink:
The condition is generated by the macro (`#[#logging_enabled]`), inverting that should just be `#[not(#logging_enabled)]`

Urhengulas:
> The condition is generated by the macro (`#[#logging_enabled]`), inverting that should just be `#[not(#logging_enabled)]`

Oh, that's nice. I though that the `not` (and others) only work inside the `cfg(...)`, but good to know it is more flexible.

jonas-schievink:
Err, no, you're right, it has to be `#[cfg(not(...))]`, but a slight refactoring should let you do that

Urhengulas:
> Err, no, you're right, it has to be `#[cfg(not(...))]`, but a slight refactoring should let you do that

Also just noticed xD `error[E0658]: attributes on expressions are experimental`

PR #497, opened 2021-05-30
`macros`: match unused vars if logging is disabled

### LINKED ISSUE #494: semver-breaking: "Unused variable" errors with deny(unused).

Urhengulas:
Fixes #494

### DIFF
--- a/macros/src/lib.rs
+++ b/macros/src/lib.rs
@@ -480,11 +480,9 @@ fn cfg_if_logging_enabled(level: Level) -> TokenStream2 {
     let features_release = necessary_features_for_level(level, false);
 
     quote!(
-        cfg(
-            any(
-                all(debug_assertions, any(#( feature = #features_dev ),*)),
-                all(not(debug_assertions), any(#( feature = #features_release ),*))
-            )
+        any(
+            all(    debug_assertions,  any(#( feature = #features_dev     ),*)),
+            all(not(debug_assertions), any(#( feature = #features_release ),*))
         )
     )
 }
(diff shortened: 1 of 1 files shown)


Urhengulas:
I am a bit unsure how to test this. Running the snapshot tests with `--no-default-features` would probably work, since it then would emit the unused-vars-warnings which should mismatch the snapshots, but this seems a bit over the top.

macros/src/lib.rs
```diff
@@ -512,6 +526,7 @@ fn log(level: Level, log: FormatArgs) -> TokenStream2 {
 
     let sym = mksym(&ls, level.as_str(), true);
     let logging_enabled = cfg_if_logging_enabled(level);
+    let logging_disabled = cfg_if_logging_disabled(level);
     quote!({
         #[#logging_enabled] {
```
jonas-schievink, line 531: If you change this to

```suggestion
        #[cfg(#logging_enabled)] {
```

and change the code above to not include `cfg()`, you don't need 2 separate functions and `quote!` invocations, that should be a bit easier to read

macros/src/lib.rs
```diff
@@ -525,6 +540,11 @@ fn log(level: Level, log: FormatArgs) -> TokenStream2 {
                 }
             }
         }
+        // if logging is disabled match args, so they are not unused
+        #[#logging_disabled]
```
jonas-schievink, line 544:
```suggestion
        #[cfg(not(#logging_enabled))]
```

macros/src/lib.rs
Urhengulas: That's smart 🥲

PR #505, opened 2021-06-11
[1/n] - Logger trait v2.

Dirbaio:
Part 1 of N of #492 

- Panics on reentrant acquire.
- Remove `Write` trait.

Issue #492, opened 2021-05-27
RFC: Traits and wire format 2.0


Dirbaio:
I'm opening this issue to collect discussions that are scattered across many issues in a single coherent proposal.

## Goals

The goal is to improve code size (#456), while fixing #455 #457 #458 as a bonus. The other main metrics (wire data size, and speed) shouldn't be sacrificed.

Better code size requires **simpler code**. Less things to do means smaller code.

In particular, making formatting require **no state** leads to the biggest code size wins from my experiments in #456. Currently `defmt` stores some state in `InternalFormatter`. This state has to be stack-allocated, initialized and passed around at every log callsite, which requires a nontrivial amount of instructions.

### 1. Use [`rzcobs`](https://github.com/Dirbaio/rzcobs) to encode the stream.

The wire stream contains many zero bytes, rzCOBS can compress them very fast. This offsets wire size increase from removing LEB128 encoding (proposed below).

- **Advantage**: Improved wire size. In a [test](https://github.com/knurling-rs/defmt/issues/456#issuecomment-827175455) it reduces wire size by 38%
- **Advantage**: More resilience to corrupted stream. Since rzCOBS has builtin framing, if a frame corrupts the decoder can recover and keep decoding the next frames.
- **Disadvantage**: Slower. In the same test it's 10% slower. Maybe it should be made optional so the user can choose.

### 2. Remove LEB128 encoding

- **Advantage**: Faster.
- **Advantage**: usize/isize will now be fixed width, which allows for interesting optimizations later.
- **Disadvantage**: Almost no code size gain, since LEB128 encoding is not inlined.
- **Disadvantage**: Bigger wire size, but it's offset by rzCOBS.

### 3. Remove bool compression

Unfortunately it's simply not possible to do without state. Bools will be encoded as a single
byte. false = 0x00, true = 0x01. At least 0x00s will be compressed by rzCOBS.

- **Advantage**: Faster.
- **Advantage**: Smaller code (because no state).
- **Disadvantage**: Bigger wire size, but not that much (printing many bools is not that common, somewhat offset by rzCOBS).

### 4. Apply "needs_tag" optimization to "first level" only.

Naively, array/slice contents would be encoded as `tag+data+tag+data+..`. This wastes
space for the redundant tags, since all array items are the same type, which should *theoretically* have the same tag.

[...]

PR #507, opened 2021-06-13
Remove code-size-costly optimizations

Dirbaio:
Part 2 of N of #492. Depends on #505 

WIP, opening to testrun CI


decoder/src/decoder.rs
```diff
@@ -248,15 +224,9 @@ impl<'t, 'b> Decoder<'t, 'b> {
...
-                    let arc = Arc::new(Bool::FALSE);
-                    args.push(Arg::Bool(arc.clone()));
-                    self.bools_tbd.push(arc.clone());
-                    if self.bools_tbd.len() == MAX_NUM_BOOL_FLAGS {
-                        // reached end of compression block: sprinkle values into args
-                        self.read_and_unpack_bools()?;
-                    }
-                }
+                Type::Bool => args.push(Arg::Bool(Arc::new(Bool(AtomicBool::new(
+                    self.bytes.read_u8()? != 0,
```
jonas-schievink, line 225: I would prefer to make it an error if this is neither 0 nor 1, that gives us more ways to detect a malformed stream, and leaves some unassigned values for later optimizations

jonas-schievink (changes requested):
Looks good!

decoder/src/decoder.rs
Dirbaio: Makes a lot of sense, done! I also removed the Arc.

jonas-schievink (approved):

Urhengulas:
The same applies for this PR, as for #505 :)

PR #508, opened 2021-06-14
[5/n] Format trait v2

Dirbaio:
Part of #492 . Depends on ~~#505 #507 #521~~ #524

- Implemented new Format trait according to #492.
- Adjusted everything else accordingly.


Dirbaio:
This is now ready to go! @japaric @jonas-schievink

src/formatter.rs
```diff
@@ -230,34 +14,3 @@ pub struct Str {
     /// 14-bit address
```
japaric, line 14: nit: this is no longer true because this can be up to 16 bits in size

japaric (approved):
This is awesome! It's great to see the `InternalFormatter` go away. Also thanks for the macros to simplify the primitive implementations.

Something I noticed that doesn't need to be addressed in this PR:

I no longer see a point to the `internp!` vs `intern!` distinction because we no longer LEB encode the tag.

Originally, `internp!` was created to put "primitive" tags like `=u8` at the beginning of the `.defmt` section so they get smaller values (`<128`). Because we LEB encoded those tags, primitive tags would be encoded as 1 byte on the wire. The rationale was that primitives are more likely to appear behind `=?` so they'll take less space on the wire but now we always encode the tag as 2 bytes.

So I think `internp!` and  the `.defmt.prim` input section can probably go away.

Urhengulas:
> This is awesome! It's great to see the `InternalFormatter` go away. Also thanks for the macros to simplify the primitive implementations.
> 
> Something I noticed that doesn't need to be addressed in this PR:
> 
> I no longer see a point to the `internp!` vs `intern!` distinction because we no longer LEB encode the tag.
> 
> Originally, `internp!` was created to put "primitive" tags like `=u8` at the beginning of the `.defmt` section so they get smaller values (`<128`). Because we LEB encoded those tags, primitive tags would be encoded as 1 byte on the wire. The rationale was that primitives are more likely to appear behind `=?` so they'll take less space on the wire but now we always encode the tag as 2 bytes.
> 
> So I think `internp!` and the `.defmt.prim` input section can probably go away.

Should we track this in an issue? Or note it in the RFC?

Dirbaio:
There is still *some* advantage of putting primitives first. Tags less than 256 have one zero byte. When rzCOBS-encoded, these will be 9 bits (8+1) instead of 16 bits.

japaric:
> When rzCOBS-encoded, these will be 9 bits (8+1) instead of 16 bits.

(I have not been following all PRs but) this is not being used in the code base at the moment, right? at least, that didn't seem to be the case from looking at the decoder tests. Would rzCOBS be applied *not* in the log macros but rather in the transport layer crate, e.g. defmt-rtt?

Dirbaio:
Yeah, it's not done yet, but coming next. I have a PoC [here](https://github.com/Dirbaio/defmt-codesize-bench/blob/main/src/rtt_4_rzcobs.rs).

Yes, it's in the Logger impl. It has to be because impls that allow concurrent logging (eg one channel per irq prio level) need to have one rzcobs encoder state for each channel. If it was on the `defmt` side I don't think there's a way to do that...

PR #509, opened 2021-06-16
`impl Format for NonZero*`

### LINKED ISSUE #501: impl Format for `NonZero*` types
[`NonZeroU8`](https://doc.rust-lang.org/std/num/struct.NonZeroU8.html), [NonZeroUsize](https://doc.rust-lang.org/std/num/struct.NonZeroUsize.html) etc etc currently can't be formatted with `defmt`, which is pretty painul if you've e.g. got an Enum where one of the fields has a `NonZeroXX` type associated with it – then you've got to manually impl `Format` for the whole thing.

This is issue is a bookmark to improve that when we have time :)

Urhengulas:
Fixes #501


src/impls/core_/num.rs
```diff
@@ -0,0 +1,26 @@
+use core::num;
+
+use super::*;
+
+macro_rules! non_zero {
+    ($type:ty) => {
+        impl Format for $type {
+            fn format(&self, fmt: Formatter) {
+                crate::write!(fmt, "{}", self.get());
```
jonas-schievink, line 9: This format string could specify the type directly, which would be more efficient. Not a blocker, but please create an issue tracking this (or add a FIXME) if you don't want to do it in this PR.

jonas-schievink (approved):

src/impls/core_/num.rs
Urhengulas: I see. I will try to do it here.

Urhengulas: What do you think?

jonas-schievink: perfect!

Urhengulas: 😊

PR #504, opened 2021-06-07
Xtask qemu overwrite

derekdreery:
This adds an `--overwrite` option to the snapshot test, that overwrites the original output.

I've broken it out of #503 and will rebase once that is merged (in some form).

PR #512, opened 2021-06-18
Add overwrite option for xtask cross results.

derekdreery:
#504 rebased and squashed into single commit.

### DIFF
--- a/xtask/src/main.rs
+++ b/xtask/src/main.rs
@@ -9,7 +9,9 @@ use structopt::StructOpt;
 mod targets;
 mod utils;
 
-use crate::utils::{load_expected_output, run_capturing_stdout, run_command, rustc_is_nightly};
+use crate::utils::{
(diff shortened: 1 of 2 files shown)


xtask/src/main.rs
```diff
@@ -33,7 +35,11 @@ enum TestCommand {
     TestCross,
     TestHost,
     TestLint,
-    TestSnapshot,
+    TestSnapshot {
```
Urhengulas, line 38: Let's add a small text, so people know about this option, when running `cargo xtask --help`. I don't have strong feelings about the wording, but this is my suggestion:
```suggestion
    /// Run snapshot tests or optionally update the expected output
    TestSnapshot {
```

Urhengulas (changes requested):

derekdreery:
I'll have a look and see if I can implement this now :)

PR #513, opened 2021-06-19
book/duplicates.md: discriminator -> disambiguator

eupn:
While reading the latest version of `defmt` book, I decided to try and list ELF symbols and realized that in the book, the word `discriminator` is used in the ["Dealing with duplicates"](https://defmt-next.ferrous-systems.com/duplicates.html#dealing-with-duplicates) chapter, but in reality, the code and JSON in the symbol table uses a differently named field (`disambiguator`) with the same meaning.

This PR updates the book to keep these terms in sync.

### DIFF
--- a/book/src/duplicates.md
+++ b/book/src/duplicates.md
@@ -42,14 +42,14 @@ static SYM: u8 = 0;
 
 which results in a collision.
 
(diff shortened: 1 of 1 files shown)


Urhengulas:
Hello @eupn, what is your motivation for this change?

eupn:
Hi @Urhengulas: there's a discrepancy between the word (discriminator vs disambiguator) used in the code and in the book and I'm trying to rectify it here.

Urhengulas:
Thank you @eupn!

PR #518, opened 2021-06-23
`build.rs`: Obtain version from macro; simplify

### LINKED ISSUE #517: Obtain git hash from `git_version` crate
Use the `git_version` crate for obtaining the git hash, instead of the custom build-script we used so far. Similar to https://github.com/knurling-rs/probe-run/pull/221

Urhengulas:
Fixes #517.

### DIFF
--- a/build.rs
+++ b/build.rs
@@ -1,56 +1,50 @@
-use semver::Version;
-use std::{env, error::Error, fs, path::Path, path::PathBuf, process::Command};
+use std::{env, error::Error, fs, path::PathBuf};
+
+use git_version::git_version;
 
 fn main() -> Result<(), Box<dyn Error>> {
+    // `--match v*` matches defmt tags, e.g. `v0.2.3`, but ignores related crates, e.g. `defmt-decoder-v0.2.2`
+    const GIT_DESCRIBE: &str = git_version!(args = ["--long", "--match", "v*"], cargo_prefix = "");
+
+    let version = match semver::Version::parse(GIT_DESCRIBE) {
+        // if parsing fails it is a git version
+        Err(_) => extract_git_hash(GIT_DESCRIBE).to_string(),
+        // if success it is semver
+        Ok(semver) => match semver.major {
+            // minor is breaking when major = 0
+            0 => format!("{}.{}", semver.major, semver.minor),
+            // ignore minor, patch, pre and build
+            _ => semver.major.to_string(),
+        },
+    };
+
+    // Load linker-script and insert version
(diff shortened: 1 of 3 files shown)


Urhengulas:
I am a bit unsure if setting `QEMU_RUN_IGNORE_VERSION: 1` for the whole snapshot job is what we want, but I don't see why not.

Lotterleben:
why "Might"?

Urhengulas:
> why "Might"?

I still had the question if we want to replace whole `build.rs` with this change. This good resolved with jorge in the issue.

build.rs
```diff
@@ -1,56 +1,49 @@
...
-    } else {
-        assert!(
-            !Path::new(".git").exists(),
-            "you need to install the `git` command line tool to use the git version of `defmt`"
-        );
-
-        // no git info -> assume crates.io
-        let semver = Version::parse(&std::env::var("CARGO_PKG_VERSION")?)?;
-        if semver.major == 0 {
+    const GIT_DESCRIBE: &str = git_version!(args = ["--long", "--match", "v*"], cargo_prefix = "");
```
Lotterleben, line 7: what happens here if the user doesn't have git installed (as previously caught by https://github.com/knurling-rs/defmt/pull/518/files#diff-d0d98998092552a1d3259338c2c71e118a5b8343dd4703c0c7f552ada7f9cb42L24)?

build.rs
```diff
@@ -1,56 +1,49 @@
...
+    let version = match Version::parse(GIT_DESCRIBE) {
```
Lotterleben, line 8: super nitpick: the match body reads a bit unintuitive if you're not aware that `Version::parse()` parses *semver* strings. I'd keep the full crate qualifier here, e.g.

```suggestion
    let version = match semver::Version::parse(GIT_DESCRIBE) {
```
(and remove the import at the top)

Lotterleben (approved):

build.rs
Urhengulas: Since we specify `cargo_prefix`, it will use the `CARGO_PKG_VERSION` (see the third bullet point https://docs.rs/git-version/0.3.4/git_version/macro.git_version.html)

build.rs
Urhengulas: 👍🏾

Urhengulas: Addressed in 9d5f5fe

Urhengulas:
I will test this a bit to make sure it doesn't create a lot of version-mismatch errors.

PR #519, opened 2021-06-24
target-side `env_logger`-like env filter

### LINKED ISSUE #183: emit filter (target side): mechanism to filter dev and release profiles differently
use case: let's say you want "trace" level and up under the dev profile and "warn" level and up under the release profile. This cannot be accomplished today with the "defmt-*" cargo features. The closest is "defmt-default" which enables "trace" and up under dev and **"info"** and up under release

prior art: the `log` crate provides two sets of Cargo features, one for dev and one for release, to provide this feature. See https://docs.rs/log/0.4.11/log/#compile-time-filters . We could do the same and add a second set of Cargo features, e.g. "defmt-release-$level"; the existing set of features ("defmt-info") would then apply to the dev profile. So the example use case would be covered by this mix of Cargo features:

``` toml
[features]
default = ["dependency-a/defmt-trace", "dependency-a/defmt-release-warn"]
```

unresolved questions:
- "defmt-default" would have no matching pair; mixing "defmt-default" and "defmt-release-trace" should maybe be an error?
- should we rename the Cargo features to include the phrase "max_level" that the `log` uses?

### LINKED ISSUE #255: Alternatives to Cargo feature proliferation
defmt requires that every library using it declare 6 Cargo features to allow choosing logging levels, `defmt-default`, `defmt-trace`, `defmt-debug`, `defmt-info`, `defmt-warn`, `defmt-error`.

It would be nice if there was another solution that didn't require this whole stampede of Cargo features. It might be possible to use a compile-time environment variable for this that works similar to the well-known `RUST_LOG`.

I'm not entirely sure how this would be accomplished though. It seems that we'd need build scripts that inspect the variable and use `rerun-if-env-changed`, and output the right `rustc-cfg` directives. I'm hoping that this can be accomplished without forcing every user to define a `build.rs`.

### LINKED ISSUE #459: Controlling logging more granularly than crate-level
I have a project that has multiple processes running at once (16 RTIC tasks). To get enough context to solve an issue I often need to log very verbose things. Some of the logging is temporary but some of it I spend a lot of time refining so it remains useful when I come back to work on another problem with that module. 

I can't just have the logging always enabled. I have plenty of spare CPU and flash, but RTT gets blocked causing my real-time tasks to fail and there isn't a good way to filter log messages on the host side so it would be far too much noise. From what I can tell there's currently no way to enable/disable logging at anything less than crate granularity. It would be very onerous to break up my crate into smaller crates just for logging.

What I'm currently doing is creating a cargo feature to control verbose logging in each module and using a macro like this:

```rust
macro_rules! trace_usb {
    ($str:literal) => {
        if cfg!(feature = "trace-usb") {
            defmt::trace!($str);
        }
    };
    ($str:literal, $($arg:expr),+) => {
        if cfg!(feature = "trace-usb") {
            defmt::trace!($str, $($arg),+);
        } else {
            $( let _ = $arg; )+ // To avoid unused variable warnings
        }
    };
}
```

This works okay, but it's a lot of boilerplate and the log lines now point to the macro declaration rather than the use site. I'd prefer something like this:

```rust
defmt::log_if!(feature = "trace-usb", "Foobar {:?}", x);
// or just
defmt::log_if!("trace-usb", "Foobar");
```

https://github.com/knurling-rs/defmt/issues/255 discussed things related to this, but its focused on *not* using cargo features and is stalled on that.

I propose something easier: just use cargo features for now. Even more of them.

japaric:
This PR implements an alternative approach to disabling/enabling logging in crates. Currently we use [Cargo features](https://defmt.ferrous-systems.com/setup-library.html).
This PR removes the need for those Cargo features and replaces them with the env variable: `DEFMT_LOG`.

The `DEFMT_LOG` env var specifies which crates should emit defmt logs and at which level. Its syntax is the same as the syntax used by the [`env_logger`](https://docs.rs/env_logger/0.8.4/env_logger/) crate.

~~The `DEFMT_LOG` env var parser currently rejects paths that include modules: so `DEFMT_LOG=krate=info` is OK, but `DEFMT_LOG=krate::module=info` is rejected. This is intentional and should let us accept module paths and filter based on modules in the future (as in add module support in a backwards compatible fashion).~~ No longer true, `DEFMT_LOG` now supports module paths.

One of the main advantages of `DEFMT_LOG` is that allows filtering logs with module level granularity. The current Cargo features based approach is more coarse grained and only supports enabling / disabling entire crates.

*NOTE* this is a *target* side *emit* filter and orthogonal to the *host* side *display* filter proposed in knurling-rs/probe-run#74. Disabling logs with the target filter makes the target program smaller (, maybe also slightly faster) and it reduces the amount of data sent from the target to the host. Disabling logs with the host filter does not change the target program at all not it changes the amount of data transmitted from the target to the host.

Changing the value of the `DEFMT_LOG` env var causes all crates that transitively depend on the `defmt_macros` crate to be recompiled.

Example usage:
``` console
$ bat src/bin/hello.rs
```
``` rust
fn main() -> ! {
    defmt::info!("hello");
    foo::foo();
    bar::bar();
    app::exit()
}

mod foo {
    pub fn foo() {
        defmt::info!("foo");
    }
}

mod bar {
    pub fn bar() {
        defmt::info!("bar");
    }
}
```
``` console
$ DEFMT_LOG=hello::foo cargo r --bin hello
0 INFO  foo
└─ hello::foo::foo @ src/bin/hello.rs:16

$ DEFMT_LOG=hello::bar cargo r --bin hello
0 INFO  bar
└─ hello::bar::bar @ src/bin/hello.rs:22

$ DEFMT_LOG=hello::foo,hello::bar cargo r --bin hello
0 INFO  foo
└─ hello::foo::foo @ src/bin/hello.rs:16
1 INFO  bar
└─ hello::bar::bar @ src/bin/hello.rs:22

$ DEFMT_LOG=hello cargo r --bin hello
0 INFO  hello
└─ hello::__cortex_m_rt_main @ src/bin/hello.rs:8
1 INFO  foo
└─ hello::foo::foo @ src/bin/hello.rs:16
2 INFO  bar
└─ hello::bar::bar @ src/bin/hello.rs:22

$ rg '\[features' Cargo.toml || echo no Cargo features
no Cargo features
```

---

## how to test this

as this is not even merged into the `main` branch and breaking changes (with respect to defmt v0.2.0) have landed into `defmt`, the whole process is quite involved

1. install a `probe-run` that uses the decoder in this branch

``` console
$ git clone https://github.com/knurling-rs/probe-run
$ cd probe-run
$ # change the `#[dependencies]` section
$ edit Cargo.toml
```

``` diff
 colored = "2.0"
-defmt-decoder = { version = "=0.2.2", features = ['unstable'] }
+defmt-decoder = { git = "https://github.com/knurling-rs/defmt", branch = "env-filter", features = ['unstable'] }
 difference = "2.0"
```

``` cansole
$ cargo install --path .
```

2. use `[patch.crates-io]` in `Cargo.toml` (the one in the root of the workspace) to use the git version of `defmt` instead of v0.2.x. The `app-template` already includes this section in [its `Cargo.toml`](https://github.com/knurling-rs/app-template/blob/52981902ebeb77d612635e7953882653e8019ebc/Cargo.toml#L73-L79)

``` toml
[patch.crates-io]
defmt = { git = "https://github.com/knurling-rs/defmt", branch = "env-filter" }
defmt-rtt = { git = "https://github.com/knurling-rs/defmt", branch = "env-filter" }
defmt-test = { git = "https://github.com/knurling-rs/defmt", branch = "env-filter" }
panic-probe = { git = "https://github.com/knurling-rs/defmt", branch = "env-filter" }
```

*IMPORTANT* use `branch = "env-filter"` instead of `rev = "somehash"`

3. set the `PROBE_RUN_IGNORE_VERSION` env variable before you run your firmware e.g.

``` console
$ PROBE_RUN_IGNORE_VERSION=1 cargo run --bin hello
```

4. try setting / changing the value of the `DEFMT_LOG` env variable!

``` console
$ export PROBE_RUN_IGNORE_VERSION=1
$ DEFMT_LOG=hello=trace cargo run --bin hello
0 INFO  Hello, world!
└─ hello::__cortex_m_rt_main @ src/bin/hello.rs:8
```

---

As there's other breaking-change work currently ongoing (#508) I plan to leave this work on hold for now -- but will be answering to feedback! -- and resume it when that's concluded.

## Unresolved questions

- rename `DEFMT_LOG` to `DEFMT_CAPTURE` or `DEFMT_TARGET_CAPTURE`? if we keep the `DEFMT_LOG` name, what should we name the env var to be used in knurling-rs/probe-run#74?

## TODO list
- [x] RFC approval
- [x] address inline TODO comments
- [x] support log level without module path e.g. `DEFMT_LOG=info`
- [x] support the `off` pseudo log-level (disables logging)
- [x] fix module path handling so that `DEFMT_LOG=krate::module` doesn't match `krate::module1`, `krate::module2`, etc.
- [x] fix CI
- [x] documentation. important changes in behavior to note
  - default log level is error (when DEFTM_LOG is not set)
  - compilation profile (`dev` / `profile`) has no effect on the log level filter
  - ["this removes the ability to set a default log level"](https://github.com/knurling-rs/defmt/pull/519#issuecomment-868869284)

closes #183 
closes #255
closes #459


macros/src/env_filter.rs
```diff
@@ -0,0 +1,132 @@
...
+    level: Level,
+}
+
+#[cfg(test)]
+mod tests {
+    use super::*;
+
+    // TODO unclear if we want the same behavior as `env_logger`
+    #[test]
+    fn when_duplicates_entries_in_defmt_log_use_last_entry() {
```
japaric, line 110: actually, now I see that this an override mechanism (from right to left) which is a reasonable thing to have. For example, `RUST_LOG=info,krate=trace` means: "emit info level logs from all crates, but (override) for the `krate` emit trace level logs"

japaric:
one big difference between this approach and the current one is that: which profile (`dev` vs `release`) is used to compile the code has *no* effect on the target-side filtering with the `env_logger` approach. Unsure if that's good or bad but it's different.

``` console
$ # see issue description example

$ # defmt 0.2.0
$ cargo r --bin hello
0 ERROR a
1 WARN  b
2 INFO  c
3 DEBUG d
4 TRACE e

$ # defmt 0.2.0
$ cargo r --bin hello --release
0 ERROR a
1 WARN  b
2 INFO  c

$ # this PR
$ cargo r --bin hello
0 ERROR a

$ # this PR
$ cargo r --bin hello --release
0 ERROR a
```

jonas-schievink:
I think another important thing to note is that this removes the ability to set a default log level. This is potentially useful when writing tests, when you don't want to set the same environment variable all the time. (something like dotenv can work around this, but that seems suboptimal, since it's a third-party tool)

Urhengulas:
Looks great! Could you please elaborate how the default levels and following line fit together? I am a bit confused by this 🙈 
https://github.com/knurling-rs/defmt/blob/688f3ebcc6e5cc1320854f3c57e85522ff2dd776/macros/src/lib.rs#L479

Urhengulas:
> I think another important thing to note is that this removes the ability to set a default log level. This is potentially useful when writing tests, when you don't want to set the same environment variable all the time. (something like dotenv can work around this, but that seems suboptimal, since it's a third-party tool)

You mean that one can't set a default for your own crate? Cause afai understand there are still some default levels. Do you see any other use cases where this would be a "problem" (or annoyance), except testing?

jonas-schievink:
Right, there's a fixed default to `error` messages. Previously you could set default Cargo features to configure any default you like (and when testing it's useful to enable more than just errors, for example the defmt-test progress messages are at `info` level).

japaric:
> Could you please elaborate how the default levels and following line fit together? I am a bit confused by this 

that line of code means log at this level or *higher* "severity". For instance if you set `DEFMT_LOG=info` you get info, warn and error level logs. If you don't set the env var you get only error level logs. In code, the "severity" in defined in the enum declaration (note the derive(PartialOrd)):
https://github.com/knurling-rs/defmt/blob/fde2aab0bc541081d7def42fb049a36fa793a2bc/parser/src/lib.rs#L137-L144
So trace < debug < info < warn < error in terms of severity. This is the same as the `log` crate.

japaric:
> I think another important thing to note is that this removes the ability to set a default log level. 

good point. I have noted this down in the issue description as a reminder to add it to the documentation.

thinking about the case of testing, in `std` code one can use the following code as a "setup" for unit tests:
``` rust
std::env::set_var("RUST_LOG", "my_crate=trace");
env_logger::init();
```

this is not possible with `DEFMT_LOG` because the filtering happens at compile time, not runtime.

in the particular case, of [testing embedded code](https://ferrous-systems.com/blog/test-embedded-app/) the only way to set `DEFMT_LOG` I can think of is setting it in xtasks. So `cargo xtask test target` would do `DEFMT_LOG=my_crate=trace cargo test -p fw-tests`; `cargo xtask test host-target` could do `DEFMT_LOG=my_crate=trace cargo run -p firmware; cargo test -p host-target-tests` etc. Using xtasks it's certainly more heavyweight than what one can do today but if you already set up xtask for testing your embedded code then it doesn't seem that much extra work.

Urhengulas:
> > Could you please elaborate how the default levels and following line fit together? I am a bit confused by this
> 
> that line of code means log at this level or _higher_ "severity". For instance if you set `DEFMT_LOG=info` you get info, warn and error level logs. If you don't set the env var you get only error level logs. In code, the "severity" in defined in the enum declaration (note the derive(PartialOrd)):
> 
> https://github.com/knurling-rs/defmt/blob/fde2aab0bc541081d7def42fb049a36fa793a2bc/parser/src/lib.rs#L137-L144
> 
> So trace < debug < info < warn < error in terms of severity. This is the same as the `log` crate.

Okay, I think I got it now:
- if you don't mention your crate in `DEFMT_LOG` you only get `error` logs
- if you just mention it without setting an explicit level you get `trace`
- if you mention it with a level you get this one

Correct? I guess I was mainly confused about the two different defaults.

japaric:
> Correct? I guess I was mainly confused about the two different defaults.

yes, that is correct. The "defaults" match the behavior of the [`env_logger`](https://docs.rs/env_logger/0.8.4/env_logger/) crate. (TBH I didn't know about the `trace` default when you specify only the crate name nor the `off` pseudo-level until I started writing tests for this feature and had to read the `env_logger` docs in more detail though I have used the crate several times before)

japaric:
pushed *very WIP* module path support. This works:
``` rust
fn main() -> ! {
    defmt::info!("hello");
    foo::foo();
    bar::bar();
    app::exit()
}

mod foo {
    pub fn foo() {
        defmt::info!("foo");
    }
}

mod bar {
    pub fn bar() {
        defmt::info!("bar");
    }
}
```
``` console
$ DEFMT_LOG=hello::foo cargo r --bin hello
0 INFO  foo
└─ hello::foo::foo @ src/bin/hello.rs:16

$ DEFMT_LOG=hello::bar cargo r --bin hello
0 INFO  bar
└─ hello::bar::bar @ src/bin/hello.rs:22

$ DEFMT_LOG=hello::foo,hello::bar cargo r --bin hello
0 INFO  foo
└─ hello::foo::foo @ src/bin/hello.rs:16
1 INFO  bar
└─ hello::bar::bar @ src/bin/hello.rs:22

$ DEFMT_LOG=hello cargo r --bin hello
0 INFO  hello
└─ hello::__cortex_m_rt_main @ src/bin/hello.rs:8
1 INFO  foo
└─ hello::foo::foo @ src/bin/hello.rs:16
2 INFO  bar
└─ hello::bar::bar @ src/bin/hello.rs:22
```

japaric:
status update: jonas-schievink pointed out that with this change, the `defmt-test` test-runner will not print its progress **unless** the `DEFMT_LOG` var is set to `info`, which is a regression in functionality of the default behavior.

To address the issue, I have proposed adding a `println!` macro that cannot be filtered in RFC #541. This would be used in `defmt-test` to preserve the progress output. Until a decision is made on that (by accepting it or proposing a different solution) this PR is blocked.

japaric:
Update: I have fleshed out the RFC text in the PR description and moved it to issue #571. This PR is blocked on that RFC being approved and `defmt::println!` being implemented.

japaric:
update: rebased the PR and addressed inline comments. this is still missing the `defmt_test` update indicated in the RFC (#571); that update requires `println!` (PR #569). The other thing that's missing is fixing the snapshot tests, which requires changes to the xtask (setting DEFMT_LOG there)

book/src/SUMMARY.md
```diff
@@ -3,7 +3,6 @@
 [Introduction](./introduction.md)
 - [User guide](./user-guide.md)
   - [Setup](./setup.md)
-    - [Applications](./setup-app.md)
```
Urhengulas, line 6: You probably also want to delete the file?

japaric: deleted

Urhengulas (approved):
Looks very good.

---

One question regarding file-structure: Shouldn't `macros/src/function_like/log.rs` be `macros/src/function_like/log/mod.rs`? That's the usual structure I know.

_(We can also fix this in a follow-up PR, since it seems that other modules in `macros` also not comply with it)_

japaric:
bors r=Urhengulas

japaric:
rebased
bors r=Urhengulas

PR #521, opened 2021-06-27
[3/n] Remove u24

Dirbaio:
Part 3 of N of #492 

- Now that we're going to use rzCOBS for encoding the stream, extra zero bytes are not that expenseive. Using a u32 instead of a u24 adds one zero byte, which when encoded is just 1 extra bit.
- Users are unlikely to use u24, as it's quite obscure (I didn't know it existed until I found it while reading defmt's source).
- It makes the code more complicated, because it's not natively supported by Rust. In the code size optimizations of the macro codegen I'm working on, it really breaks the "symmetry" of the code.

Therefore I propose we remove it.

### DIFF
--- a/book/src/ser-integers.md
+++ b/book/src/ser-integers.md
@@ -3,16 +3,16 @@
 Integers will be serialized in little endian order using `to_le_bytes()`.
 `usize` and `isize` values will be subject to LEB128 compression.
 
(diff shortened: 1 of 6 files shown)


japaric (approved):
sounds reasonable to me!

Issue #434, opened 2021-03-23
μ (U+03BC) vs. µ (U+00B5)


Urhengulas:
We are using the micro sign (`µ`) for the `µs` display hint ([see defmt-book](https://defmt.ferrous-systems.com/hints.html)). If a user accidentally uses the greek letter mu (`μ`) which looks the same on most fonts, they get an not super helpful `error: unknown display hint: "μs"`.

It woud be nice to have an error message more along the lines of
>  error: unknown display hint: "μs"; did you mean "µs"?

MathiasKoch:
Why not just use `us` ?

japaric:
>Why not just use us ?

I would vote for accepting `us` (and maybe deprecating / removing the *mu* version in the future)

Urhengulas:
> Why not just use `us` ?

Agree!

PR #522, opened 2021-06-29
Replace `µs` hint with `us`

jonas-schievink:
Fixes https://github.com/knurling-rs/defmt/issues/434 via search-and-replace


Urhengulas:
Did you make sure to replace both `µ` (micro sign) and `μ` (small mu), in case we used both by accident?

jonas-schievink:
Good catch! Doesn't seem like we use the other one though, so this should be ready.

Urhengulas:
Perfect!

bors r+

PR #525, opened 2021-06-30
[6/n] Macro optimization

Dirbaio:
Part of #492. Depends on #524 #508 

I've been experimenting with optimizing the macro codegen to "batch" known-size values together before writing. Doesn't get the code size savings I expect on the QEMU tests, so WIP for now. Maybe requires more work, maybe it's not worth in the end.


Urhengulas:
Hi @Dirbaio, Are you still planning to work on this PR? 😁

Dirbaio:
It got sent to conflict hell by #531, so it requires a rewrite. I still want to do it, but don't have much free time so I don't know when. 

The optimizations don't break wire format, so it shouldn't be blocking for the 0.3 release though. It can be added in a 0.3.x release later.

Urhengulas:
I will close this PR for now. If you want to pick it up again just re-open it, or open a new one 😄

Issue #307, opened 2020-12-15
`bitflags!` support


jonas-schievink:
Currently, using `#[derive(Format)]` on a [`bitflags!`](https://docs.rs/bitflags)-generated type results in output that looks like `Bitflags { bits: 3564356 }`, which isn't very readable.

It would be nice if defmt offered a convenient way to format bitflags the same way as the generated `Debug` implementation. Doing this efficiently probably requires adding first-class support for bitflags.

jonas-schievink:
I think the simplest way to do this (in terms of additional API surface) would be to add a `bitflags!` macro that has the same interface as the one from the `bitflags` crate, and in fact just forwards its arguments to that (so that any improvements to the bitflags crate are picked up automatically).

That macro would then use some internal magic to also generate a an efficient `Format` impl.

jonas-schievink:
Ah, this is going to be much more difficult to implement efficiently since bitflags can be arbitrary constants, but a proc macro can't access the final value.

jonas-schievink:
Removing from the milestone as I don't think this will happen until 0.3

jonas-schievink:
I still really want this, since nicely-formatted bitflags are incredibly useful. I think one possible way to support this would be to create a `static` per bitflags constant that stores the corresponding mask, and then reading that back out (somehow) in the decoder. They can be stored in a separate section to not use up any defmt format indices. For example:

```rs
defmt::bitflags! {
    pub struct Flags: u8 {
        const LE_LIMITED_DISCOVERABLE = 1 << 0;
        const LE_GENERAL_DISCOVERABLE = 1 << 1;
    }
}
```

This would create 2 additional statics initialized to `1 << 0` and `1 << 1`, respectively. The symbol name would contain the name of the constant, so `LE_LIMITED_DISCOVERABLE` and `LE_GENERAL_DISCOVERABLE`. It also has to contain some unique identifier so that multiple bitflags types with the same constant names are supported, but `Symbol` already handles this.

The decoder would then read the static's initializers out of the binary (which is somewhat tricky, and not something we currently do, but should be implementable), and essentially run `bitflags`' `Debug` impl code at runtime: for each `const` defined for the bitflags type, check if all the bits are set, and if so, print the constant's name.

The runtime encoding of `Flags` would then be maximally efficient: just its format string index followed by `Flags`' raw bits.

PR #528, opened 2021-07-10
Support bitflags

jonas-schievink:
Closes https://github.com/knurling-rs/defmt/issues/307

TODO:
* [x] check that `#[cfg]` on bitflags consts works


macros/src/bitflags.rs
```diff
@@ -0,0 +1,153 @@
...
+        crate::symbol::disambiguator(),
+    );
+    let sym = mksym(&format_str, "bitflags", false);
+
+    let ident = &input.ident;
+    let ty = input.ty;
+    let flags = codegen.flags.iter().map(|f| &f.def);
+    let result = quote! {
+        const _: () = {
+            fn assert<T: defmt::export::UnsignedInt>() {}
```
japaric, line 35: doesn't upstream `bitflags!` already implement this check? or are we limiting what upstream `bitflags!` can do by only accepting unsigned ints here whereas upstream also accepts signed ints? (seems unlikely)

macros/src/bitflags.rs
```diff
@@ -0,0 +1,153 @@
...
+                FlagStatic { def }
+            })
+            .collect::<Vec<_>>();
+
+        Codegen { flags }
+    }
+}
+
+#[allow(dead_code)]
+struct Input {
```
japaric, line 107: it seems quite unfortunate to have to parse the `bitflags!` syntax but I don't see any other way: we need the flag names and values and there's no trait in bitflags that encodes that information (it wouldn't be possible to encode flag names in a trait).

macros/src/bitflags.rs
```diff
@@ -0,0 +1,153 @@
...
+                    "bitflags_value",
+                    &format!("{}::{}", input.ident, flag.ident),
+                )
+                .mangle();
+
+                let def = quote! {
+                    #[cfg_attr(target_os = "macos", link_section = ".defmt,end")]
+                    #[cfg_attr(not(target_os = "macos"), link_section = ".defmt.end")]
+                    #[export_name = #sym]
+                    static #name: u128 = {
```
japaric, line 89: wait, I'm confused about this. why is this `u128`? this static lives in the fake "defmt" address space so we should *not* be loading (LDR instructions) values that have "defmt" addresses. also, we want to keep those address apart from each other by a single byte, for compaction purposes (this is a 16-bit address space)

is the target using the `coerced_value`s?.

macros/src/bitflags.rs
```diff
@@ -0,0 +1,153 @@
...
+                let repr_ty = &input.ty;
+
+                let sym = Symbol::new(
+                    "bitflags_value",
+                    &format!("{}::{}", input.ident, flag.ident),
+                )
+                .mangle();
+
+                let def = quote! {
+                    #[cfg_attr(target_os = "macos", link_section = ".defmt,end")]
```
japaric, line 86: (the difference between these two cfg attributes is quite hard to spot: `.` vs `,`)

japaric (commented):
I have done a first pass on this PR and will continue tomorrow (probably) but I have commented on things that looked a bit odd to me.

decoder/src/elf2table/mod.rs
```diff
@@ -89,28 +90,68 @@ pub fn parse_impl(elf: &[u8], check_version: bool) -> Result<Option<Table>, anyh
...
+                symbol::SymbolTag::Defmt(Tag::BitflagsValue) => {
+                    // Bitflags values always occupy 128 bits / 16 bytes.
+                    if entry.size() != 16 {
+                        bail!(
+                            "bitflags value does not occupy 16 bytes (symbol `{}`)",
+                            name
                         );
                     }
+
+                    let defmt_data = defmt_section.data()?;
```
japaric, line 115: ah, I guess this is where the read the data stored in the `static` produce by `defmt::bitflags`. could we *not* put the value in the static but instead put it in its symbol name?

jonas-schievink: No, since that would require evaluating arbitrary constant expressions during proc macro evaluation

macros/src/bitflags.rs
jonas-schievink: The target just uses what the `bitflags` crate generates, I've chosen to cast the value to a `u128` here so that we can treat all underlying types the same (the decoder will store all unsigned integers in a `u128` already, so this is convenient).

macros/src/bitflags.rs
jonas-schievink: Yeah, this seems unavoidable. But I don't think it's a big issue, `syn` makes writing a parser for this extremely easy, and the `bitflags` syntax hasn't really changed much (if at all) since 1.0 was released.

macros/src/bitflags.rs
jonas-schievink: The bitflags crate does not check this type, its docs just note

> The flags should only be defined for integer types, otherwise unexpected type errors may occur at compile time.

japaric: then this is a great addition!

macros/src/bitflags.rs
japaric: I think it'd be more worried about new syntax being added (w/o it being a breaking change) and then us needing to update our bitflags parser to accommodate that, than having to support a new, totally different bitflags 2.0 syntax.

it seems very unlikely that syntax will grow however given that the last (patch) release was 2 years ago so now I'm not too worried about this.

macros/src/bitflags.rs
jonas-schievink: Ah, and to expand on this a bit: the address of these statics is irrelevant, they are never referenced by the program, and they don't contribute to the `.defmt` address space usage (in 9ad68bb, I've changed how we compute how much of that address space is in use to make them not count towards the limit).

All these statics end up in `.defmt.end`, which is just a metadata section (and should maybe be renamed to reflect that). We're only interested in their symbol names and value.

decoder/src/elf2table/mod.rs
japaric: > No, since that would require evaluating arbitrary constant expressions during proc macro evaluation

right, that's what I feared. ~~then, would it be possible to put the flag values in a *different* output linker section (e.g. `.defmt.bitflags`) so that each flag doesn't waste 15 (16? (+)) additional bytes of the main `.defmt` linker section?

(+) do we ever use the address of the `static`s defined in the `assert` function?

I guess before that question, the more relevant question is: could bitflag flags exhaust the 16-bit `.defmt` address space? You need ~4000 flags for that and you can have more than one flag per `bitflags!` invocation.~~

I see that you put these values after the format string indices in the linker script, which is roughly equivalent to having a separate section, so they won't consume the 16-bit `.defmt` address space. Never mind this comment then.

parser/src/lib.rs
```diff
@@ -50,12 +50,21 @@ pub enum DisplayHint {
     Debug,
     /// `:us`, formats integers as timestamps in microseconds
     Microseconds,
+    /// `__internal_bitflags_NAME` instructs the decoder to print the flags that are set, instead of
+    /// the raw value.
+    Bitflags {
+        name: String,
+        package: String,
+        disambiguator: String,
```
japaric, line 58: shouldn't the disambiguator  field (in this variant and the `BitFlagsKey` struct) be a `u64`? I think right now it's a stringified u64?

decoder/src/lib.rs
```diff
@@ -96,11 +105,21 @@ impl StringEntry {
     }
 }
 
+/// Data that uniquely identifies a `defmt::bitflags!` invocation.
+#[derive(Debug, PartialEq, Eq, Hash)]
+struct BitflagsKey {
+    /// Name of the bitflags struct (this is really redundant with `disambig`).
```
japaric, line 111: does it need to be in the key? (as in can we remove it? or are using the `ident` in the key in the formatting?). package + disambiguator should be enough to uniquely identify this in the entire dependency graph, right?

macros/src/bitflags.rs
```diff
@@ -0,0 +1,170 @@
...
+                };
+                FlagStatic { def }
+            })
+            .collect::<Vec<_>>();
+
+        Codegen { flags }
+    }
+}
+
+#[allow(dead_code)]
```
japaric, line 108: could you move the parsing implementation into its own module and e.g. have `Input` / `Flag` expose public accessors for the parts that codegen needs?

src/lib.rs
```diff
@@ -283,6 +283,9 @@ pub use defmt_macros::global_logger;
 /// ```
 pub use defmt_macros::timestamp;
 
+/// TODO docs
```
japaric, line 286: this still needs to be documented

decoder/src/frame.rs
```diff
@@ -160,12 +212,52 @@ fn format_args_real(
...
+                // `defmt::bitflags!`.
+                let key = BitflagsKey {
+                    ident: name.clone(),
+                    package: package.clone(),
+                    disambig: disambiguator.clone(),
+                };
+                match self.table.bitflags.get(&key) {
+                    Some(flags) => {
+                        // FIXME flag print order does not match definition order
+                        // (does some part of the toolchain alphabetically sort them?)
```
japaric, line 230: interesting observation: perhaps the linker sorts the symbols alphabetically within a section? maybe we could prefix the symbols (`00_$first`, `01_$second`, etc.) so that even after sorting alphabetically they stay in the declaration order? that doesn't have to be addressed in this PR but we should create an issue for this

decoder/src/frame.rs
```diff
@@ -70,68 +68,122 @@ impl<'t> Frame<'t> {
...
-        } else {
-            match self.frame.level {
-                Level::Trace => "TRACE".to_string(),
-                Level::Debug => "DEBUG".to_string(),
-                Level::Info => "INFO".to_string(),
-                Level::Warn => "WARN".to_string(),
-                Level::Error => "ERROR".to_string(),
-            }
-        };
+    fn format_args_real(
```
japaric, line 76: (lots of churn in this file; I'm going to just trust you on these changes)

japaric (commented):
Thanks for implementing this! I gave this a second pass and it looks good to me. Left some inline comment about the code. I'd block landing this on adding documentation (API + book); feel free to send this PR to bors when docs are in place.

Feel free to ping me if you'd like some more review on future changes.

macros/src/symbol.rs
```diff
@@ -70,3 +69,17 @@ fn escape(s: &str) -> String {
...
 }
+
+pub fn package() -> String {
+    env::var("CARGO_PKG_NAME").unwrap_or_else(|_| "<unknown>".to_string())
+}
+
+pub fn disambiguator() -> u64 {
+    // We want a deterministic, but unique-per-macro-invocation identifier. For that we
+    // hash the call site `Span`'s debug representation, which contains a counter that
+    // should disambiguate macro invocations within a crate.
```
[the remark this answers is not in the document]
jonas-schievink: Filed https://github.com/knurling-rs/defmt/issues/532 – and noted a related issue with renaming dependencies there

parser/src/lib.rs
jonas-schievink: In the ELF / "defmt ABI", the disambiguator field is a string. We do currently always set it to a stringified `u64`, but I'd prefer to treat it as an opaque string instead.

decoder/src/lib.rs
jonas-schievink: Yeah, I mostly added that because I don't really like that we rely on `Span`'s hash identity in the disambiguator field, and wanted to add something more than that

decoder/src/frame.rs
jonas-schievink: Most of them are mechanical to thread `&Table` through

macros/src/bitflags.rs
jonas-schievink: Moved it to a submodule, but kept it in the same file

japaric (approved):
still looks good to me

PR #531, opened 2021-07-15
refactor the `macros` crate

japaric:
~~NOT READY FOR REVIEW. I just wanted to signal that I'm working on this~~


japaric:
(done with the refactor but now this needs a rebase. will rebase later and mark the PR as 'ready for review')

macros/Cargo.toml
```diff
@@ -18,6 +18,7 @@ unstable-test = []
 
 [dependencies]
 defmt-parser = { path = "../parser", features = ["unstable"], version = "=0.2.2" }
+proc-macro-error = "1.0.4"
 proc-macro2 = "1.0.27"
 quote = "1.0.9"
-syn = { version = "1.0", features = ["full"] }
+syn = { version = "1.0", features = ["full"] }
```
jonas-schievink, line 24: Missing newline (should we check this in CI? it crops up fairly often)

```suggestion
syn = { version = "1.0", features = ["full"] }

```

jonas-schievink: I don't think the GH suggestion works correctly here (it will add 2 newlines)

jonas-schievink (approved):
Looks good, r=me with added newline

macros/Cargo.toml
japaric: I think that's my editor settings. ( Been meaning to do something about it for a while because it seems to do the opposite of what other editors do :upside_down_face: ) checking it in CI seems fine to me

PR #539, opened 2021-07-17
Add optional rzCOBS encoding+framing

Dirbaio:
Part of #492 

Encoding is chosen with the `encoding-raw`, `encoding-rzcobs` features. If no encoding is chosen, `rzcobs` is used.

The used encoding is stored in the ELF symbol table, and the decoder automatically picks the correct version.


decoder/src/elf2table/mod.rs
```diff
@@ -69,9 +94,18 @@ pub fn parse_impl(elf: &[u8], check_version: bool) -> Result<Option<Table>, anyh
     };
 
     if check_version {
-        self::check_version(version).map_err(anyhow::Error::msg)?;
+        self::check_version(&version).map_err(anyhow::Error::msg)?;
     }
 
+    let encoding = match encoding {
+        Some(e) => match &e[..] {
```
japaric, line 101: nit: could you move this parsing into a `FromStr` implementation for `Encoding`?

decoder/src/stream/rzcobs.rs
```diff
@@ -0,0 +1,95 @@
...
+use crate::{DecodeError, Frame, Table};
+
+/// Decode a full message.
+///
+/// `data` must be a full rzCOBS encoded message. Decoding partial
+/// messages is not possible. `data` must NOT include any `0x00` separator byte.
+fn rzcobs_decode(data: &[u8]) -> Result<Vec<u8>, DecodeError> {
+    let mut res = vec![];
+    let mut data = data.iter().rev().cloned();
+    while let Some(x) = data.next() {
```
japaric, line 11: isn't there a crate we could depend on for this decoding part?

src/encoding/rzcobs.rs
```diff
@@ -0,0 +1,267 @@
...
+//   nnnnnnnn => output n-1 bytes from stream, output 0x00
+//   11111111 => output 254 bytes from stream
+//
+// zCOBS/rzCOBS:
+//   00000000 => end of frame
+//   0xxxxxxx => foreach x from LSB to MSB: if x=0 output 1 byte from stream, if x=1 output 0x00
+//   1nnnnnnn => output n+7 bytes from stream, output 0x00
+//   11111111 => output 134 bytes from stream
+
+pub(crate) struct Encoder {
```
japaric, line 12: isn't there a crate we can depend on for the encoding functionality?

firmware/defmt-rtt/src/lib.rs
```diff
@@ -27,6 +27,7 @@ struct Logger;
 
 static TAKEN: AtomicBool = AtomicBool::new(false);
 static INTERRUPTS_ACTIVE: AtomicBool = AtomicBool::new(false);
+static mut ENCODER: defmt::Encoder = defmt::Encoder::new();
```
japaric, line 30: I think we should wrap this in a `RacyCell` to avoid potential UB. As it is, I think LLVM will mark this with `noalias` (?) metadata which may not be the right thing for a shared global. I believe all Atomic* types internally use UnsafeCells to avoid issues like that (and they still have unsolved UB issues like `rust-lang/rust 55005`; though that is due to the "deferenceable" attribute). cc @jonas-schievink to sanity check my claim here

I don't think there's a `RacyCell` on crates.io but it basically is a `Sync` UnsafeCell:
``` rust
pub struct RacyCell<T>(core::cell::UnsafeCell<T>);

// this type has no synchronization; synchronization must be enforced manually
unsafe impl<T> Sync for RacyCell<T> {}

impl<T> RacyCell<T> {
    pub const fn new(val: T) -> Self {
        Self(core::cell::UnsafeCell::new(val))
    }

    pub fn get(&self) -> *mut T {
        self.0.get()
    }
}

static UNSYNCHRONIZED_THING: RacyCell<NotPlainOldDataType> =
    RacyCell::new(NotPlainOldDataType::new());

fn usage() {
    // disable interrupts
    unsafe { (*UNSYNCHRONIZED_THING.get()).mutating_method() }
    // enable interrupts
}
```

japaric (commented):
left some inline comments / questions. does `probe-run` need changes after this has landed? IOW, will landing this break (e.g. not compile / fail to decode) `probe-run` even if it uses the right decoder version?

firmware/defmt-rtt/src/lib.rs
jonas-schievink: I don't think this is necessary. `UnsafeCell` only enables interior mutability, but `static mut` is already mutable. It just needs synchronization to ensure no 2 `&mut` to it ever exist at the same time, but `UnsafeCell` also needs that.

japaric: > It just needs synchronization to ensure no 2 &mut to it ever exist at the same time, but UnsafeCell also needs that.

I agree with the Rust side of things; I was more worried with the less visible LLVM codegen side of things but I checked and see that `rustc` is not emitting `noalias` (ATM at least) and that `dereferenceable` is applied to _all_ mutable references regardless of whether they come from `UnsafeCell` or not so sticking an `UnsafeCell` into a `static` vs using a plain `static mut` produces the same LLVM IR.

Never mind me then. I'm OK with keeping this (`static mut` version) as it is.

jonas-schievink: `noalias` is not applied to `&mut` references due to LLVM bugs, it will be applied once those are fixed. But from looking at the code it seems that even that wouldn't cause problems, since it never creates conflicting references to the static.

src/encoding/rzcobs.rs
Dirbaio: There's the `rzcobs` crate, but the API is not quite right. The Encoder needs to take ownership of a Writer which is somewhat annoying for Logger impls. Also, it processes stuff byte-by-byte, while handling entire byte slices allows for more optimizations (not in the current code though). 

I thought integrating it here would be easier and more flexible, since it's not that much code after all. Can try to change it to use `rzcobs` if wanted.

decoder/src/stream/rzcobs.rs
Dirbaio: Yes, this is from the `rzcobs` crate. For encoding it's not quite right, for decoding we could use it as-is. I thought it was strange to have an in-tree impl for encoding and a crate for decoding, but can change it if you think it's better to use external crates.

Dirbaio:
> will landing this break (e.g. not compile / fail to decode) probe-run even if it uses the right decoder version?

I think it'll still build because I've left `Table::decode` as-is. Since the default is `rzcobs`, it will fail to decode, though. It should switch to the StreamDecoder API.

> EDIT: would sticking to the "raw" encoder / decoder for a bit not immediately break probe-run?

Yes, `raw` is identical to the previous behavior, just concatenating all frames with no framing. But since it'll need a release anyway to update to the new `defmt-decoder` (wire format has breaking changes) it might as well be updated to StreamDecoder from day 1, the required changes are tiny.

japaric:
> Yes, this is from the rzcobs crate. For encoding it's not quite right, for decoding we could use it as-is. I thought it was strange to have an in-tree impl for encoding and a crate for decoding, but can change it if you think it's better to use external crates.

In that case, it seems fine to "vendor" that code. This looks good to me.

bors r+

decoder/src/elf2table/mod.rs
```diff
@@ -13,17 +13,37 @@ use std::{
...
+                    .trim_start_matches("_defmt_version_ = ")
+                    .trim_end_matches('"')
+                    .to_string(),
+            )
+        } else {
+            None
+        }
+    };
+
+    // No need to remove quotes for `_defmt_encoding_`, since it's
```
Urhengulas, line 41: @Dirbaio  Is this line missing the end of the sentence?

src/encoding/mod.rs
```diff
@@ -0,0 +1,37 @@
...
+
+// This wrapper struct is to avoid copypasting the public docs in all the impls.
+
+/// TODO
+pub struct Encoder {
+    inner: inner::Encoder,
+}
+
+impl Encoder {
+    /// TODO
```
Urhengulas, line 16: Also could you please add the missing doc-comments?

Dirbaio:
Added missing docs in #568

decoder/src/elf2table/mod.rs
Dirbaio: Fixed in #568

decoder/src/elf2table/mod.rs
Urhengulas: Went with `TryFrom<Option<String>>` in https://github.com/knurling-rs/defmt/pull/572

Issue #426, opened 2021-03-06
It's impossible to support multiple versions of defmt in a crate


jonas-schievink:
If you do this:

```toml
[dependencies]
defmt01 = { version = "0.1.0", package = "defmt", optional = true }
defmt02 = { version = "0.2.0", package = "defmt", optional = true }
```

it results in Cargo complaining:

```
error: failed to select a version for `defmt`.
    ... required by package `bxcan v0.4.0 (/home/jonas/dev/bxcan)`
versions that meet the requirements `^0.1.0` are: 0.1.3, 0.1.2, 0.1.1, 0.1.0

the package `defmt` links to the native library `defmt`, but it conflicts with a previous package which links to `defmt` as well:
package `defmt v0.2.0`
    ... which is depended on by `bxcan v0.4.0 (/home/jonas/dev/bxcan)`
```

This is because defmt makes use of the `links` key in Cargo.toml to prevent linking multiple versions of it into the same executable. It looks like Cargo handles this very poorly and demands that even optional dependencies don't use the same `links` directive (we should probably check if this is a Cargo bug).

Lotterleben:
Note that this has the side effect that, given this dependency tree:

```
        defmt
         |       |
 libuseful  |
         |       |
        my_crate
```
e.g. `my_crate` depending on `defmt` and `libuseful`, and `libuseful` using `defmt` as well, `my_crate` is *forced* to update their `defmt` dep (and potentially rewrite all of their log statements) if `libuseful` bumps their defmt version

jonas-schievink:
Found the upstream issue that's causing this bug: https://github.com/rust-lang/cargo/issues/7880

Dirbaio:
Have you considered excluding the wire format from the semver guarantees? 

Minor bumps like 0.2.2 -> 0.2.3 would be guaranteed to keep API compatibility (so existing code will still compile), but are allowed to do breaking changes to the wire format.

Probe-run could support decoding many wire format versions at once (for example by depending on multiple defmt-decoder versions). Doing a minor defmt upgrade would cause the user to get prompted to upgrade probe-run, but no other breakage should happen.

This would allow defmt to evolve faster without splitting the ecosystem. API breaking changes should be very rare.

Also you guys mentioned that extending the format string is a breaking change. Why is that? Is it due to old decoders now being unable to decode the stream? If so, that'd be solved by this too. Extending the format string would break wire format but not API.

jonas-schievink:
[...]

PR #540, opened 2021-07-18
Separate "crate version" from "wire format version" 

### LINKED ISSUE #287: Separate "crate version" from "wire format version"
Currently the `defmt` crate version is embedded in the ELF, and is used by the decode to check compatibility of the wire format. The version is the crates.io version or the Git commit hash if building from a git clone.

Probe-run refuses to run when the version doesn't match exactly. You have to clone it, patch it to use the right defmt version, rebuild it. This is quite painful in these scenarios:

- When developing defmt, trying to run a firmware using the WIP modified version.
- When releasing a lib / project that depends on git defmt, for people trying to use it.

Many (most?) changes don't break wire format. Even API-breaking changes like #258 can not-break the wire format. 

Why not separate the "wire format version" concept from the "crate version"?

- `defmt` crate has a WIRE_FORMAT_VERSION const. 
- `defmt-decoder` crate has a WIRE_FORMAT_VERSION const. 
- `defmt-decoder` checks the WIRE_FORMAT_VERSION from the ELF matches its const, fails if not.
- the comparison can be a simple string equality, no need for semver stuff I think.
- When doing wire-format breaking changes, the wire format version gets bumped

Advantages:

- No need for build.rs magic at all.
- Easier hacking, easier use of git defmt versions that are still compatible with released wire format version
- possibility of doing a major version `defmt` crate release that doesn't bump the wire format version. (ie with API-breaking changes but not with wire-format-breaking changes)

Dirbaio:
Depends on #539. Fixes #287.

This decouples the "wire format version" from the crate version (Cargo package version or git commit). Wire format version can now be bumped independently. This alleviates pains listed in #539., and more importantly it will allow making breaking changes to the wire format in a semver-minor `defmt` crate release. See #426.

Version is `3` to be able to retroactively number `0.1.x = 1, 0.2.x = 2`. A followup PR will add support to `defmt-decoder` to decode both version 2 and 3 concurrently.


Urhengulas:
Hi @Dirbaio, Can you please resolve the merge conflict? 😁

Urhengulas:
Heh, your change makes https://github.com/knurling-rs/defmt/pull/518 pretty-much obsolete 🙈 I will revert it.

src/lib.rs
```diff
@@ -18,6 +18,27 @@
 #[cfg(feature = "alloc")]
 extern crate alloc;
 
+// This must be in the root lib.rs, otherwise it doesn't appear in the final binary.
+
+#[used]
+#[cfg_attr(target_os = "macos", link_section = ".defmt,end.VERSION")]
+#[cfg_attr(not(target_os = "macos"), link_section = ".defmt.end")]
+#[export_name = "_defmt_version_ = 3"]
+static DEFMT_VERSION: u8 = 0;
```
jonas-schievink, line 27: It would be good to have some docs on this.

```suggestion
/// The defmt ABI and wire format version.
///
/// This number has to be updated every time there is a backwards-incompatible change to
/// - the symbol naming scheme
/// - the symbol and section layout
/// - the data encoding / wire format
#[used]
#[cfg_attr(target_os = "macos", link_section = ".defmt,end.VERSION")]
#[cfg_attr(not(target_os = "macos"), link_section = ".defmt.end")]
#[export_name = "_defmt_version_ = 3"]
static DEFMT_VERSION: u8 = 0;
```

jonas-schievink (approved):
Looks good, though some in-code docs would be good to have here, since this is quite important to get right.

src/lib.rs
Urhengulas: Is there a way to catch these backwards-incompatible changes in the CI?

jonas-schievink: We can probably come up with a backwards-compat check again, yes

decoder/src/elf2table/mod.rs
```diff
@@ -13,17 +13,37 @@ use std::{
...
+                    .trim_start_matches("_defmt_version_ = ")
+                    .trim_end_matches('"')
+                    .to_string(),
+            )
+        } else {
+            None
+        }
+    };
+
+    // No need to remove quotes for `_defmt_encoding_`, since it's
```
jonas-schievink, line 41: Looks like we missed an incomplete sentence here

Dirbaio: Fixed in #568

PR #550, opened 2021-07-22
`defmt::flush()`

### LINKED ISSUE #515: defmt: flush API
# What

`defmt_rtt::flush()` is a blocking API that waits until the host has read all the pending data in the target's (caller's) RTT buffer

# Use case

Avoid decoding errors when doing a software reset, or calling `cortex_m::asm::{bootstrap,bootload}`

For example, this likely produces a decoding error in `probe-run`

``` rust
defmt::info!("about to reset!");
SCB::sys_reset();
```

To work around the decoding error you can add a delay

``` rust
defmt::info!("about to reset!");
cortex_m::asm::delay(1_000_000);
SCB::sys_reset();
```

It's not obvious how large to make the delay so the `flush` API would be the preferred option

``` rust
defmt::info!("about to reset!");
defmt_rtt::flush();
SCB::sys_reset();
```

# Unresolved questions

- Can we put the API in the `defmt` crate by making it transport agnostic somehow?
  - the problem with a transport specific API is that it cannot be used in libraries unless one is OK with forcing all users of the libraries to use that specific transport
- in the `defmt_rtt` implementation, how do we prevent the issue of blocking forever if no host / debugger is attached?

Urhengulas:
Fixes #515 

## TODO
- [x] skeleton: add method to trait, TODO-implements in crates and tests
- [x] defmt-rtt
- [x] defmt-semihosting
- [x] defmt-itm
- [x] adapt delays based on docs and testing

### DIFF
--- a/firmware/defmt-itm/src/lib.rs
+++ b/firmware/defmt-itm/src/lib.rs
@@ -74,7 +74,7 @@ unsafe impl defmt::Logger for Logger {
 
         // delay "a bit" to drain the queue
         // This is a heuristic and might be too short in reality. Please open an issue if it is!
-        asm::delay(100); // TODO: how long does it take to drain?
+        asm::delay(100);
     }
 
     unsafe fn release() {
--- a/firmware/defmt-itm/src/lib.rs
+++ b/firmware/defmt-itm/src/lib.rs
@@ -16,7 +16,11 @@
 
 use core::sync::atomic::{AtomicBool, Ordering};
 
-use cortex_m::{interrupt, itm, peripheral::ITM, register};
+use cortex_m::{
+    asm, interrupt, itm,
+    peripheral::{itm::Stim, ITM},
+    register,
+};
 
 #[cfg(armv6m)]
 compile_error!(
@@ -65,7 +69,11 @@ unsafe impl defmt::Logger for Logger {
     }
 
     unsafe fn flush() {
-        todo!()
+        // wait for the queue to be able to accept more data
+        while !stim_0().is_fifo_ready() {}
+
+        // delay "a bit" to drain the queue
+        asm::delay(100); // TODO: how long does it take to drain?
     }
 
     unsafe fn release() {
@@ -88,5 +96,13 @@ unsafe impl defmt::Logger for Logger {
 fn do_write(bytes: &[u8]) {
     // NOTE(unsafe) this function will be invoked *after* `enable` has run so this crate now has
     // ownership over the ITM thus it's OK to instantiate the ITM register block here
-    unsafe { itm::write_all(&mut (*ITM::ptr()).stim[0], bytes) }
+    unsafe { itm::write_all(stim_0(), bytes) }
+}
+
+/// Get access to stimulus port 0
+///
+/// # Safety
+/// Can only be invoked *after* `enable` has run
+unsafe fn stim_0<'a>() -> &'a mut Stim {
+    &mut (*ITM::PTR).stim[0]
 }
--- a/firmware/defmt-itm/src/lib.rs
+++ b/firmware/defmt-itm/src/lib.rs
@@ -73,6 +73,7 @@ unsafe impl defmt::Logger for Logger {
         while !stim_0().is_fifo_ready() {}
 
         // delay "a bit" to drain the queue
+        // This is a heuristic and might be too short in reality. Please open an issue if it is!
         asm::delay(100); // TODO: how long does it take to drain?
     }
 
--- a/firmware/defmt-itm/src/lib.rs
+++ b/firmware/defmt-itm/src/lib.rs
@@ -64,6 +64,10 @@ unsafe impl defmt::Logger for Logger {
         unsafe { ENCODER.start_frame(do_write) }
     }
 
+    unsafe fn flush() {
+        todo!()
+    }
+
     unsafe fn release() {
         // safety: accessing the `static mut` is OK because we have disabled interrupts.
         ENCODER.end_frame(do_write);
--- a/firmware/defmt-rtt/src/lib.rs
+++ b/firmware/defmt-rtt/src/lib.rs
@@ -15,7 +15,7 @@
 //!
 //! As an effect this implementation may block forever if `probe-run` disconnects on runtime. This
 //! is because the RTT buffer will fill up and writing will eventually halt the program execution.
-//! 
+//!
 //! `defmt::flush` would also block forever in that case.
 
 #![no_std]
@@ -105,22 +105,17 @@ const NOBLOCK_TRIM: usize = 1;
 
 impl Channel {
     fn write_all(&self, mut bytes: &[u8]) {
-        // NOTE `flags` is modified by the host after RAM initialization while the device is halted
-        // it cannot otherwise be modified so we don't need to check its state more often than
-        // just here
-        if host_is_connected(self) {
-            while !bytes.is_empty() {
-                let consumed = self.blocking_write(bytes);
-                if consumed != 0 {
-                    bytes = &bytes[consumed..];
-                }
-            }
-        } else {
-            while !bytes.is_empty() {
-                let consumed = self.nonblocking_write(bytes);
-                if consumed != 0 {
-                    bytes = &bytes[consumed..];
-                }
+        // the host-connection-status is only modified after RAM initialization while the device is
+        // halted, so we only need to check it once before the write-loop
+        let write = match host_is_connected(self) {
+            true => Channel::blocking_write,
+            false => Channel::nonblocking_write,
+        };
+
+        while !bytes.is_empty() {
+            let consumed = write(self, bytes);
+            if consumed != 0 {
+                bytes = &bytes[consumed..];
             }
         }
     }
@@ -234,5 +229,6 @@ unsafe fn handle() -> &'static Channel {
 }
 
 fn host_is_connected(channel: &Channel) -> bool {
+    // we assume that a host is connected if we are in blocking-mode. this is what probe-run does.
     channel.flags.load(Ordering::Relaxed) == BLOCK_IF_FULL
 }
--- a/firmware/defmt-rtt/src/lib.rs
+++ b/firmware/defmt-rtt/src/lib.rs
@@ -47,6 +47,10 @@ unsafe impl defmt::Logger for Logger {
         unsafe { ENCODER.start_frame(do_write) }
     }
 
+    unsafe fn flush() {
+        todo!()
+    }
+
     unsafe fn release() {
         // safety: accessing the `static mut` is OK because we have disabled interrupts.
         ENCODER.end_frame(do_write);
--- a/firmware/defmt-rtt/src/lib.rs
+++ b/firmware/defmt-rtt/src/lib.rs
@@ -11,7 +11,7 @@
 //!
 //! # Blocking/Non-blocking
 //!
-//! `probe-run` puts RTT into blocking-mode, to avoid loosing data.
+//! `probe-run` puts RTT into blocking-mode, to avoid losing data.
 //!
 //! As an effect this implementation may block forever if `probe-run` disconnects on runtime. This
 //! is because the RTT buffer will fill up and writing will eventually halt the program execution.
--- a/firmware/defmt-rtt/src/lib.rs
+++ b/firmware/defmt-rtt/src/lib.rs
@@ -48,7 +48,8 @@ unsafe impl defmt::Logger for Logger {
     }
 
     unsafe fn flush() {
-        todo!()
+        // SAFETY: if we get here, the global logger mutex is currently acquired
+        handle().flush();
     }
 
     unsafe fn release() {
@@ -98,7 +99,7 @@ impl Channel {
         // NOTE `flags` is modified by the host after RAM initialization while the device is halted
         // it cannot otherwise be modified so we don't need to check its state more often than
         // just here
-        if self.flags.load(Ordering::Relaxed) == BLOCK_IF_FULL {
+        if host_is_connected(self) {
             while !bytes.is_empty() {
                 let consumed = self.blocking_write(bytes);
                 if consumed != 0 {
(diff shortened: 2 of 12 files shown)


macros/src/attributes/global_logger.rs
```diff
@@ -43,6 +43,12 @@ fn codegen(strukt: &ItemStruct) -> TokenStream {
             <#ident as defmt::Logger>::acquire()
         }
 
+        #[no_mangle]
+        // TODO: should this be unsafe?
```
japaric, line 47: I don't think it matters because this function won't be called directly and when it's called via FFI it will always be unsafe to call.

firmware/defmt-rtt/src/lib.rs
```diff
@@ -47,6 +47,18 @@ unsafe impl defmt::Logger for Logger {
         unsafe { ENCODER.start_frame(do_write) }
     }
 
+    fn flush() {
+        Self::acquire();
```
japaric, line 51: I think you want the `acquire` and `release` one level up in `defmt::flush`. Otherwise every implementation would need to repeat the acquire/release

src/traits.rs
```diff
@@ -84,6 +84,9 @@ pub unsafe trait Logger {
     /// Panics if already acquired in the current execution context. Otherwise it must never fail.
     fn acquire();
 
+    /// Block until host has read all pending data
+    fn flush();
```
japaric, line 88: this should be `unsafe` and have the same safety requirement as `write`: MUST be called after `acquire` and before `release`

firmware/defmt-rtt/src/lib.rs
```diff
@@ -172,6 +184,38 @@ impl Channel {
...
+        let prev_read = 0;
+        let mut cycle_count = 0;
+        const CHECK_CYCLE: usize = 100;
+
+        // busy wait
+        while read() != write() {
+            // did the write pointer progress in the last `CHECK_CYCLE` loop cycles?
+            if cycle_count % CHECK_CYCLE != 0 {
+                cycle_count += 1;
+            // TODO: Is it bad to call read a second time instead of using the value from the while-condition?
```
japaric, line 210: seems ... fine but not entirely sure. I would expect the same `read` value to be used throughout one `while` iteration though

japaric (commented):
replied to TODO questions

firmware/defmt-rtt/src/lib.rs
Urhengulas: > I would expect the same read value to be used throughout one while iteration though

I agree. This just means that I need to bind the variable, which isn't needed in the current implementation 😅

macros/src/attributes/global_logger.rs
Urhengulas: So I'll go with the current convention ans will also make it unsafe

firmware/defmt-rtt/src/lib.rs
```diff
@@ -172,6 +184,38 @@ impl Channel {
...
+    fn flush(&self) {
+        // TODO: What is the best `Ordering`? The current ones are taken from `fn blocking_write`.
+        let read = || self.read.load(Ordering::Relaxed);
+        let write = || self.write.load(Ordering::Acquire);
+
+        // If read-pointer is still 0, probably no host is connected
+        // TODO: What happens if flush is called before any logging call? Assumption: Probably we will just
+        // return, because nothing is written to the Channel, so both read and write pointer should be 0.
+        if read() == 0 {
+            // TODO: Do we want to return an error or return silently?
```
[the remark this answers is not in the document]
Urhengulas: `Logger::write` sais that "The write operation must not fail.", which isn't super clear to me why this is. I will extend it with your comment.

firmware/defmt-rtt/src/lib.rs
```diff
@@ -172,6 +184,38 @@ impl Channel {
...
         len
     }
+
+    fn flush(&self) {
+        // TODO: What is the best `Ordering`? The current ones are taken from `fn blocking_write`.
+        let read = || self.read.load(Ordering::Relaxed);
+        let write = || self.write.load(Ordering::Acquire);
+
+        // If read-pointer is still 0, probably no host is connected
+        // TODO: What happens if flush is called before any logging call? Assumption: Probably we will just
```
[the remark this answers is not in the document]
Urhengulas: I've tried to gather all possibilities:
```
case 1: flush is called before sending any data           -> write=0,     read=0  -> should return
case 2: no host connected                                 -> write>0,     read=0  -> should return
case 3: target flushes before host could read any data    -> write>0,     read=0  -> should wait "a bit"
case 4: target wrote, but host didn't read all data       -> write>read,  read>0 - > should busy wait until everything is read
case 5: impossible, host read more than what is available -> write<read,  read>=0 -> should return or panic (?)
```

As you've noticed, case 2 and 3 collide. I will think how to handle them best.

Are there other cases I didn't consider?

firmware/defmt-semihosting/src/lib.rs
```diff
@@ -37,6 +37,13 @@ unsafe impl defmt::Logger for Logger {
         unsafe { ENCODER.start_frame(do_write) }
     }
 
+    unsafe fn flush() {
+        // Do nothing.
+        //
+        // In QEMU write should auto-flush, because the target is emulated.
+        // With hardware write blocks for several hundred of milliseconds which should be enough to flush the data.
```
jonas-schievink, line 44: I think the real reason this can be a no-op is that semihosting fundamentally is blocking, and does not have I/O buffers the target can control (ie. no "flush" syscall, just "write"). After write returns, the host has the data, so there's nothing left to flush.

If the semihosting transport used an internal buffer things would be different, of course.

src/traits.rs
```diff
@@ -84,6 +84,15 @@ pub unsafe trait Logger {
     /// Panics if already acquired in the current execution context. Otherwise it must never fail.
     fn acquire();
 
+    /// Block until host has read all pending data.
+    ///
+    /// The flush operation must not fail. This is a "best effort", IO errors should be discarded.
```
jonas-schievink, line 89:
```suggestion
    /// The flush operation must not fail. This is a "best effort" operation, IO errors should be discarded.
```

firmware/defmt-rtt/src/lib.rs
```diff
@@ -47,6 +47,11 @@ unsafe impl defmt::Logger for Logger {
         unsafe { ENCODER.start_frame(do_write) }
     }
 
+    unsafe fn flush() {
+        // SAFETY: ...
```
jonas-schievink, line 51:
```suggestion
        // SAFETY: if we get here, the global logger mutex is currently acquired
```

firmware/defmt-rtt/src/lib.rs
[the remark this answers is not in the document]
jonas-schievink: Can't we detect whether the host is connected by checking whether the RTT stream is in non-blocking mode? It starts in non-blocking mode and when the host connects it sets it to blocking mode.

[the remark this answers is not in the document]
Urhengulas: Will it also go back to non-blocking if the host disconnects?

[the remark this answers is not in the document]
jonas-schievink: Not right now, but I don't think handling that case is too important

firmware/defmt-itm/src/lib.rs
```diff
@@ -64,6 +68,14 @@ unsafe impl defmt::Logger for Logger {
         unsafe { ENCODER.start_frame(do_write) }
     }
 
+    unsafe fn flush() {
+        // wait for the queue to be able to accept more data
+        while !stim_0().is_fifo_ready() {}
+
+        // delay "a bit" to drain the queue
+        asm::delay(100); // TODO: how long does it take to drain?
```
Urhengulas, line 76: How long does it take to drain?

firmware/defmt-rtt/src/lib.rs
```diff
@@ -206,3 +223,26 @@ unsafe fn handle() -> &'static Channel {
...
+
+/// Busy wait, until the read- catches up with the write-pointer.
+///
+/// Returns early, if the write-pointer stops progressing.
+fn busy_wait(read: impl Fn() -> usize, write: impl Fn() -> usize) {
+    let mut prev_read = 0;
+
+    while read() != write() {
+        // wait for some CPU cycles
+        asm::delay(100); // TODO: what's a useful number?
```
Urhengulas, line 239: How long should we wait in-between checks?

firmware/defmt-rtt/src/lib.rs
```diff
@@ -206,3 +223,26 @@ unsafe fn handle() -> &'static Channel {
...
+
+        // did the write-pointer progress?
+        // if it didn't the host is either disconnected or too slow
+        match read() == prev_read {
+            true => return,
+            false => prev_read = read(),
+        }
```
Urhengulas, line 246: Do we want to catch if the host disconnects while flushing? Else we could drop the whole while-body.

firmware/defmt-rtt/src/lib.rs
```diff
@@ -206,3 +223,26 @@ unsafe fn handle() -> &'static Channel {
...
+    }
```
Urhengulas, line 247: This loop could also be implemented without `asm::delay`, but a counter-variable:
```rust
    let mut counter = 0;
    while read() != write() {
        match counter % 100 == 0 {
            // did the write-pointer progress?
            true => match read() == prev_read {
                // if it didn't the host is either disconnected or too slow
                true => return,
                false => prev_read = read(),
            },
            false => counter += 1,
        }
    }
```

This would have the advantage that the condition (`read != write`) gets checked more often and we potentially detect the completion earlier.

firmware/defmt-itm/src/lib.rs
jonas-schievink: I think the TPIU can be clocked from an essentially arbitrary source, so there is no reliable way of determining this. Since the ITM does not seem to have any other facilities for flushing its output, it looks like we can't implement `flush` reliably for this transport.

I suppose that means that `flush` is, depending on the defmt transport, either a guaranteed operation, or just advisory.

firmware/defmt-rtt/src/lib.rs
```diff
@@ -8,6 +8,15 @@
 //! // src/main.rs or src/bin/my-app.rs
 //! use defmt_rtt as _;
 //! ```
+//!
+//! # Blocking/Non-blocking
+//!
+//! `probe-run` puts RTT into blocking-mode, to avoid loosing data.
```
jonas-schievink, line 14:
```suggestion
//! `probe-run` puts RTT into blocking-mode, to avoid losing data.
```

firmware/defmt-itm/src/lib.rs
jonas-schievink: Ah, I see that we've already documented `flush` as "best effort"

src/export.rs
```diff
@@ -47,6 +47,25 @@ pub fn acquire() {
...
+        fn _defmt_release();
+    }
+    // SAFETY: ...
+    unsafe {
+        _defmt_acquire();
+        _defmt_flush();
+        _defmt_release()
+    }
+}
+
```
jonas-schievink, line 68: Are these used by macros, or why are they in the `export` module?

Urhengulas: Do you only mean `flush` or all of `aquire`, `write`, `flush` and `release`?
And no, flush isn't used by default in some macro. I just placed them there to be consistent with the other logger-methods.
Do you know a better place?

firmware/defmt-itm/src/lib.rs
Urhengulas: If we can't determine the timing for itm, I would start with a relatively low one and add a note that people should open bug reports if it is too short in their case. If that happens we can see about raising it.

src/export.rs
jonas-schievink: `export` is not part of our public API, it only contains methods that have to be `pub` so macros can refer to them. If users are supposed to call `defmt::flush()`, then it needs to be placed in `src/lib.rs`.

Urhengulas: I see. Was wondering about the purpose of export module anyways. Thanks for clarifying 🙈

firmware/defmt-itm/src/lib.rs
Urhengulas: What do you think about the comment I added?

jonas-schievink: Seems good

src/export.rs
jonas-schievink: This was not yet resolved

src/lib.rs
```diff
@@ -351,15 +355,42 @@ pub use defmt_macros::bitflags;
...
+
+/// Block until host has read all pending data.
+///
+/// The flush operation will not fail, but might not succeed in flushing _all_ pending data. It is
+/// implemented as a "best effort" operation.
+///
+/// This calls the method `flush` of the used "global [`Logger`]". The logger is likely provided by
+/// [`defmt-rtt`](https://crates.io/crates/defmt-rtt) or [`defmt-itm`](https://crates.io/crates/defmt-itm).
+#[cfg(not(feature = "unstable-test"))]
+#[inline(never)]
```
Urhengulas, line 376: I just copied the `#[inline(never)]` from the existing function in the `export` module. Do I need this here also "because FFI"?

jonas-schievink: no, we don't really need that

src/export.rs
Urhengulas: Is it now?

src/lib.rs
Urhengulas: Perfect.

firmware/defmt-rtt/src/lib.rs
```diff
@@ -206,3 +227,8 @@ unsafe fn handle() -> &'static Channel {
 
     &_SEGGER_RTT.up_channel
 }
+
+fn host_is_connected(channel: &Channel) -> bool {
+    // we assume that a host is connected if we are in blocking-mode. this is what probe-run does.
+    channel.flags.load(Ordering::Relaxed) == BLOCK_IF_FULL
```
jonas-schievink, line 233: This breaks with https://github.com/knurling-rs/probe-run/pull/257 - it sets both `BLOCK_IF_FULL` and `NOBLOCK_TRIM`. Instead it should just check if the `BLOCK_IF_FULL` flag is set and ignore the others.

Urhengulas: Thank you for fixing it in #579 :)

PR #556, opened 2021-08-06
Add impl for TryFromIntError

newAM:
This implements `defmt::Format` for [`TryFromIntError`](https://doc.rust-lang.org/core/num/struct.TryFromIntError.html) which enables code like this:
```rs
use core::convert::{TryFrom, TryInto};

let foo: u64 = u64::MAX;
let bar: u32 = defmt::unwrap!(foo.try_into());
let eggs: u32 = defmt::unwrap!(u32::try_from(foo));
```

When it fails it looks like this:
```text
ERROR panicked at 'unwrap failed: foo.try_into()'
error: `TryFromIntError(())`
```

Which is similar to what it looks like on `std` targets with `.unwrap()`
```text
thread 'main' panicked at 'called `Result::unwrap()` on an `Err` value: TryFromIntError(())', src/main.rs:5:35
```


jonas-schievink:
Thanks!

bors r+

PR #561, opened 2021-08-18
Remove unused cortex-m-rt in panic-probe

Dirbaio:
It's nice to not depend on a particular cortex-m-rt version, it'll help the 0.6 -> 0.7 breaking transition.

This shouldn't be a breaking change. Could we get a release of this as 0.2.1 to help with the transition? Thank you!

### DIFF
--- a/firmware/panic-probe/Cargo.toml
+++ b/firmware/panic-probe/Cargo.toml
@@ -12,7 +12,6 @@ version = "0.2.0"
 
 [dependencies]
 cortex-m = "0.7.2"
-cortex-m-rt = "0.7.0"
 defmt = { version = "0.2.0", path = "../..", optional = true }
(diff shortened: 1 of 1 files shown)


Urhengulas:
> Could we get a release of this as 0.2.1 to help with the transition?

Although we said we want to avoid another `0.2.x` release, I think we can do this, if it is beneficial. And maybe it is obvious, but how does it help with the transition? 😁

Dirbaio:
> how does it help with the transition?

You can't have two versions of `cortex-m-rt` in the crate tree at the same time, they conflict because they need to set globally-unique things like the linker script and interrupt vectors. 

`panic-probe` pulling in `cortex-m-rt 0.6` completely prevents users from using `cortex-m-rt 0.7`

Issue #541, opened 2021-07-19
[RFC] `defmt::println!`


japaric:
Status update (2021-09-06): Approved. Needs implementation.

## Summary

add an equivalent of `std::println!` to `defmt`

## Rationale

`defmt` log macros emulate the `log` crate; log statements have log levels (ERROR, WARN, INFO, DEBUG, TRACE) and can be filtered out (Cargo feature (current mechanism) or DEFMT_LOG (proposed in PR #519)).

A `defmt::println!` statement would have no log level associated to it and it would not be possible to filter it out.

`defmt::println` could be used in `defmt-test` to ensure that the test suite progress messages are always printed regardless of the log filter configuration -- this would make the behavior of `defmt-test` closer to the "standard" `#[test]` harness.

## Alternatives

Not doing this means that if we approve PR #519 we need some to ensure that test suite progress messages are printed even when `DEFMT_LOG` is not set.
This could mean either:

- maintaining the existing Cargo feature based filter mechanism alongside the `DEFMT_LOG` filter mechanism (this would entail figuring out how the 2 should interoperate and extra maintenance work compared to removing the Cargo feature based filter)
- OR figuring out some other mechanism

japaric:
@jonas-schievink thoughts on this? do you think it would address the `defmt-test` issue that PR #519 introduces? any other issue that #519 creates that this doesn't address?

jonas-schievink:
I don't think I really have too strong of an opinion on this at the moment. A `println!` equivalent does sound useful just by itself, and if it lets us resolve the `defmt-test` issues then that's great, on the other hand I don't think this will end up being less work than supporting the Cargo feature based filters in addition to `DEFMT_LOG`.

japaric:
Maintenance cost aside (in my mind maintaining 2 features is more work than maintaining 1 (+)) I think that for end users is going to be confusing to have 2 ways to do the same thing: "which one do I pick? when do I use one over the other?"; those questions would need to be answered with documentation (*and* when need to specify the interaction between the two features).

If I had to pick one of the 2 filter mechanisms I would go with `DEFMT_LOG` because is closer to `RUST_LOG` which is the de facto standard in `std`; OTOH, having to add Cargo features to several crates to be able to filter is not common (it's also error-prone).

[...]

PR #569, opened 2021-09-06
Feature: Add defmt `println!` macro

justahero:
This PR addresses #541, adding a `defmt::println!` macro to always log content irrespective of a log level.

This changes the `level` property in `Frame` struct to be of `Option` type, which requires a number of refactorings to handle both ways to log, either using the log macros (e.g. `defmt::info!`) or using the `defmt::println!` macro.
The addition of the `defmt::println` macro should not break existing functionality.


parser/src/lib.rs
```diff
@@ -161,6 +161,7 @@ struct Param {
 /// The log level
 #[derive(Clone, Copy, Debug, PartialEq, PartialOrd)]
 pub enum Level {
+    None,
```
Urhengulas, line 164: We could also call it `Always` , which carries the intent that it will be always logged. This is just a suggestion :D

justahero: Atm this is better than `None`.

macros/src/function_like/log.rs
```diff
@@ -76,6 +76,7 @@ fn necessary_features_for_level(level: Level, debug_assertions: bool) -> &'stati
         Level::Trace if debug_assertions => &["defmt-trace", "defmt-default"],
         Level::Debug if debug_assertions => &["defmt-debug", "defmt-trace", "defmt-default"],
 
+        Level::None => &["defmt-default"],
```
justahero, line 79: Not sure if this is the right flag, decided not to introduce another flag here because it's kind of a default behaviour.

Urhengulas: Since `defmt::println!` should _always_ print, it should not require any feature. So you probably just need to return an empty slice (`&[]`).

parser/src/lib.rs
```diff
@@ -171,6 +172,7 @@ pub enum Level {
 impl Level {
     pub fn as_str(self) -> &'static str {
         match self {
+            Level::None => "(none)",
```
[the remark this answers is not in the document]
Urhengulas: Yes, we probably don't want to show a log-level for `println!`.

decoder/src/lib.rs
```diff
@@ -169,7 +171,7 @@ impl Table {
 
     fn get_with_level(&self, index: usize) -> Result<(Level, &str), ()> {
         let (lvl, format) = self._get(index)?;
-        Ok((lvl.ok_or(())?, format))
+        Ok((lvl.unwrap_or(Level::None), format))
```
jonas-schievink, line 174: This is still an error when it happens

decoder/src/frame.rs
```diff
@@ -412,14 +413,19 @@ impl fmt::Display for DisplayFrame<'_> {
...
                     self.frame
-                        .format_args(&fmt, &self.frame.timestamp_args, None,)
+                        .format_args(fmt, &self.frame.timestamp_args, None,),
                 )
             })
             .unwrap_or_default();
+
+        let level = level
+            .map(|level| format!("{} ", level))
+            .unwrap_or_else(String::new);
```
jonas-schievink, line 423:
```suggestion
            .unwrap_or_default();
```

macros/src/lib.rs
```diff
@@ -114,6 +114,12 @@ pub fn internp(args: TokenStream) -> TokenStream {
 }
 
 /* ## Logging macros */
+#[proc_macro]
+#[proc_macro_error]
+pub fn println(args: TokenStream) -> TokenStream {
+    function_like::println::expand(args)
```
jonas-schievink, line 120: Is there a reason you couldn't reuse `log::expand` like the other logging macros? (if so, please add a comment)

src/lib.rs
```diff
@@ -218,6 +218,13 @@ pub use defmt_macros::panic_handler;
 /// [`Str`]: struct.Str.html
 pub use defmt_macros::intern;
 
+/// Always logs data irrespective of log level
```
jonas-schievink, line 221:
```suggestion
/// Always logs data irrespective of log level.
```

decoder/src/log.rs
```diff
@@ -42,16 +43,28 @@ pub fn log_defmt(
...
+        );
+    } else {
+        log::logger().log(
+            &Record::builder()
+                .args(format_args!("{}", display))
+                .target(&target)
+                .module_path(module_path)
+                .file(file)
+                .line(line)
+                .build(),
```
jonas-schievink, line 65: Since you don't call `.level(...)` here, which level will end up in the `Record`?

jonas-schievink: Ah, looks like `info`: https://docs.rs/log/0.4.14/src/log/lib.rs.html#1157

jonas-schievink (commented):
Looks generally good, but I'm still not sure what our strategy with `log` should be. Perhaps the best way forward is to not funnel defmt logs through `log` and remove that code?

decoder/src/log.rs
justahero: The alternative is to be more transparent, for example:
```rust
        log::logger().log(
            &Record::builder()
                .args(format_args!("{}", display))
                .level(level.unwrap_or(Level::Trace))
                .target(&target)
                .module_path(module_path)
                .file(file)
                .line(line)
                .build(),
        );
```

and remove the `if else` block completely.

macros/src/lib.rs
justahero: I based it on function `write::expand` which is simlar, `println::expand` does not check feature flags and sets a different interned string. But I think both versions could be made work in `log::expand`.

jonas-schievink: Ah, that makes sense. Yeah, would be good to merge the logic together.

parser/src/lib.rs
justahero: The level variant was renamed to `Level::Println.`

macros/src/function_like/log.rs
justahero: I had to specify `defmt-default`, otherwise the `log` sample program would not output the `println` statement.

decoder/src/lib.rs
justahero: Reverted to previous call to `lvl.ok_or(())?`.

decoder/src/log.rs
justahero: It's better to be consistent with the underlying default level in `log`, therefore the call is now `.level(level.unwrap_or(Level::Info)`

justahero:
This feature still requires a bit of writing in the defmt book so that users know about how & when to use this.

decoder/src/log.rs
```diff
@@ -27,12 +27,13 @@ pub fn log_defmt(
...
-        crate::Level::Info => Level::Info,
-        crate::Level::Warn => Level::Warn,
-        crate::Level::Error => Level::Error,
+    let level: Option<Level> = match frame.level() {
+        crate::Level::Trace => Some(Level::Trace),
+        crate::Level::Debug => Some(Level::Debug),
+        crate::Level::Info => Some(Level::Info),
+        crate::Level::Warn => Some(Level::Warn),
+        crate::Level::Error => Some(Level::Error),
+        crate::Level::Println => None,
```
jonas-schievink, line 36: If we keep the fallback to `Info`, this branch could just be `=> Level::Info` and we'd remove the `Option`

jonas-schievink (approved):
Code LGTM

Still not sure what the best way to proceed with the `log` interop code is. I'm thinking that deleting it (and handling defmt frames manually) might be the best way forward, otherwise downstream consumers will have `println!` converted to `INFO`-level logs, which is kinda wrong.

japaric:
I have not looked at the PR but wanted to comment on these:

> otherwise downstream consumers will have println! converted to INFO-level logs, which is kinda wrong.

I agree with this. `defmt::println!` should not go through `log`'s machinery (including its filter mechanisms). Those messages should be directly printed to stdout.

> Still not sure what the best way to proceed with the log interop code is.

The `log` adapter is not part of stable decoder library API (there's no stable decoder API) so I'd be OK with removing it.

I think a "stable" `log` adapter would be useful in the future when there's more of a proper `defmt-decoder` library.
In my mind, that adapter should only expose `defmt`'s log statements and not `defmt::println`. The reason for that is that we kind of (already) have semihosting functionality in `defmt` so `defmt::println` should use the host's stdout, `defmt::eprintln` (does not exist) should use the host's stderr, `defmt::exit` (this is not in `defmt` proper) should exit the "host process", `defmt::info` should use the host's logging infrastructure (`log` crate), and so on and so forth.
(whether we should more clearly separate semihosting functionality from deferred formatting is a separate discussion)

decoder/src/log.rs
```diff
@@ -75,8 +75,7 @@ impl<'a> DefmtRecord<'a> {
         }
 
         let timestamp = &target[DEFMT_TARGET_MARKER.len()..];
-
-        Some(Self {
+        (Self{
```
japaric, line 78: does this compile? the return type is an option but this no longer has the `Some` constructor

firmware/defmt-test/macros/src/lib.rs
```diff
@@ -210,7 +210,7 @@ fn tests_impl(args: TokenStream, input: TokenStream) -> parse::Result<TokenStrea
                 test.func.sig.ident
             );
             quote_spanned! {
-                test.func.sig.ident.span() => defmt::info!(#message, __defmt_test_number, __DEFMT_TEST_COUNT);
+                test.func.sig.ident.span() => defmt::println!(#message, __defmt_test_number, __DEFMT_TEST_COUNT);
```
japaric, line 213: :+1:

macros/src/function_like/println.rs
```diff
@@ -0,0 +1,40 @@
...
+        &fragments,
+        formatting_exprs.len(),
+        args.format_string.span(),
+    );
+
+    let format_tag = construct::interned_string(&format_string, "println", false);
+    quote!({
+        match (#(&(#formatting_exprs)),*) {
+            (#(#patterns),*) => {
+                defmt::export::istr(&#format_tag);
```
japaric, line 34: this should follow the acquire - release pattern used in [the log macros](https://github.com/knurling-rs/defmt/blob/main/macros/src/function_like/log.rs#L43-L46), including the `header` call. the main difference between the two will be that tag used to constructor the header (there's a `make_istr` call in the expansion I believe)

japaric (commented):
Changes look good to me. It's not clear to me, though, what's the output when timestamps are enabled. Could you extend the `timestamp` snapshot test to include a `defmt::println!` call?

The other thing that's not obvious from here is whether `probe-run` will also show the source code location for `defmt::println!` statements. Including the location would not be standard but certainly useful. I think either way is fine for now personally; same thing for the timestamps.

decoder/src/log.rs
```diff
@@ -27,31 +27,38 @@ pub fn log_defmt(
...
-            .args(format_args!("{}", display))
-            .level(level)
-            .target(&target)
-            .module_path(module_path)
-            .file(file)
-            .line(line)
-            .build(),
-    );
+    if frame.level().is_some() {
+        let level = match frame.level().expect("log level expected") {
```
japaric, line 37: nit: you could avoid the `expect` here by doing an `if let`, instead of `if`, in the preceding line.

japaric (approved):
looks great! feel free to send it to bors with the inline nit addressed :rocket:

decoder/src/log.rs
justahero: Good catch.

PR #570, opened 2021-09-07
Support referring to `Self` in bitflags constants

jonas-schievink:
The initial bitflags support in https://github.com/knurling-rs/defmt/pull/528 had a restriction where referring to the `Self` type in a `bitflags` constant would result in a compile error.

Turns out that's entirely unnecessary and I wasn't quite awake when I wrote that code. A minor modification of the generated code avoids this somewhat annoying restriction. Instead of copying the expression verbatim, we can just refer to the associated constant created by the `bitflags` crate, and extract the raw bits.


Urhengulas (approved):
bors r+

PR #572, opened 2021-09-08
`defmt-decoder`: `impl TryFrom<Option<String>> for Encoding`

Urhengulas:


decoder/src/lib.rs
```diff
@@ -121,6 +122,21 @@ enum Encoding {
     Rzcobs,
 }
 
+impl TryFrom<Option<String>> for Encoding {
```
jonas-schievink, line 125: I'd rather implement `FromStr` instead and `.unwrap()` the `Option<String>` (since the code already says that that condition would be a bug)

Urhengulas: That's what I started with, but then though that we can actually wrap the whole conversion process into one function call.
What advantage do you see with `FromStr`?

jonas-schievink: `TryFrom<Option<String>>` makes it fairly unclear what the input means, and `FromStr` is generally preferred over `TryFrom<String>` or `TryFrom<&str>` for parsing a string into something else, at least from what I've seen (it also works with `str::parse()`)

Urhengulas: Kk

PR #574, opened 2021-09-08
Refactor rtt [1/2]

Urhengulas:
First half of #573:
*  defmt-rtt: Move struct Channel into own module
* defmt-rtt: Make fn host_is_connected a method of Channel
* defmt-rtt: Move struct Header into own module
* defmt-rtt: Rename fn handle to fn channel + mv to mod channel

### DIFF
--- a/firmware/defmt-rtt/src/channel.rs
+++ b/firmware/defmt-rtt/src/channel.rs
@@ -22,7 +22,7 @@ impl Channel {
     pub fn write_all(&self, mut bytes: &[u8]) {
         // the host-connection-status is only modified after RAM initialization while the device is
         // halted, so we only need to check it once before the write-loop
-        let write = match host_is_connected(self) {
+        let write = match self.host_is_connected() {
             true => Channel::blocking_write,
             false => Channel::nonblocking_write,
         };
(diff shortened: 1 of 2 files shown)


jonas-schievink:
Looks good! Although I think I'd prefer to drop 6874e9b, since that moves a lot more than just the `Header` struct, and the `_SEGGER_RTT` static is something that IMO doesn't belong buried in a `header` module. (this also needs a rebase after the recent RTT fixes, but other than that LGTM)

Urhengulas:
> Looks good! Although I think I'd prefer to drop [6874e9b](https://github.com/knurling-rs/defmt/commit/6874e9b9229b2e9fc6a9e3549fda8f0da083dd2f), since that moves a lot more than just the `Header` struct, and the `_SEGGER_RTT` static is something that IMO doesn't belong buried in a `header` module. (this also needs a rebase after the recent RTT fixes, but other than that LGTM)

So - I would keep the `const`s and `static`s in the `lib.rs` and only move out the `struct Header`.

Urhengulas:
@jonas-schievink Can you please take another look?

jonas-schievink:
What do we gain from moving `Header` to another file? Since it only has 4 field and no methods, it doesn't use much code

Urhengulas:
> What do we gain from moving `Header` to another file? Since it only has 4 field and no methods, it doesn't use much code

Actually true

Urhengulas:
That's it?

PR #577, opened 2021-09-14
Fix typo in `cfg` of encoding-feature-`compile_error!`

### LINKED ISSUE #576: it's possible to enable both decoding-raw and decoding-rzcobs
this should be a compile error because it becomes ambiguous which encoding is being used.
the current behavior seems to be that `encoding-raw` "wins" if both Cargo features are enabled

Urhengulas:
Fixes #576


Urhengulas:
bors retry

PR #579, opened 2021-09-15
defmt-rtt: fix check for blocking RTT

jonas-schievink:
This check wasn't testing correctly whether the `BLOCK_IF_FULL` bit was set, leading to https://github.com/knurling-rs/probe-run/pull/257 breaking probe-run.


japaric:
Thanks!

bors r+

PR #580, opened 2021-09-15
Structure `defmt::export`

Urhengulas:


src/export/integers.rs
```diff
@@ -0,0 +1,36 @@
...
+        /// Implementation detail
+        pub fn $s(b: &$s) {
+            write(&b.to_le_bytes())
+        }
+    };
+}
+
+/// Implementation detail
+pub fn u8(b: &u8) {
+    write(&[*b])
```
jonas-schievink, line 14: `u8` also has `to_le_bytes`

src/export/integers.rs
```diff
@@ -0,0 +1,36 @@
...
+}
+
+write_to_le_bytes!(u16);
+write_to_le_bytes!(u32);
+write_to_le_bytes!(u64);
+write_to_le_bytes!(u128);
+
+/// Implementation detail
+pub fn usize(b: &usize) {
+    write(&(*b as u32).to_le_bytes())
```
jonas-schievink, line 24: `usize` and `isize` also have `to_le_bytes`

src/export/integers.rs
```diff
@@ -0,0 +1,36 @@
...
+}
+
+write_to_le_bytes!(i8);
+write_to_le_bytes!(i16);
+write_to_le_bytes!(i32);
+write_to_le_bytes!(i64);
+write_to_le_bytes!(i128);
```
jonas-schievink, line 31: This would be simpler if the macro took a list of types, then you only have to call it once

src/export/sealed.rs
```diff
@@ -0,0 +1,82 @@
...
+impl_truncate!(u32, u32);
+impl_truncate!(u32, u64);
+impl_truncate!(u32, u128);
+
+// needed so we can call truncate() without having to check whether truncation is necessary first
+impl_truncate!(u64, u64);
+impl_truncate!(u64, u128);
+
+// needed so we can call truncate() without having to check whether truncation is necessary first
+impl_truncate!(u128, u128);
```
jonas-schievink, line 41: The comment can probably be deduplicated and rewritten to be a bit clearer with the new macros, maybe something like

> We implement `Truncate<X> for X` so that the macro can unconditionally use it, even if no truncation is performed

jonas-schievink: In addition, this macro could also take a list, and an operator indicating the direction of the truncation would make it a bit clearer what's going on:

```
impl_truncate!(u8 -> u8, u16 -> u8, ...);
```

src/export/sealed.rs
```diff
@@ -0,0 +1,82 @@
+#[allow(unused_imports)]
```
jonas-schievink, line 1: Why is this file called `sealed.rs`?

src/export/sealed.rs
```diff
@@ -0,0 +1,82 @@
...
+    }
+
+    fn _format_tag() -> Str {
+        defmt_macros::internp!("Unwrap of a None option value")
+    }
+
+    fn _format_data(&self) {}
+}
+
+pub trait IntoResult {
```
jonas-schievink, line 58: It would be good to document/link what uses this API (since `export` is meant for use by macros, rust-analyzer's "Find References"  doesn't work).

For this trait, that would be the `defmt::unwrap!` macro.

src/export/sealed.rs
Urhengulas: Because the module was called like this before and I just kept the name: https://github.com/knurling-rs/defmt/blob/e021e7d013b9202f70bc762f6909401f85142c89/src/export.rs#L101

What other name do you recommend?

src/export/integers.rs
Urhengulas: but don't we need the `as u32`, `as i32`?

Urhengulas: It seems like we do, since the host tests fail if we remove it.

src/export/sealed.rs
jonas-schievink: Ah, that's weird. Perhaps `traits` would be better then?

src/export/sealed.rs
```diff
@@ -0,0 +1,82 @@
...
+    u16  => u16,
+    u16  => u32,
+    u16  => u64,
+    u16  => u128,
+    u32  => u32,
+    u32  => u64,
+    u32  => u128,
+    u64  => u64,
+    u64  => u128,
+    u128 => u128
```
jonas-schievink, line 36: These point the wrong way

Urhengulas: Is it better like this?

PR #581, opened 2021-09-16
Add impl for alloc::Layout

newAM:
This is useful for displaying the layout argument in the `alloc_error_handler`, e.g.

```rs
#[alloc_error_handler]
fn oom(layout: Layout) -> ! {
    defmt::error!("OOM {}", layout):

    loop {
        cortex_m::asm::bkpt();
    }
}
```


Urhengulas:
Thank you @newAM!

bors r+

PR #584, opened 2021-09-17
Remove outdated doc "you may only call write! once"

### LINKED ISSUE #583: Outdated doc-comment for `trait Format` (?)
It seems that the doc-comment for the `trait Format` is outdated. Definitely the "Note that [`write!`] can only be called once, as it consumes the [`Formatter`].". Maybe also the section about the manual implementation?

https://github.com/knurling-rs/defmt/blob/149c0efe48d60b75d672089e0c9612f826304b5e/src/traits.rs#L7-L46

@Dirbaio Can you please update this? 😁

Dirbaio:
SInce #508 it is possible to call it multiple times, and the result is concatenated as you'd expect.

The rest of the comment is still relevant.

Fixes #583

### DIFF
--- a/src/traits.rs
+++ b/src/traits.rs
@@ -42,8 +42,6 @@ use crate::{export, Formatter, Str};
 ///     }
 /// }
 /// ```
-///
-/// Note that [`write!`] can only be called once, as it consumes the [`Formatter`].
 pub trait Format {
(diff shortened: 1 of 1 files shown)


Urhengulas:
Thank you @Dirbaio! I wasn't sure if the rest also needed an update and therefore didn't just remove this one line myself :)

PR #585, opened 2021-09-23
Add xtask option to run a single snapshot test by name

justahero:
This is to only run a single snapshot test when checking things without running all tests again.

```shell
cargo xtask test-snapshot log
```

will only run the `log` snapshot test. In case no test name is given, all tests are run as before. The test name of the positional parameter (`<single>`) is checked against the allowed list of snapshot tests.

### DIFF
--- a/xtask/src/main.rs
+++ b/xtask/src/main.rs
@@ -30,6 +30,30 @@ const ALL_SNAPSHOT_TESTS: [&str; 12] = [
     "dbg",
 ];
 
+#[derive(Debug)]
+struct Snapshot(String);
+
+impl Snapshot {
+    pub fn name(&self) -> &str {
+        &self.0
+    }
+}
+
+impl FromStr for Snapshot {
+    type Err = String;
+
+    fn from_str(test: &str) -> Result<Self, Self::Err> {
+        if ALL_SNAPSHOT_TESTS.contains(&test) {
+            Ok(Self(String::from(test)))
+        } else {
+            Err(format!(
+                "Specified test '{}' does not exist, available tests are: {:?}",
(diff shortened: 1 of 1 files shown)


xtask/src/main.rs
```diff
@@ -40,6 +84,9 @@ enum TestCommand {
         /// Overwrite the expected output instead of comparing it.
         #[structopt(long)]
         overwrite: bool,
+        /// Runs a single snapshot test in Debug mode
+        #[structopt(long, default_value = "all")]
+        single: Snapshot,
```
jonas-schievink, line 89: Being able to pass `--single all` is a bit odd, maybe this would be better as a positional argument? Then you could do `cargo xtask test-snapshot` or `cargo xtask test-snapshot all` to run all of them, or `cargo xtask test-snapshot bitflags` to run a single one

xtask/src/main.rs
```diff
@@ -15,6 +15,50 @@ use crate::utils::{
...
+}
+
+impl FromStr for Snapshot {
+    type Err = String;
+
+    fn from_str(source: &str) -> Result<Self, Self::Err> {
+        match source {
+            "all" => Ok(Snapshot::All),
+            test => {
+                if ALL_SNAPSHOT_TESTS.contains(&test) {
```
jonas-schievink, line 47: This doesn't work with the conditionally-available `alloc` test, but that's probably not a big problem

jonas-schievink (approved):
Implementation looks good, feel free to send to bors

xtask/src/main.rs
justahero: Refactored the `single` option into a positional parameter.

PR #591, opened 2021-09-24
Remove timestamps from snapshot test

jonas-schievink:
This keeps the diff much cleaner when a line is added. Previously the entire rest of the file changed because of the increasing timestamps.

The timestamp functionality is still tested in `timestamp.rs`.


japaric (approved):
awesome!

bors r+

PR #592, opened 2021-09-27
xtask: add backward compability test

### LINKED ISSUE #506: Backwards-compatability tests
**Status update (2021-09-10)** needs implementation. see notes in https://github.com/knurling-rs/defmt/issues/506#issuecomment-916869244

#503 broke the backwards-compatability tests and since we don't guarantee backwards-compatability we kinda disabled them by just running with the current decoder. 

My question is now if we should just remove these tests all together since we don't guarantee it anyways, or if we should pin it to the current version of the decoder?
I am fine with both with a slight preference for removing.

japaric:
**UPDATE**: I have left the defmt-version as it is since we have not made a defmt 0.3.0 release. I have also tweaked the defmt-test to not exit with non-zero exit code and remove some test-specific logic from xtask. See individual commits messages for details.

---

implements https://github.com/knurling-rs/defmt/issues/506#issuecomment-916869244
closes #506

I tested this against c7b20d53861d0c4fa476db4791d5745412f4de57 (I called that `defmt-version-3` locally) and found that a bitflags change broke the wire format. Not a big deal but I guess we should bump the version to 4? cc @jonas-schievink 

``` console
bitflags (dev)
Error: malformed bitflags value string 'Flags::0::FLAG_0'
```

the other thing that I found out is that some of the snapshot test exit with non-zero code:

``` console
defmt-test (dev)
INFO (1/7) running `change_init_struct`...
INFO (2/7) running `test_for_changed_init_struct`...
INFO (3/7) running `assert_true`...
INFO (4/7) running `assert_imported_max`...
INFO (5/7) running `result`...
INFO (6/7) running `should_error`...
INFO (7/7) running `fail`...
ERROR panicked at '`#[should_error]` test failed with outcome: Ok(this should have returned `Err`)'

Timer with period zero, disabling
```

`cargo xtask test-snapshot` is not checking the exit code; it only looks at stdout so that non-zero exit code is not a problem for those tests. in the backcompat test we want to look for decoding errors, not at stdout, so I was using the exit code of `cargo run` but this `defmt-test` is problematic with that check.


japaric:
> found that a bitflags change broke the wire format.

bisected this down to PR #564

xtask/src/backcompat.rs
```diff
@@ -0,0 +1,175 @@
...
+    }
+
+    fn path(&self) -> &Path {
+        &self.executable_path
+    }
+}
+
+fn formatted_test_name(name: &str, release_mode: bool) -> String {
+    format!("{} ({})", name, if release_mode { "release" } else { "dev" })
+}
```
Urhengulas, line 93: Maybe also use this at https://github.com/knurling-rs/defmt/blob/3cda211d192141f2ef2956c9adcdc9f091d2d37c/xtask/src/main.rs#L336?

xtask/src/backcompat.rs
```diff
@@ -0,0 +1,175 @@
...
+
+fn clone_repo(tempdir: &Path) -> anyhow::Result<()> {
+    let repo_path = Path::new(env!("CARGO_MANIFEST_DIR")).parent().unwrap();
+    run_silently(
+        Command::new("git")
+            .arg("clone")
+            .arg(repo_path)
+            .arg(".")
+            .current_dir(tempdir),
```
Urhengulas, line 102: Why do we need to clone the `defmt` repo again? Isn't it already checked out by `actions/checkout@v2`?

Urhengulas (approved):
Looks good. I've left two comments. You add a bit of redundant code, but that's not a huge problem.

xtask/src/backcompat.rs
japaric: we don't "need" to: we could certainly rollback the `actions/checkout@v2` repo; do the cargo install; and then forward it again to the original checkout.

it felt cleaner to isolate the build of the old qemu-run; the other ("reuse") option may invalidate the existing build artifacts in `target` and cause further recompilations when building the snapshot (`firmware/qemu`) programs (which I guess would mainly be an issue with a local `xtask test-all` and not with this particular CI job)

Urhengulas: Makes sense

japaric:
had to bump the revision because defmt::println! broke the wire format (expected)
bors r=Urhengulas

PR #598, opened 2021-10-01
`defmt-print`: Recover from decoding-errors

### LINKED ISSUE #559: defmt-print cannot recover from decoding errors
It look like `defmt-print` exits with an error when decoding fails, even if rzCOBS is used for the transport encoding. Since a framed transport allows recovering from such errors, we should make sure that `defmt-print` can do that.

Urhengulas:
This PR uses the new `StreamDecoder`-API in `defmt-print` and utilizes it to recover from decoding errors.

Fixes #559

### DIFF
--- a/decoder/src/lib.rs
+++ b/decoder/src/lib.rs
@@ -119,7 +119,7 @@ struct BitflagsKey {
 }
 
 #[derive(Copy, Clone, Debug, PartialEq)]
-enum Encoding {
+pub enum Encoding {
     Raw,
     Rzcobs,
 }
@@ -252,6 +252,10 @@ impl Table {
             Encoding::Raw => Box::new(stream::Raw::new(self)),
             Encoding::Rzcobs => Box::new(stream::Rzcobs::new(self)),
         }
+    }
+
+    pub fn encoding(&self) -> Encoding {
+        self.encoding
     }
 }
 
--- a/decoder/src/lib.rs
+++ b/decoder/src/lib.rs
@@ -205,14 +205,15 @@ impl Table {
         elf2table::get_locations(elf, self)
     }
 
-    /// decode the data sent by the device using the previosuly stored metadata
+    /// Decode the data sent by the device using the previously stored metadata.
     ///
-    /// * bytes: contains the data sent by the device that logs.
-    ///          contains the [log string index, timestamp, optional fmt string args]
+    /// * `bytes`
+    ///   * contains the data sent by the device that logs.
+    ///   * contains the [log string index, timestamp, optional fmt string args]
     pub fn decode<'t>(
         &'t self,
         mut bytes: &[u8],
-    ) -> Result<(Frame<'t>, /*consumed: */ usize), DecodeError> {
+    ) -> Result<(Frame<'t>, /* consumed: */ usize), DecodeError> {
         let len = bytes.len();
         let index = bytes.read_u16::<LE>()? as u64;
 
--- a/decoder/src/lib.rs
+++ b/decoder/src/lib.rs
@@ -138,7 +138,7 @@ impl FromStr for Encoding {
 }
 
 impl Encoding {
-    pub const fn recoverable(&self) -> bool {
+    pub const fn can_recover(&self) -> bool {
         match self {
             Encoding::Raw => false,
             Encoding::Rzcobs => true,
--- a/decoder/src/lib.rs
+++ b/decoder/src/lib.rs
@@ -119,6 +119,7 @@ struct BitflagsKey {
 }
 
 #[derive(Copy, Clone, Debug, PartialEq)]
+#[non_exhaustive]
 pub enum Encoding {
     Raw,
     Rzcobs,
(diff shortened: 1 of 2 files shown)


print/src/main.rs
```diff
@@ -51,58 +51,62 @@ fn main() -> anyhow::Result<()> {
...
-                }
+            match stream_decoder.decode() {
+                Ok(frame) => forward_to_logger(&frame, location_info(&locs, &frame, &current_dir)),
+                Err(DecodeError::UnexpectedEof) => break,
+                Err(DecodeError::Malformed) => match table.encoding() {
+                    // raw encoding doesn't allow for recovery. therefore we abort.
+                    Encoding::Raw => return Err(DecodeError::Malformed.into()),
+                    // rzcobs encoding allows recovery from decoding-errors. therefore we stop
+                    // decoding the current, corrupted, data and continue with new one
+                    Encoding::Rzcobs => break,
```
japaric, line 75: I think this should be `continue` because may be more data cached in the `stream_decoder`. if there isn't then the next iteration will hit the `Eof` case (which will make the program fetch more data from stdin)

print/src/main.rs
```diff
@@ -51,58 +51,62 @@ fn main() -> anyhow::Result<()> {
...
-                }
-                Err(defmt_decoder::DecodeError::UnexpectedEof) => break,
-                Err(defmt_decoder::DecodeError::Malformed) => {
-                    log::error!("failed to decode defmt data: {:x?}", frames);
-                    return Err(defmt_decoder::DecodeError::Malformed.into());
-                }
+            match stream_decoder.decode() {
+                Ok(frame) => forward_to_logger(&frame, location_info(&locs, &frame, &current_dir)),
+                Err(DecodeError::UnexpectedEof) => break,
+                Err(DecodeError::Malformed) => match table.encoding() {
```
japaric, line 70: could you move the 'can this encoding recover?' logic into a method on the `Encoding` enum itself? I think consumers should not be deciding on the `Encoding` variant but rather the `defmt-decoder` library should provide them that information.

decoder/src/lib.rs
```diff
@@ -119,7 +119,7 @@ struct BitflagsKey {
 }
 
 #[derive(Copy, Clone, Debug, PartialEq)]
-enum Encoding {
+pub enum Encoding {
```
japaric, line 122: although all of `defmt-decoder` is unstable making this enum public is a semver hazard because now adding a variant is a breaking change in the API and that requires a semver bump. could you add a `#[non_exhaustive]` attribute here to prevent that potential issue?

(we could even go further and prevent matching on the enum by hiding it inside a newtype `pub struct Encoding(EncodingKind); enum EncodingKind { .. }` but that may be overkill)

japaric (commented):
changes look good. left some suggestions

decoder/src/lib.rs
Urhengulas: > could you add a `#[non_exhaustive]` attribute here to prevent that potential issue?

Done!

print/src/main.rs
Urhengulas: 👍🏾

print/src/main.rs
Urhengulas: 👍🏾

decoder/src/lib.rs
```diff
@@ -136,6 +137,15 @@ impl FromStr for Encoding {
     }
 }
 
+impl Encoding {
+    pub const fn recoverable(&self) -> bool {
```
japaric, line 141: nit: naming convention (style?) is to name this `is_recoverable` (or maybe `can_recover`) when the method returns a `bool`

print/src/main.rs
```diff
@@ -51,58 +51,61 @@ fn main() -> anyhow::Result<()> {
...
-                }
-                Err(defmt_decoder::DecodeError::UnexpectedEof) => break,
-                Err(defmt_decoder::DecodeError::Malformed) => {
-                    log::error!("failed to decode defmt data: {:x?}", frames);
-                    return Err(defmt_decoder::DecodeError::Malformed.into());
-                }
+            match stream_decoder.decode() {
+                Ok(frame) => forward_to_logger(&frame, location_info(&locs, &frame, &current_dir)),
+                Err(DecodeError::UnexpectedEof) => break,
+                Err(DecodeError::Malformed) => match table.encoding().recoverable() {
```
japaric, line 70: (probably down to personal preference but `if boolean` is more common than `match boolean` IMO)

japaric (approved):
:+1:  Thanks! Feel free to send to bors with nits addressed or unaddressed

decoder/src/lib.rs
Urhengulas:
```suggestion
    pub const fn can_recover(&self) -> bool {
```

print/src/main.rs
Urhengulas, line 70:
```suggestion
                Err(DecodeError::Malformed) => match table.encoding().can_recover() {
```

Urhengulas: I think it is more common to use if. But match saves one line 😋

Urhengulas:
bors retry

PR #601, opened 2021-10-05
Move crate `defmt` to `defmt/`

Urhengulas:
I'd like `cargo watch` or `cargo check` in the root directory to check all the host crates, but before this change it will only check the `defmt` crate.


Urhengulas:
Just noticed that I need to adapt the crates in `firmware/`

PR #604, opened 2021-10-06
defmt-test: `#[cfg(test)]` the `#[defmt_test::tests]` module

japaric:
[it recently came to my attention](https://github.com/knurling-rs/app-template/commit/1ebc0ac43f18e6919dd98c1cb49934f24c9c1a81) that one can use `defmt-test` for "unit testing" by using it in `src/lib.rs` (although this is pretty limited at the moment because one can only use the `defmt_test::tests` *once* in a library crate)

if one does use `defmt_test` like that then they may run into the issue that running the test files in `tests` fails with this cryptic linker message: 

``` console
  = note: rust-lld: error: duplicate symbol: main
```

the issue is that the code generated by `defmt_test::tests` (which includes a function with unmangled name `main`) is not conditionally compiled but we do want that to be `#[cfg(test)]`-ed because it contains test code.

r? @jonas-schievink


japaric:
argh, our defmt-test snapshot test is unusual because is a `bin/*.rs` file and not a proper `test/*.rs` file which does get compiled with `--cfg test` when you do `cargo test`. this will require `xtask` changes to land. turning into a draft PR for now

PR #606, opened 2021-10-07
Update change log with recent entries

justahero:


justahero:
This project is the only one that uses the bors author & reviewer annotation, therefore I added them in for consistency.

Issue #390, opened 2021-02-15
defmt-test: #[ignore] attribute support


japaric:
Support this

``` rust
// inside the `#[defmt_test::test]` module
#[ignore] 
#[test]
fn foo() {}
```

Tests marked as `ignore`-d should be compiled but not executed.
Ignored tests should be reported on the console output but should not be account for in the progress counter (e.g. `1/3`).

Urhengulas:
* Clarification question: The test should get compiled, but never executed. What will then get reported in the console?
* Also is compatibility with the `--include-ignored` flag desired? ([link](https://doc.rust-lang.org/reference/attributes/testing.html#the-ignore-attribute))

japaric:
> Clarification question: The test should get compiled, but never executed. What will then get reported in the console?

Ideally, something similar to what the built-in test harness reports: 
``` console
test tests::it_works ... ignored
```

> Also is compatibility with the --include-ignored flag desired?

we don't support "test" related flags in probe-run. That's blocked on [this RFC](https://github.com/knurling-rs/probe-run/issues/148)

japaric:
implemented in #618

PR #618, opened 2021-10-21
Support #[ignore] attribute in defmt_test

justahero:
This PR adds support for the `#[ignore]` attribute in `defmt_test` to solve #390. Similar to the [`ignore`](https://doc.rust-lang.org/reference/attributes/testing.html#the-ignore-attribute) attribute, tests can be annotated with the attribute to ignore particular tests.

* use [`trybuild`](https://github.com/dtolnay/trybuild) crate to check multiple conditions with the proc macro

**Note** the message is different from the suggestion in #390 and should be revised accordingly if needed


firmware/defmt-test/tests/ui/test-uses-invalid-type-annotation.rs
```diff
@@ -0,0 +1,10 @@
+fn main() {}
+
+#[defmt_test::tests]
```
japaric, line 3: the accompanying .stderr file is empty so this test is failing.
but also CI seems to not be running the UI tests because it's currently green. could you update the xtask to run the new UI tests in defmt-test?

japaric (changes requested):
spotted 2 issues

firmware/defmt-test/tests/ui/test-uses-invalid-type-annotation.rs
justahero: Running ui tests with `trybuild` on Rust **stable** and **nightly** has a few issues on its own, therefore the snapshot tests are only run in a new `xtask` command on **stable** to be consistent.

justahero: Also removed this test for this PR, because `trybuild` is not passing `--cfg test` to `rustc` when running tests, therefore skipping code checks in the resulting `#[cfg(test)]` block. Therefore this test file will succeed, even though the resulting code would fail when for example used as a test in the app template testsuite.

japaric (approved):
Thanks for working on this!

firmware/defmt-test/macros/src/lib.rs
```diff
@@ -181,9 +194,13 @@ fn tests_impl(args: TokenStream, input: TokenStream) -> parse::Result<TokenStrea
         } else {
             quote!(#ident())
         };
-        unit_test_calls.push(quote!(
-            #krate::export::check_outcome(#call, #should_error);
-        ));
+        if ignore {
+            unit_test_calls.push(quote!(let _ = #call;));
```
jonas-schievink, line 198: This will still call the test function if it has `#[ignore]` on it, leading to the issue reported in https://github.com/knurling-rs/defmt/issues/731

PR #622, opened 2021-10-31
Refactor rtt [2/2]

Urhengulas:
This PR contains the not-error-prone changes from #573 and addresses the review given to #573.


Urhengulas:
@japaric Can we actually merge this one? As stated in the description this PR merely adds a bit of documentation and outsources the calculation of the available buffer size to a utility function.

PR #626, opened 2021-11-12
Fix error message  in case of version-mismatch

bobmcwhirter:
Make the error message more meaningful. 

Currently it tells you probe-run supports defmt X while the firmware supports defmt Y, which is reasonable. But the substitution then tells you to install a version of probe-run that supports X (the version you're already using), which is contradictory, since Y (firmware supported version) is the real version you need to install for.


bobmcwhirter:
Exact error output prior to this change:

```
Error: defmt wire format version mismatch: firmware is using 0.2, `probe-run` supports 3
suggestion: `cargo install` a different version of `probe-run` that supports defmt 3
```

Urhengulas:
Good catch!

bors r+

Urhengulas:
bors cancel

PR #633, opened 2021-11-25
fix #628

### LINKED ISSUE #628: make RTT buffer size configurable
currently it's fixed to [1 KiB](https://github.com/knurling-rs/defmt/blob/89843557a40163ca6e1a3fbed5e890223e14628a/firmware/defmt-rtt/src/lib.rs#L97-L99).

coming up with a configuration mechanism is part of this issue. One option is use an env variable, e.g. DEFMT_RTT_BUFFER_SIZE.

as a size power of 2 improves performance it would be good to inform that to user where configuring the size of this buffer

spookyvision:
open questions:

also make this honor the env var?
```rust
// print/src/main.rs:28
const READ_BUFFER_SIZE: usize = 1024;
```
(answered: no, irrelevant)

---

add some hints in the app template?
(answered: yes)

### DIFF
--- a/firmware/defmt-rtt/build.rs
+++ b/firmware/defmt-rtt/build.rs
@@ -1,18 +1,16 @@
+use std::{env, path::PathBuf};
+
 fn main() {
     println!("cargo:rerun-if-env-changed=DEFMT_RTT_BUFFER_SIZE");
 
-    let size = option_env!("DEFMT_RTT_BUFFER_SIZE")
-        .map(|s| s.parse().ok())
-        .flatten()
+    let size = env::var("DEFMT_RTT_BUFFER_SIZE")
+        .map(|s| {
+            s.parse()
+                .expect("coult not parse DEFMT_RTT_BUFFER_SIZE as usize")
+        })
         .unwrap_or(1024_usize);
 
-    let non_power_of_two = (size & (size - 1)) != 0;
-
-    if non_power_of_two {
-        println!("cargo:warning=RTT buffer size of {} is not a power of two, performance will be degraded.", size)
-    }
-
-    let out_dir_path = std::path::PathBuf::from(std::env::var_os("OUT_DIR").unwrap());
+    let out_dir_path = PathBuf::from(env::var_os("OUT_DIR").unwrap());
     let out_file_path = out_dir_path.join("consts.rs");
 
     std::fs::write(
--- a/firmware/defmt-rtt/build.rs
+++ b/firmware/defmt-rtt/build.rs
@@ -1,0 +1,23 @@
+fn main() {
+    println!("cargo:rerun-if-env-changed=DEFMT_RTT_BUFFER_SIZE");
+
+    let size = option_env!("DEFMT_RTT_BUFFER_SIZE")
+        .map(|s| s.parse().ok())
+        .flatten()
+        .unwrap_or(1024_usize);
+
+    let non_power_of_two = (size & (size - 1)) != 0;
+
+    if non_power_of_two {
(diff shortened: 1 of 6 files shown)


firmware/defmt-rtt/src/lib.rs
```diff
@@ -94,9 +94,19 @@ const MODE_BLOCK_IF_FULL: usize = 2;
 /// Don't block if the RTT buffer is full. Truncate data to output as much as fits.
 const MODE_NON_BLOCKING_TRIM: usize = 1;
 
-// TODO make configurable
-// NOTE use a power of 2 for best performance
-const SIZE: usize = 1024;
+// buffer size. Default: 1024; can be customized by setting `DEFMT_RTT_BUFFER_SIZE`
```
jonathanpallant, line 97: Should these be `///`?

firmware/defmt-rtt/build.rs
```diff
@@ -0,0 +1,23 @@
...
+    let size = option_env!("DEFMT_RTT_BUFFER_SIZE")
+        .map(|s| s.parse().ok())
+        .flatten()
+        .unwrap_or(1024_usize);
+
+    let non_power_of_two = (size & (size - 1)) != 0;
+
+    if non_power_of_two {
+        println!("cargo:warning=RTT buffer size of {} is not a power of two, performance will be degraded.", size)
+    }
```
jonas-schievink, line 13: Since this won't show up for most users, and has no way to silence it if it does, maybe remove it entirely?

firmware/defmt-rtt/build.rs
```diff
@@ -0,0 +1,23 @@
+fn main() {
+    println!("cargo:rerun-if-env-changed=DEFMT_RTT_BUFFER_SIZE");
+
+    let size = option_env!("DEFMT_RTT_BUFFER_SIZE")
+        .map(|s| s.parse().ok())
```
jonas-schievink, line 5: We should panic if this does not parse as a usize here

firmware/defmt-rtt/build.rs
spookyvision: I'm unfortunately not following, can you describe a concrete problem case?

jonathanpallant: As an aside, `size.count_ones() == 1` should do the same thing. But yeah, warnings don't help here. The other option is having two env vars, one for `RTT_BUFFER_SIZE` and one for `RTT_BUFFER_SIZE_NON_ALIGNED` and panic if the first isn't 2**N.

spookyvision: I was under the impression that a user would configure this variable to conserve memory (at least, that's my personal use case and why I opened this issue/can of worms in the first place); what am I missing re: warnings don't help?

jonas-schievink: Warnings don't show up for users unless they use defmt-rtt as a path dependency

firmware/defmt-rtt/build.rs
```diff
@@ -0,0 +1,21 @@
+use std::{env, path::PathBuf};
+
+fn main() {
+    println!("cargo:rerun-if-env-changed=DEFMT_RTT_BUFFER_SIZE");
+
+    let size = env::var("DEFMT_RTT_BUFFER_SIZE")
+        .map(|s| {
+            s.parse()
+                .expect("coult not parse DEFMT_RTT_BUFFER_SIZE as usize")
```
jonas-schievink, line 9:
```suggestion
                .expect("could not parse DEFMT_RTT_BUFFER_SIZE as usize")
```

PR #635, opened 2021-11-25
mention #628

spookyvision:


jonathanpallant:
Marking your own homework @spookyvision? ;)

spookyvision:
> Marking your own homework @spookyvision? ;)

having a full blown PR for trivial non-code fixes is already more noise than I like, adding a reviewer to the mix even worse in my book. Ultimately I don't care much either way though.

PR #641, opened 2021-11-26
Release v0.3.1

jonathanpallant:
Updates all crates to v0.3.1.

The alternative is to only update the changed crates:

* defmt-macros
* defmt-decoder
* defmt-rtt

### DIFF
--- a/firmware/defmt-rtt/Cargo.toml
+++ b/firmware/defmt-rtt/Cargo.toml
@@ -8,7 +8,7 @@ license = "MIT OR Apache-2.0"
 name = "defmt-rtt"
 readme = "README.md"
 repository = "https://github.com/knurling-rs/defmt"
-version = "0.3.0"
+version = "0.3.1"
 
 [dependencies]
 cortex-m = "0.7"
--- a/firmware/defmt-rtt/Cargo.toml
+++ b/firmware/defmt-rtt/Cargo.toml
@@ -8,8 +8,8 @@ license = "MIT OR Apache-2.0"
 name = "defmt-rtt"
 readme = "README.md"
 repository = "https://github.com/knurling-rs/defmt"
-version = "0.3.0"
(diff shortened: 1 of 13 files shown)


firmware/defmt-rtt/Cargo.toml
```diff
@@ -8,8 +8,8 @@ license = "MIT OR Apache-2.0"
 name = "defmt-rtt"
 readme = "README.md"
 repository = "https://github.com/knurling-rs/defmt"
-version = "0.3.0"
+version = "0.3.1"
 
 [dependencies]
 cortex-m = "0.7"
-defmt = { version = "0.3", path = "../../defmt" }
```
Urhengulas, line 15: On a more philosophical note (aka. I am not sure myself what the best option is): Since `defmt-rtt` doesn't pin the version of `defmt` we are not required to bump it, afaik, but it will just pick up more recent versions. The question therefore is, if we should do that, or just leave it at `0.3.0` until `0.4.0`.

Urhengulas: In the end this doesn't really matter, it just occurred to me when seeing it.

Urhengulas (changes requested):
I think you forgot to bump `defmt-decoder` as well.

firmware/defmt-rtt/Cargo.toml
jonathanpallant: I would just leave it, until there is a change that defmt-rtt requires in order to function correctly.

Urhengulas: Makes sense. Are you going to revert it then?

Urhengulas: You need to revert anyway, because there is no `defmt 0.3.1`

Urhengulas (approved):

Issue #632, opened 2021-11-24
Negative values for i16 formatted as hexadecimals have enough FFs for an i128.


mehmetalianil:
When I print out a i16 as an i16 as in here:
```rust
defmt::debug!("x: {=i16:#x},y: {=i16:#x},z: {=i16:#x}", 
                x,
                y,
                z);
```
I get a twos complement value, but for negatives, I get enough `f`s for an i128:
```
DEBUG x: 0xfffffffffffffffffffffffffffffeda,y: 0xfffffffffffffffffffffffffffffd30,z: 0x40b3
```

I don't say that it is a bug, but could be deceiving.

Urhengulas:
@mehmetalianil Thank you for your issue report.

I have a slight problem understanding the problem. Could you help me understand it please? Therefore it might be helpful if you could elaborate your code example with the input values you used (the types and values for x, y, z).

> When I print out a i16 as an i16

Did you mean to use `i16` two times, or is one of them meant to be a different number type?

justahero:
Hello @Urhengulas,

the following steps I used to reproduce the behaviour.

Create a new project via [app-template](https://github.com/knurling-rs/app-template) repository.
Then replace the content of one of the programs, e.g. `hello.rs`, with the following code

```rust
#![no_main]
#![no_std]

use defmt_test_example as _; // global logger + panicking-behavior + memory layout

#[cortex_m_rt::entry]
fn main() -> ! {
    defmt::info!("x: {=i16:#x}", -1_i16);
    defmt_test_example::exit()
}
```

Run the program with `DEFMT_LOG=info cargo rb hello`, the generated output is as follows:

```
0 INFO  x: 0xffffffffffffffffffffffffffffffff
└─ hello::__cortex_m_rt_main @ src/bin/hello.rs:8
```

The issue here is, the `i16` value is represented with the [`Arg::Ixx`](https://github.com/knurling-rs/defmt/blob/main/decoder/src/lib.rs#L282) variant. Unfortunately when the resulting `Frame` for hex values is formatted in [Frame::format_i128](https://github.com/knurling-rs/defmt/blob/1bd226b03e80d02bb02064bfcdac36291aa2f52a/decoder/src/frame.rs#L261-L289) the integer is treated as a `i128` type. The displayed value is not handled as `-1_i16` but rather as `-1_i128`, therefore the hex value is filled with `f`.

Urhengulas:
I see. Thank you @justahero! 😄

[...]

PR #646, opened 2021-12-01
Fix formating of Arg::Ixx based integer hex values

justahero:
This PR solves #632, it formats signed integers as hex values based on their given type, instead of formating it using a  `i128` based value.

Before this change the following log code:

```rust
defmt::info!("x: {=i16:#x}", -1_i16);
```

produced the output:

```
0 INFO  x: 0xffffffffffffffffffffffffffffffff
└─ hello::__cortex_m_rt_main @ src/bin/hello.rs:8
```

After the change the decoder takes the initial integer type into account when formating integer based hex values, the output is now:

```
0 INFO  x: 0xffff
└─ hello::__cortex_m_rt_main @ src/bin/hello.rs:8
```

The reason is that the decoder parsed all integer values as either [`Arg::Ixx`](https://github.com/knurling-rs/defmt/blob/main/decoder/src/lib.rs#L282) or [`Arg::Uxx`](https://github.com/knurling-rs/defmt/blob/main/decoder/src/lib.rs#L280) variants with the biggest data type to store the value in (`i128` & `u128`), while keeping the original type information in a separate [`Parameter`](https://github.com/knurling-rs/defmt/blob/main/parser/src/lib.rs#L21-L28) as part of the `Frame`.

The change now passes the integer type information to function [`Frame::format_i128`](https://github.com/knurling-rs/defmt/blob/main/decoder/src/frame.rs#L261). The new helper struct `Hex128`  implements both the [`LowerHex`](https://doc.rust-lang.org/std/fmt/trait.LowerHex.html) and [`UpperHex`](https://doc.rust-lang.org/std/fmt/trait.UpperHex.html) traits by delegating the call to the appropriate type implementation.

* fix formating of integer based hex values
* expand snapshot program `log` to output a number of different hex values


decoder/src/frame.rs
```diff
@@ -9,6 +9,35 @@ use chrono::TimeZone;
 use colored::Colorize;
 use defmt_parser::{DisplayHint, Fragment, Level, ParserMode, TimePrecision, Type};
 
+/// Used to convert a `i128` value into right target type in hex
+struct Hex128(i128, Type);
+
+impl std::fmt::LowerHex for Hex128 {
+    fn fmt(&self, f: &mut fmt::Formatter<'_>) -> fmt::Result {
+        match self.1 {
+            Type::I8 => fmt::LowerHex::fmt(&(self.0 as i8), f),
```
jonathanpallant, line 18: What about unsigned types? If not required, we should add a comment to explain why they are missing.

justahero: The struct `Hex128` is renamed to `I128Hex` to indicate this applies only to signed integers. Unsigned integers are formatted correctly as hex numbers. The snapshot test `log` also outputs a line with unsigned integers.

jonas-schievink:
Nice!

bors r+

PR #647, opened 2021-12-15
Fail parsing of missing display hint

justahero:
Before this change a format string was allowed to contain `{:}` without a display hint, which resulted in a `DisplayHint::NoHint`.


japaric (approved):
Agree on this, theoretically, being a breaking-change that's likely to go unnoticed.

PR #651, opened 2022-01-11
`probe-run` json output

Urhengulas:
This PR adds the new `trait log::Log` implementation `JsonLogger`, which can be used to output the logs as json. Adding this new output required a bit of refactoring in order to handle the `PRINTLN` log-level.

<details>

<summary>This code <i>(click to expand)</i> ...</summary>

```rust
#![no_main]
#![no_std]

use defmt::Format;
use krate as _; // global logger + panicking-behavior + memory layout

#[cortex_m_rt::entry]
fn main() -> ! {
    defmt::println!("I am a {}-statement", "PRINTLN");

    let s = S { a: 8 };
    s.abc();

    test_m::inner_test_m::test_f();

    panic!("P A N I C");
}

#[derive(Format)]
struct S {
    a: u8,
}

impl S {
    fn abc(&self) {
        defmt::trace!("{}", self);
    }
}

mod test_m {
    pub mod inner_test_m {
        pub fn test_f() {
            defmt::warn!("I am nested in {} modules.", 2);
        }
    }
}
```

</details>

... produces following output:

```console
$ DEFMT_LOG=trace cargo rb hello

(HOST) INFO  flashing program (2 pages / 8.00 KiB)
└─ probe_run @ src/main.rs:93
(HOST) INFO  success!
└─ probe_run @ src/main.rs:126
{"data":"I am a PRINTLN-statement","decoder_version":"0.3.1","host_timestamp":1642688547813629646,"level":"PRINTLN","location":{"file":"src/bin/hello.rs","line":9,"module_path":{"crate_name":"hello","modules":[],"function":"__cortex_m_rt_main"}},"target_timestamp":"0"}
{"data":"S { a: 8 }","decoder_version":"0.3.1","host_timestamp":1642688547813863496,"level":"TRACE","location":{"file":"src/bin/hello.rs","line":26,"module_path":{"crate_name":"hello","modules":["{impl#0}"],"function":"abc"}},"target_timestamp":"1"}
{"data":"I am nested in 2 modules.","decoder_version":"0.3.1","host_timestamp":1642688547814069506,"level":"WARN","location":{"file":"src/bin/hello.rs","line":33,"module_path":{"crate_name":"hello","modules":["test_m","inner_test_m"],"function":"test_f"}},"target_timestamp":"2"}
{"data":"panicked at 'P A N I C', src/bin/hello.rs:16:5","decoder_version":"0.3.1","host_timestamp":1642688547814328323,"level":"ERROR","location":{"file":"\u001b[2m/home/urhengulas/.cargo/registry/src/github.com-1ecc6299db9ec823\u001b[0m/\u001b[1mpanic-probe-0.3.0\u001b[0m/src/lib.rs","line":91,"module_path":{"crate_name":"panic_probe","modules":["print_defmt"],"function":"print"}},"target_timestamp":"3"}
stack backtrace:
   0: HardFaultTrampoline
      <exception entry>
   1: lib::inline::__udf
        at ./asm/inline.rs:172:5
   2: __udf
        at ./asm/lib.rs:49:17
   3: cortex_m::asm::udf
        at /home/urhengulas/.cargo/registry/src/github.com-1ecc6299db9ec823/cortex-m-0.7.3/src/asm.rs:43:5
   4: rust_begin_unwind
        at /home/urhengulas/.cargo/registry/src/github.com-1ecc6299db9ec823/panic-probe-0.3.0/src/lib.rs:72:9
   5: core::panicking::panic_fmt
        at /rustc/02072b482a8b5357f7fb5e5637444ae30e423c40/library/core/src/panicking.rs:107:14
   6: core::panicking::panic
        at /rustc/02072b482a8b5357f7fb5e5637444ae30e423c40/library/core/src/panicking.rs:48:5
   7: hello::__cortex_m_rt_main
        at src/bin/hello.rs:16:5
   8: main
        at src/bin/hello.rs:7:1
   9: Reset
(HOST) ERROR the program panicked
└─ probe_run::backtrace @ src/backtrace/mod.rs:105
```

The backtrace and the `–`-lines go to stderr, while the json frames goes to stdout, and look like following if you format them:

```json
{
    "data": "I am a PRINTLN-statement",
    "decoder_version": "0.3.1",
    "host_timestamp": 1642686487770043158,
    "level": "PRINTLN",
    "location": {
        "file": "src/bin/hello.rs",
        "line": 9,
        "module_path": {
            "crate_name": "hello",
            "modules": [],
            "function": "__cortex_m_rt_main"
        }
    },
    "target_timestamp": "0"
}
{
    "data": "S { a: 8 }",
    "decoder_version": "0.3.1",
    "host_timestamp": 1642686487770289902,
    "level": "TRACE",
    "location": {
        "file": "src/bin/hello.rs",
        "line": 26,
        "module_path": {
            "crate_name": "hello",
            "modules": [
                "{impl#0}"
            ],
            "function": "abc"
        }
    },
    "target_timestamp": "1"
}
{
    "data": "I am nested in 2 modules.",
    "decoder_version": "0.3.1",
    "host_timestamp": 1642686487770518118,
    "level": "WARN",
    "location": {
        "file": "src/bin/hello.rs",
        "line": 33,
        "module_path": {
            "crate_name": "hello",
            "modules": [
                "test_m",
                "inner_test_m"
            ],
            "function": "test_f"
        }
    },
    "target_timestamp": "2"
}
{
    "data": "panicked at 'P A N I C', src/bin/hello.rs:16:5",
    "decoder_version": "0.3.1",
    "host_timestamp": 1642686487770772892,
    "level": "ERROR",
    "location": {
        "file": "/home/urhengulas/.cargo/registry/src/github.com-1ecc6299db9ec823/panic-probe-0.3.0/src/lib.rs",
        "line": 91,
        "module_path": {
            "crate_name": "panic_probe",
            "modules": [
                "print_defmt"
            ],
            "function": "print"
        }
    },
    "target_timestamp": "3"
}
```

**TODO**
- [x] ~~add `column` information~~ -> blocked
- [x] ~~`is_method` information~~ -> omitted
- [x] Should we keep the `{impl#0}`s in the module path, or drop them? -> keep them
- [x] backtrace -> backtrace is re-directed to stderr and therefore doesn't interfere with output anymore
- [x] handle host logs -> added `is_host` attribute
    - [x] Q: `is_host` is redundant to some extend. This is, because (in the current implementation) it will always be `true`, if `target_timestamp` is `none` and if `target_timestamp` isn't `none`, then `is_host` doesn't exist. If one knows about this relationship it is easy to infer. Should we therefore drop it?
- [ ] Transform `target_timestamp` to `seconds` and `nanos` (as in the RFC) or keep the string version?
- [ ] document in book
  - mention schema
  - mention `--json` flag
  - what goes to to stdout, stderr

Fixes https://github.com/knurling-rs/probe-run/issues/178.


decoder/src/log/json_logger.rs
```diff
@@ -0,0 +1,97 @@
...
+    pub fn new(should_log: impl Fn(&log::Metadata) -> bool + Sync + Send + 'static) -> Box<Self> {
+        Box::new(Self {
+            should_log: Box::new(should_log),
+        })
+    }
+}
+
+fn extract_path(
+    module_path: Option<&str>,
+) -> (
```
japaric, line 73: this function should return `Option<(String, Vec<String>, String, bool)>` IMO because it'll never return (Some, None, Some, None) for example

decoder/src/log/json_logger.rs
```diff
@@ -0,0 +1,97 @@
...
+    let idx = module_path.len() - 1;
+
+    let krate = module_path[..1][0].to_string();
+    let modules = module_path[1..idx]
+        .iter()
+        .map(|b| b.to_string())
+        .collect::<Vec<_>>();
+    let function = module_path[idx..][0].to_string();
+
+    let is_method = modules.iter().any(|m| m.starts_with("{impl#"));
```
japaric, line 94: I'm not sure how stable the "{impl(..)}" segment is. I think old versions of Rust did not emit a symbol like that. I think it would be better for semver purposes to not emit an `is_method` and instead let end users implement the is_method logic if they need it.

decoder/src/log/json_logger.rs
```diff
@@ -0,0 +1,97 @@
...
+) {
+    let module_path = if let Some(a) = module_path {
+        a.split("::").collect::<Vec<_>>()
+    } else {
+        return (None, None, None, None);
+    };
+
+    let idx = module_path.len() - 1;
+
+    let krate = module_path[..1][0].to_string();
```
japaric, line 87: this slicing will panic if `module_path` does not contain the module path separator "::". I would suggest returning None from this functio in that case.

decoder/src/log/pretty_logger.rs
```diff
@@ -1,123 +1,119 @@
...
+                // defmt goes to stdout, since it's the primary output produced by this tool.
+                let stdout = io::stdout();
+                let sink = stdout.lock();
 
-    /// Returns the formatted defmt timestamp.
-    pub fn timestamp(&self) -> &str {
-        self.timestamp
-    }
+                match record.is_println() {
+                    false => Self::print_defmt_record(self, record, sink),
```
japaric, line 38: I would suggest method syntax here: self.print_defmt_record

decoder/src/log/json_logger.rs
Urhengulas: That is also what I prefer, but I included it for now, since it is part of the RFC.

decoder/src/log/pretty_logger.rs
Urhengulas: okeee ;)

decoder/src/log/json_logger.rs
```diff
@@ -68,30 +65,27 @@ impl JsonLogger {
...
-) {
-    let module_path = if let Some(a) = module_path {
-        a.split("::").collect::<Vec<_>>()
-    } else {
-        return (None, None, None, None);
-    };
-
-    let idx = module_path.len() - 1;
+#[derive(Serialize)]
+struct Path {
```
japaric: I would suggest naming this `ModulePath` so that it's not mistaken with `std::path::Path`

decoder/src/log/json_logger.rs
```diff
@@ -64,15 +61,31 @@ impl JsonLogger {
...
+    path: Option<Path>,
+    target_timestamp: String,
+
+    // backtrace is omitted from output if it is `None`
+    #[serde(skip_serializing_if = "Option::is_none")]
+    backtrace: Option<Vec<()>>,
+}
+
+#[derive(Debug, Deserialize, Serialize)]
+pub struct Path {
```
japaric, line 78: if this now includes the module path and the source code location maybe call it `Location` -- here and in the Json field.

japaric (commented):
Looking good

decoder/src/log/json_logger.rs
Urhengulas: The problem is that we already have a public `defmt_decoder::elf2table::Location`. I will try to come up with better names.

Urhengulas: Renamed it to Location anyways.

japaric:
some thoughts on the latest semantic changes mentioned in the PR description (have not looked at the code yet)

I'm not sure the host messages should be output as JSON. there's only a few of them and the main interest are the logs produced by the firmware. I would leave the host logs as human readable text printed to stderr. that way you can get some feedback when piping probe-run to something else, e.g.

``` console
$ cargo r --bin app | forward-logs
flashed firmware
running firmware
(maybe long pause here)
program finished without errors
```

it would be good to document the schema of the JSON output stating which fields are guaranteed to be there and which ones are nullable. or even better provide a crate with the schema as a Rust struct that implements Deserialize/Serialize. this is also related to my previous point: if we leave out the host messages the schema can become simpler and there would be less nullable fields I think.

I'm not sure the decoder_version is too useful in the JSON messages. the messages are produced *after* decoding so the consumer of the messages is probably not too interested in the decoder version or its configuration. what would be more relevant is the version of the JSON format itself. if we include a "schema_version" field in each JSON object the receiver can probe the first message to decide which schema version to use. going back to the previous point, a schema crate could provide different version of the `JsonFrame`: `schema::v1::JsonFrame`, `schema::v2::JsonFrame`. it does seem a bit wasteful to repeat the `schema_version` field in every JSON object so we could always start the JSON stream with a "header" JSON object that contains the schema_version then we can omit the schema_version from all the following JSON objects.

Urhengulas:
Thank you for the review! 😁 

### host logs
I've happily removed the host-logs from the equation, because, as you said, it makes the schema easier. This changed required to add an instance of the `PrettyLogger` as an attribute to the `JsonLogger`, which I kinda don't like, but it works.

### schema version
I was thinking about this quite a bit. I've chosen to not have a `schema_version` instead of the `decoder_version`, because this would be another version to coordinate/bump/care about. Having the `decoder_version` a user can use the `JsonFrame` from that particular decoder version for decoding the frame and building their code on top. The `JsonFrame` is exposed as `defmt_decoder::log::JsonFrame`.

But (!) I can see that someone who just needs the `JsonFrame` wouldn't want to depend `defmt_decoder` since that pulls in many dependencies, which is a strong argument.

Therefore I would go forward with moving the schema to a separate crate. Do you have a good name? `defmt-schema`? `defmt-json-schema`?

### version in messages vs. header
I think it makes sense to have the version in each message.

In a scenario where one target logs into one datastore and is only using one schema version that seems a bit unnecessary.

In another scenario where multiple targets with potentially different schema versions log into the same datastore the story might look a bit different. The user might process/analyze the messages a "long" time after they got emitted and therefore needs to know which message used which version.
With a version header this is obviously still possible, but puts more effort on the user, since they need to annotate each message.

japaric (approved):
Looks great!

PR #655, opened 2022-01-25
`book`: Document probe-run json mode

Urhengulas:
Document the json output, which was implemented in #651.

Rendered version: https://deploy-preview-655--admiring-dubinsky-56dff5.netlify.app/json-output.html

### DIFF
--- a/book/src/json-output.md
+++ b/book/src/json-output.md
@@ -88,6 +88,9 @@ If you wish to deserialize the entire data back into a Rust program, you will ne
 You can use all of this together with `serde_json` like following:
 
 ```rust
+# extern crate defmt_json_schema;
+# extern crate serde_json;
+
 use defmt_json_schema::{v1, SchemaVersion};
 
 const DATA: &str = r#"{"schema_version":1}
--- a/book/src/json-output.md
+++ b/book/src/json-output.md
@@ -78,104 +78,47 @@ It indicates the version of the json format you are using. `probe-run` will alwa
 
 > 🤔: So, what can I do with the JSON output?
 
-There really are no boundaries. You can process the JSON with any programming language you like. If you want to do that in rust, we provide some data transfer objects which implement `serde`'s `trait Serialize` and `trait Deserialize`, in the [`defmt-json-schema`] crate.
+There really are no boundaries. You can process the JSON with any programming language you like and also store it any data store of your choice to process and analyze it later. If you are saving the output for later, it might make sense to store the schema version together with additional metadata like e.g. a device id or firmware version. 
 
-See how they can be use in [this example](https://github.com/knurling-rs/defmt/blob/b396257be01f477eda2ac16f7e62dece31749963/decoder/defmt-json-schema/examples/simple.rs) and how they are defined below.
-
-### JsonFrame
-
-The most important one is the `struct JsonFrame`. It contains all the information for one log-statement.
-
-It is defined like this:
-
-```rust
-pub struct JsonFrame {
-    pub data: String,
-    pub host_timestamp: i64,
-    pub level: Option<log::Level>,
-    pub location: Location,
-    pub target_timestamp: String,
-}
-
-pub struct Location {
-    pub file: Option<String>,
-    pub line: Option<u32>,
-    pub module_path: Option<ModulePath>,
-}
-
-pub struct ModulePath {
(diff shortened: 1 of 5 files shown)


book/src/json-output.md
```diff
@@ -0,0 +1,157 @@
+# JSON output
+
+> Structured logging for `probe-run`.
+
+As an alternative to it's human-focused output, `probe-run` offers structured json output. There are two use-cases:
```
jonathanpallant, line 5: I think it should be JSON not json throughout. Except when it's an argument like `--json-mode`.

jonathanpallant, line 5: Its not it's. *it's* == *it is*, while *its* is the gender neutral version of *his* or *hers*.

book/src/json-output.md
```diff
@@ -0,0 +1,157 @@
...
+    pub modules: Vec<String>,
+    pub function: String,
+}
+
+# mod log { pub struct Level; }
+```
+
+... which results in following json schema ...
+
+```json
```
jonathanpallant, line 109: This is JSON Schema, not JSON. The standard Github JSON renderer makes a mess of it.

book/src/json-output.md
```diff
@@ -0,0 +1,157 @@
...
+            "modules": [],
+            "function": "__cortex_m_rt_main"
+        }
+    },
+    "target_timestamp": "1"
+}
+```
+
+## SchemaVersion
+
```
jonathanpallant, line 148: What's this for, or how would you use it?

book/src/json-output.md
```diff
@@ -0,0 +1,157 @@
...
+It indicates the version of the json format you are using. `probe-run` will always output it as a header at the beginning of each stream of logs. We anticipate that the format will slightly change while `probe-run` and `defmt` evolve. Using this version you always know which revision is in use and can act upon that.
+
+🤗: Thank you for the explanation!
+
+## JsonFrame
+
+The [`defmt-json-schema`] crate provides a data transfer object `struct JsonFrame` which is used to serialize the log frame to json and can be used to deserialize it back to rust as well.
+
+The `JsonFrame` is defined like this ...
```
jonathanpallant, line 81: It might be worth an example of loading a saved JSON file and `{:?}` printing all the objects.

jonathanpallant (commented):
Looks really good. Just a few minor notes.

book/src/json-output.md
Urhengulas: It does look bad on github, but when rendered in the mdbook it looks good I'd say.

book/src/json-output.md
Urhengulas: Like this? 😁 https://github.com/knurling-rs/defmt/blob/1dea0c1de8469ba27a3b56760a9b98f22870aaaf/decoder/defmt-json-schema/examples/simple.rs#L1-L18

I will link it here.

book/src/json-output.md
Urhengulas: Are you talking about the example of the JSON output, or the SchemaVersion?

jonathanpallant: The SchemaVersion. It wasn't clear what "can be used to deserialise the JSON back to Rust" had to do with a struct containing a field called `schema_version`. So maybe an example would help?

book/src/json-output.md
jonathanpallant: OK, well I guess most people are looking at the MdBook.

book/src/json-output.md
jonathanpallant: Oh wait, I think I get it. Maybe add something like:

> If you wish to deserialise the entire file back into a Rust program, you will need to be able to decode the `SchemaVersion` object at the start of the file, as well as the `JsonFrame` objects. To do that, we supply a `SchemaVersion` struct in .... which you can use by ....

book/src/json-output.md
jonathanpallant: Yes, that does help things make sense!

book/src/json-output.md
Urhengulas: I've reformulated it. What do you think now?

jonathanpallant: Nice :) 👍

Urhengulas:
bors r=@jonathanpallant

PR #656, opened 2022-01-29
defmt: implements Format for cells

DBLouis:
This patch implement the `Format` trait for `Cell<T>` and `RefCell<T>`

### DIFF
--- a/defmt/src/impls/core_/cell.rs
+++ b/defmt/src/impls/core_/cell.rs
@@ -1,0 +1,46 @@
+use super::*;
+
+impl<T> Format for core::cell::Cell<T>
+where
+    T: Format + Copy,
+{
+    fn format(&self, fmt: Formatter) {
+        crate::write!(fmt, "Cell {{ value: {=?} }})", self.get())
+    }
+}
+
+impl<T> Format for core::cell::RefCell<T>
+where
+    T: Format,
+{
+    default_format!();
+
+    #[inline]
+    fn _format_tag() -> Str {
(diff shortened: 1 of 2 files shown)


Urhengulas:
Hi @DBLouis, Thank you for your PR! Could you please move your code to `defmt/src/impls/core_/cell.rs` and add that as a new `mod`?

Urhengulas:
Thank you!

Secondly, I'd propose to stay close to the format strings in `std`. 

```rust
```

Urhengulas:
Thank you!

Secondly, I'd propose to stay close to the representations in `std`. 

This code:
```rust
use std::cell::{Cell, RefCell};

fn main() {
    let a = Cell::new(5);
    println!("{:?}", a);

    let b = RefCell::new(5);
    println!("{:?}", &b);

    let _c = b.try_borrow_mut().unwrap();
    println!("{:?}", &b);

    let d = b.try_borrow_mut().unwrap_err();
    println!("{:?}", d);
}
```
Produces that output:
```
Cell { value: 5 }
RefCell { value: 5 }
RefCell { value: <borrowed> }
BorrowMutError
```

_(I couldn't figure out how to trigger a `BorrowError`, but it should be just "BorrowError")_

Do you want to quickly change that? Else I can also do it.

DBLouis:
You can trigger a `BorrowError` by borrowing mutably first then attempting to borrow immutably [Playground](https://play.rust-lang.org/?version=stable&mode=debug&edition=2021&gist=999396095cf8582f4ec84bb3ac771da2)

Urhengulas:
Thank you! 

bors r+

PR #659, opened 2022-02-10
Make export::acquire() and export::release() unsafe

jannic:
Calling export::release() may enable interrupts, which is unsound
if done inside a critical section or other code which expects interrupts
to be disabled.

Calling export::acquire() is less dangerous (no obvious way to cause
unsoundness), but is still a bad idea and breaks the safety contract
of critical_section::acquire(), which must only be called paired with
critical_section::release().


Urhengulas:
Thank you!

bors r+

PR #662, opened 2022-02-17
Add attributes for Format derive to use Debug2Format on specific fields

mattico:
I often have structs or enums where most fields impl `Format`, but a few third-party types do not. It is a pain to have to manually impl `Format` for the whole type when only one field doesn't work. I added an attribute that lets you override specific fields to use `Debug2Format` or `Display2Format` like so:

```rust
#[derive(defmt::Format)]
pub enum Error {
    Serial(serial::Error),
    Timeout(us_timer::TimeoutError),
    Serde(#[defmt(Debug2Format)] serde_json::Error),
    ResponseTooLarge,
}
```

### DIFF
--- a/macros/src/derives/format/codegen/fields.rs
+++ b/macros/src/derives/format/codegen/fields.rs
@@ -2,7 +2,7 @@ use std::fmt::Write as _;
 
 use proc_macro2::TokenStream as TokenStream2;
 use quote::{format_ident, quote};
-use syn::{Fields, Index, Type};
+use syn::{Attribute, Fields, Index, Meta, NestedMeta, Type};
 
 use crate::consts;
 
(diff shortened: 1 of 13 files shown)


Urhengulas:
Looks really good! Thank you @mattico!

Urhengulas (approved):

Urhengulas:
Since I am not super proficient with macros I'd like @jonas-schievink to take a quick look before merging.

jonas-schievink:
Great to see this, thanks!

bors r+

macros/src/derives/format/codegen/fields.rs
```diff
@@ -36,31 +36,33 @@ pub(crate) fn codegen(
...
-        if let Some(ident) = field.ident.as_ref() {
-            write!(format_string, "{}: {{={}:?}}", ident, ty).ok();
+        let ident = field
+            .ident
+            .clone()
+            .unwrap_or_else(|| format_ident!("arg{}", index));
+
+        if let Some(FormatOption::Debug2Format) = format_opt {
+            stmts.push(quote!(defmt::export::fmt(&defmt::Debug2Format(&#ident))));
+        } else if let Some(FormatOption::Display2Format) = format_opt {
```
mattico, line 48: I suppose these didn't need to be `if let`s

Urhengulas: True, either normal if-else or match might be better suited.

PR #666, opened 2022-03-10
make defmt::flush an no-operation when unstable-test feature is enabled

japaric:
that feature is only enabled when running the test suite

defmt::flush was causing linker errors ("undefined symbol _defmt_acquire") on Windows with Rust
1.59.0+ when defmt is built with 'unstable-test' enabled. the linker errors are not observed in
binary builds that link to defmt when the feature is disabled

this commit removes the symbol dependencies to _defmt_acquire, etc. when 'unstable-test' is enabled


jonathanpallant (approved):

japaric:
bors r=jonathanpallant

Urhengulas:
I honestly still don't understand why this was a problem for `flush`, but not for `acquire`, `release`. Can someone explain?

PR #669, opened 2022-03-22
--json docs + Readme

Dajamante:
- completed documentation (v0.3.3+ is necessary to use the `--json` flag)
- example of json output with formatter
- small link broken in the readme

### DIFF
--- a/book/src/json-output.md
+++ b/book/src/json-output.md
@@ -71,13 +71,7 @@ Afterwards `levels.json` looks like this:
 {"data":"println","host_timestamp":1643113389707313290,"level":null,"location":{"file":"src/bin/levels.rs","line":15,"module_path":{"crate_name":"levels","modules":[],"function":"__cortex_m_rt_main"}},"target_timestamp":"4"}
 ```
 
-> 🤔: That seems convenient, but what is this schema version in the first line?
-
-It indicates the version of the json format you are using. `probe-run` will always output it as a header at the beginning of each stream of logs. We anticipate that the format will slightly change while `probe-run` and `defmt` evolve. Using this version you always know which revision is in use and can act upon that.
-
-> 🤔: What if I want formatted output?
-
-You can use `probe-run --chip some-chip --json my-elf | my-formatter`, with for example `jq` on Linux. It will give you this output.
+One use case is changing the human readable output with a piped command. 
 
 ```json
 {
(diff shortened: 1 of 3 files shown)


japaric (commented):
thanks. I left some suggestions.

book/src/json-output.md
```diff
@@ -68,6 +71,19 @@ Afterwards `levels.json` looks like this:
 {"data":"println","host_timestamp":1643113389707313290,"level":null,"location":{"file":"src/bin/levels.rs","line":15,"module_path":{"crate_name":"levels","modules":[],"function":"__cortex_m_rt_main"}},"target_timestamp":"4"}
 ```
 
+One use case is changing the human readable output with a piped command.
```
japaric, line 74: I would suggest moving this addition to the very end and mention that `jq` is an existing tool that can be used to change the output format. But again, the example output should be be JSON but something more readable.

japaric (commented):
Not part of this PR but could you fix this line; it has a missing link:

>  To do that, we supply a few things in [defmt_json_schema]:

Dajamante:
> To do that, we supply a few things in [defmt_json_schema]:

I don't think the link is very informative... 
https://crates.io/crates/defmt-json-schema

I will remove the linking instead, since Johann wrote extensive information.

book/src/json-output.md
japaric: > But again, the example output should be be JSON but something more readable.

*should **not** be JSON*.

by "example output" I meant the code block. Ideally, it should be something like

``` console
$ cargo r --example foo | jq "something here"
I am a PRINTLN-statement
the next log statement
```

japaric:
> I don't think the link is very informative...

I think the link still helps because you can tell it's a crate and what version it's available.

book/src/json-output.md
Dajamante: > I would suggest moving this addition to the very end and mention that `jq` is an existing tool that can be used to change the output format. But again, the example output should be be JSON but something more readable.

Dumb one, do you mean at the very end of the page or the very end of the paragraph? I fear it will be a bit out of context at the very end of the page.

japaric: I meant at the end of the page. After re-reading the page I think a good place may be after the first paragraph of the "Data transfer objects"

> So, what can I do with the JSON output?
> There really are no boundaries. (..)
> (NEW) One option is to use a program like `jq` to extract the parts of interest. *blah* *blah* *show console example*
> If you wish to deserialize (..)

book/src/json-output.md
```diff
@@ -72,6 +75,47 @@ Afterwards `levels.json` looks like this:
 
 It indicates the version of the json format you are using. `probe-run` will always output it as a header at the beginning of each stream of logs. We anticipate that the format will slightly change while `probe-run` and `defmt` evolve. Using this version you always know which revision is in use and can act upon that.
 
+> 🤔: What if I want formatted output?
```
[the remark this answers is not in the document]
Dajamante: I can also remove the whole reference to `jq`, as it is not relevant information here.

Dajamante:
> > I don't think the link is very informative...
> 
> I think the link still helps because you can tell it's a crate and what version it's available.

Sure I can put the link.

book/src/json-output.md
Dajamante: Maybe that sentence is enough. Everybody already saw a json output. The console example does not bring anything.

PR #675, opened 2022-05-31
Prepare `defmt 0.3.2` release

Urhengulas:
This release is due to a compilation error which occurs if `defmt 0.3.1` is used together with `defmt-macros 0.3.1` or `defmt-macros 0.3.0`. Therefore we increase the minimal version of `defmt-macros` to `0.3.2`.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -7,6 +7,11 @@ and this project adheres to [Semantic Versioning](http://semver.org/).
 
 ## [Unreleased]
 
+## [0.3.3] - 2022-05-31
+
+- [#675] Release `defmt 0.3.2` and fix `defmt-macros`-releated compile-error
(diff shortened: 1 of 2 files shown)


CHANGELOG.md
```diff
@@ -7,6 +7,11 @@ and this project adheres to [Semantic Versioning](http://semver.org/).
 
 ## [Unreleased]
 
+## [0.3.3] - 2022-05-31
```
japaric, line 10: I think the numbering in the changelog is somehow wrong because this section should be 0.3.2 but there's already a section named 0.3.2 :thinking:

japaric (commented):
cargo.toml changes look good to me. left a question about the changelog

Urhengulas:
bors r=japaric

PR #681, opened 2022-06-24
Make use of i/o locking being static since rust `1.61`.

Urhengulas:
See https://blog.rust-lang.org/2022/05/19/Rust-1.61.0.html#static-handles-for-locked-stdio.


BriocheBerlin (approved):
LGTM!

PR #683, opened 2022-07-08
Make sure the whole RTT structure is in RAM

jannic:
Currently, the name field of the RTT channel will be placed in
flash (rodata). That causes the debug probe to read from flash while
parsing the RTT header.

Usually this doesn't matter, but if the firmware disables flash access
before the debug probe scanned the header, it will break.

The situation where I stumbled over this was writing a firmware which
writes to flash on an RP2040. While writing to the (external) flash
chip, no concurrent flash accesses must happen.

This patch forces the name to the `.data` section, so the whole RTT
header can be read from RAM.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -7,11 +7,13 @@ and this project adheres to [Semantic Versioning](http://semver.org/).
 
 ## [Unreleased]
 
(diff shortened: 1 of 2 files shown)


Urhengulas:
bors r=jonas-schievink

PR #684, opened 2022-07-10
Fix syn dependency version

Disasm:
`syn::Error::into_compile_error`, which is used in `defmt-macros`, is available since `1.0.56`, this PR updates the dependency version to fix build issues in some environments.


Urhengulas:
Thank you!

bors r+

Urhengulas:
bors retry

Urhengulas:
bors retry

PR #699, opened 2022-10-04
Non atomic RTT

### LINKED ISSUE #597: Support for targets without native atomics
The `esp32c3` is a riscv based cpu, _without_ the atomic extension. Unlike the arm `thumbv6` target, it does _not_ have atomic load/stores for the following types `u8`, `u16`, `u32`, which means we cannot use `defmt` (and other rtt based crates) as it stands.

I recently landed a PR in `atomic-polyfill` that adds load/store for those types mentioned above, implemented using critical sections. I [have a PR](https://github.com/mvirkkunen/rtt-target/pull/21) in `rtt-target` to use that, but it has not yet been accepted. 

In theory for better performance, we could ditch the critical section and implement our own atomic operation that use fences to ensure atomicity (this is what is done in llvm for `thumbv6`) - however this can cause UB when mixed with lock based CAS atomics (see: [here](https://github.com/rust-lang/rust/pull/81752), [here](https://reviews.llvm.org/D47553), and [here](https://gcc.gnu.org/bugzilla/show_bug.cgi?id=86005) ) therefore is not suitable to add to `atomic-polyfill` as it stands.

I would like the esp32c3 to be able to support the c3, and I am sure there are other riscv, non-a targets too.

What do you think the best idea is going forward?

Urhengulas:
This PR replaces all atomic variables with their non-atomic counterpart wrapped in a `critical_section::Mutex<RefCell<_>>`. The `RefCell` is needed, because `critical_section::Mutex` does not provide interior mutability.

This change makes `defmt-rtt` available on targets without native atomics. 
Fixes #597.

Atm, it is unclear what the runtime and code-size impact of that change is. Likely both would improve if using a `Cell`, instead of a `RefCell`, but it is unclear if that would be correct.

### DIFF
--- a/firmware/defmt-rtt/src/channel.rs
+++ b/firmware/defmt-rtt/src/channel.rs
@@ -1,7 +1,4 @@
-use core::{
-    ptr,
-    sync::atomic::{AtomicUsize, Ordering},
-};
+use core::{cell::RefCell, ptr};
 
 use crate::{consts::BUF_SIZE, MODE_BLOCK_IF_FULL, MODE_MASK};
 
@@ -13,13 +10,13 @@ pub(crate) struct Channel {
     pub buffer: *mut u8,
     pub size: usize,
     /// Written by the target.
-    pub write: AtomicUsize,
+    pub write: critical_section::Mutex<RefCell<usize>>,
     /// Written by the host.
-    pub read: AtomicUsize,
+    pub read: critical_section::Mutex<RefCell<usize>>,
     /// Channel properties.
     ///
     /// Currently, only the lowest 2 bits are used to set the channel mode (see constants below).
-    pub flags: AtomicUsize,
+    pub flags: critical_section::Mutex<RefCell<usize>>,
 }
 
 impl Channel {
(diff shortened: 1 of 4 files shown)


Urhengulas:
While it seems to work to replace `AtomicT` with `critical_section::Mutex<Cell<T>>`, we are not confident that there are no failure cases left that we didn't think about.

firmware/defmt-rtt/src/channel.rs
```diff
@@ -13,13 +10,13 @@ pub(crate) struct Channel {
     pub buffer: *mut u8,
     pub size: usize,
     /// Written by the target.
-    pub write: AtomicUsize,
+    pub write: critical_section::Mutex<RefCell<usize>>,
```
TheButlah, line 13: Is there a benefit to the mutex approach instead of using the `atomic-polyfill` crate from `embassy`?
If I understand correctly, that crate will use the native atomics when preset, and polyfill atomics using `critical-section` when native ones aren't available

Urhengulas: That might indeed be the better option.

Urhengulas: See https://github.com/knurling-rs/defmt/pull/702

Issue #693, opened 2022-08-27
Linker error on ESP32-C3: .text is not a multiple of alignment


TheButlah:
```
❯ cargo build
    Blocking waiting for file lock on package cache
    Blocking waiting for file lock on package cache
    Blocking waiting for file lock on package cache
   Compiling esp32c3-hal v0.1.0
   Compiling firmware v0.0.0 (/Users/ryan.butler/P/slimevr/rust/firmware)
warning: unused imports: `debug`, `info`
  --> src/main.rs:19:13
   |
19 | use defmt::{debug, error, info};
   |             ^^^^^         ^^^^
   |
   = note: `#[warn(unused_imports)]` on by default

error: linking with `rust-lld` failed: exit status: 1
  |
[...]

PR #702, opened 2022-10-05
Atomic polyfill

Urhengulas:
This PR replaces `core::sync::atomic::*` with `atomic_polyfill::*`.

This change makes `defmt-rtt` available on targets without native atomics.
Fixes https://github.com/knurling-rs/defmt/issues/597.

There should be no negative impact on targets with atomic support, since in these cases the native atomics will be used.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -7,6 +7,7 @@ and this project adheres to [Semantic Versioning](http://semver.org/).
 
 ## [Unreleased]
 
+- [#xxx]: `defmt-rtt`: Switch to `atomic-polyfill` for better target support
 - [#701]: Pre-relase cleanup
 - [#695]: `defmt-rtt`: Refactor rtt [3/2]
 - [#689]: `defmt-rtt`: Update to critical-section 1.0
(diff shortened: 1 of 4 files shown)


Urhengulas:
We decided to pause this change until we do a coordinated effort to get risc-v support for `defmt` and `probe-run`. All the relevant issues will be tracked with the tag `topic: risc-v`.

TheButlah:
Thats unfortunate given that the change is a minor one. I'll continue to use this code in a fork I guess

Urhengulas:
> Thats unfortunate given that the change is a minor one. I'll continue to use this code in a fork and patch my dependencies, I guess

Is `defmt` + `defmt-rtt` really usable on `esp32c3` with this change? Isn't `esp32c3` support anyways blocked on https://github.com/knurling-rs/defmt/issues/693? Not even talking about probe-run which supports risc-v even less.

TheButlah:
Yes, this PR fully fixes `demt` on the esp32c3 when compiled for `-imc`. I have been successfully using defmt for a while, albeit on the `-imac` target (I had to change because the atomic trap handler was broken). I have switched to `-imc` due to performance reasons  and because of a bug in the atomic trap handler (the esp32c3 doesn't natively support atomics)

The linker script works fine for `direct-boot` mode, which is what most people using `esp-hal` use anyway. Pretty sure it would work for `booloader` mode too, but no one bothered to try it.

PR #706, opened 2022-11-07
Satisfy Clippy

Urhengulas:


firmware/qemu/src/bin/dbg.rs
```diff
@@ -20,9 +21,11 @@ fn main() -> ! {
     let _: (i32, i32) = dbg!(x - 2, x + 2);
 
     // dbg! with zero arguments
+    #[allow(clippy::let_unit_value)]
     let _: () = dbg!();
```
jonas-schievink, line 25: This line could instead be `let () = dbg!();`, which is a bit cleaner and might also silence the warning

firmware/qemu/src/bin/dbg.rs
```diff
@@ -20,9 +21,11 @@ fn main() -> ! {
...
 
     // dbg! with trailing comma
+    #[allow(clippy::double_parens)]
```
jonas-schievink, line 28: I think this is a bug in clippy: https://github.com/rust-lang/rust-clippy/issues/9000

firmware/qemu/src/bin/dbg.rs
Urhengulas: Kk. But it sadly does not silence the warning.

jonas-schievink:
CI fails due to https://github.com/la10736/rstest/issues/169 – I'm not sure why our nightly runs are blocking merges though, we might not want to do that

Urhengulas:
> CI fails due to [la10736/rstest#169](https://github.com/la10736/rstest/issues/169) – I'm not sure why our nightly runs are blocking merges though, we might not want to do that

You are right, they shouldn't

Urhengulas:
bors r=jonas-schievink

PR #707, opened 2022-11-20
Use atomic_polyfill on targets without atomics

TheButlah:
Most targets, even those without atomic instructions, support atomic load and stores. RISC-V is one of the few platforms that lacks this ability, and rustc is planning on solving that in the future (https://github.com/rust-lang/rust/issues/99668).

However, in the meantime it is impossible to use `defmt-rtt` on the riscv non-atomic platform, such as the ESP32-C3.

`atomic-polyfill` is a useful crate that emulates the atomics via `critical_section`, *but only on platforms without atomic support*. On platforms with atomic support, it just re-exports the types from `core::sync::atomic`.

Would this PR be considered for inclusion in `defmt-rtt`? A prior PR was rejected here: https://github.com/knurling-rs/defmt/pull/702

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -7,6 +7,7 @@ and this project adheres to [Semantic Versioning](http://semver.org/).
 
 ## [Unreleased]
 
+- [#707]: `defmt-rtt`: Use `atomic-polyfill`
 - [#704]: Release `defmt-macros 0.3.3`, `defmt-print 0.3.3`, `defmt-rtt 0.4.0`
 - [#703]: `defmt-print`: Update to `clap 4.0`.
 - [#701]: Pre-relase cleanup
--- a/firmware/defmt-rtt/Cargo.toml
+++ b/firmware/defmt-rtt/Cargo.toml
@@ -11,6 +11,6 @@ repository = "https://github.com/knurling-rs/defmt"
 version = "0.4.0"
 
 [dependencies]
+atomic-polyfill = "1"
 defmt = { version = "0.3", path = "../../defmt" }
 critical-section = "1.1"
-atomic-polyfill = "1"
--- a/firmware/defmt-rtt/Cargo.toml
+++ b/firmware/defmt-rtt/Cargo.toml
@@ -13,3 +13,4 @@ version = "0.4.0"
 [dependencies]
 defmt = { version = "0.3", path = "../../defmt" }
 critical-section = "1.1"
+atomic-polyfill = "1"
--- a/firmware/defmt-rtt/src/channel.rs
+++ b/firmware/defmt-rtt/src/channel.rs
@@ -1,7 +1,5 @@
-use core::{
-    ptr,
-    sync::atomic::{AtomicUsize, Ordering},
-};
+use atomic_polyfill::{AtomicUsize, Ordering};
+use core::ptr;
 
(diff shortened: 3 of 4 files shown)


TheButlah:
To clarify: things have changed since then, chiefly these work now:
https://github.com/esp-rs/esp-hal/issues/191
https://github.com/esp-rs/esp-hal/issues/175

It works on -imac, but this PR will allow it to work on -imc also (preferred because it's not using a trap handler anymore)

Urhengulas:
I will try  to come back to you about this next year

Dirbaio:
`defmt-rtt` only uses atomic load/store, no CAS, so using `atomic-poyfill` shouldn't be necessary. It'll acquire/release the critical section unnecessarily for each load/store.

Since all accesses are done within a critical section already, I think they could be replaced with UnsafeCell + write/read_volatile + fences.

Urhengulas:
> `defmt-rtt` only uses atomic load/store, no CAS, so using `atomic-poyfill` shouldn't be necessary. It'll acquire/release the critical section unnecessarily for each load/store.
> 
> Since all accesses are done within a critical section already, I think they could be replaced with UnsafeCell + write/read_volatile + fences.

Thank you @Dirbaio. Could you elaborate how "UnsafeCell + write/read_volatile + fences" would roughly look like?

Dirbaio:
Using `UnsafeCell<u32>` instead of `AtomicU32`, and do volatile loads/stores to it. It's sound because these are always done within the critical section.

It's an ugly workaround though. Ideally Rust should expose load/store atomics on riscv32-imc (what the esp32c3 is) like it does in cortex-m. See relevant Rust issue: https://github.com/rust-lang/rust/issues/99668

Urhengulas:
@TheButlah do you have bandwidth to change your PR to `UnsafeCell` as described by @adamgreig?

TheButlah:
Is it really a good idea to add unsafe code when non-CAS atomics work fine?

I think the issue was more of a lack of knowledge about the relevant `-C` flags on my end. Ideally users could be educated that the following flags should be used on RISCV:

`.cargo/config.toml`
```
[target.riscv32imc-unknown-none-elf]
rustflags = [
  # enable the atomic codegen option for RISCV
  "-C", "target-feature=+a",

  # tell the core library have atomics even though it's not specified in the target definition.
  # NOTE: This is fine for load/store, but we don't support compare-and-swap, so we don't get
  # full atomic features. For that, atomic_polyfill or the atomic trap handler should be used
  # instead. But we don't need CAS in the firmware.
  "--cfg", "target_has_atomic_load_store",
  "--cfg", 'target_has_atomic_load_store="8"',
  "--cfg", 'target_has_atomic_load_store="16"',
  "--cfg", 'target_has_atomic_load_store="32"',
  "--cfg", 'target_has_atomic_load_store="ptr"',
]
```

Urhengulas:
Sooo ... there aren't changes to defmt necessary? I am slightly unsure about the status.

Dirbaio:
they're not, I think this can be closed.

It'll Just Work when https://github.com/rust-lang/rust/issues/99668 gets fixed for riscv. Until then users can use [these](https://github.com/knurling-rs/defmt/pull/707#issuecomment-1488035240) flags.

PR #725, opened 2023-01-20
Replace chrono with time to avoid pulling in old, vulnerable versions of time

Yatekii:


decoder/src/log/json_logger.rs
```diff
@@ -25,7 +25,9 @@ impl Log for JsonLogger {
             // defmt goes to stdout, since it's the primary output produced by this tool.
             let mut sink = io::stdout().lock();
 
-            let host_timestamp = Utc::now().timestamp_nanos();
+            let host_timestamp = OffsetDateTime::now_utc()
+                .unix_timestamp_nanos()
+                .max(i64::MAX as i128) as i64;
```
Urhengulas, line 30: Why do we do this comparison here?

Urhengulas (changes requested):
Hi @Yatekii, The PR looks good 😀 I only have one question regarding the `.max`.

decoder/src/log/json_logger.rs
Yatekii: This clamps the value to i64::MAX so it does not wrap and do weird stuff if the i128 value is too big :)

Urhengulas: I guess I am just confused, but shouldn't it be `.min` then?

If I understand correctly, you want `i64::MAX` as the upper bound. But now, `host_timestamp` will always be `i64::MAX` if the timestamp is smaller than `i64::MAX`.

But we, if I understand correctly, want to always have the actual timestamp, except when the timestamp is bigger than `i64::MAX`, then we want to cap it at `i64::MAX`.

Yatekii: Yeah, I am just being an idiot here, sorry. Will fix.

Urhengulas (approved):
Thank you!

bors r+

PR #726, opened 2023-01-20
Replace unmaintained difference with dissimilar

Yatekii:


Urhengulas (approved):
Thank you Noah!

bors r+

PR #733, opened 2023-03-05
defmt: add impl for core::net

newAM:
This adds formatting for `core::net` types.  `core::net` is currently unstable, so this is feature gated.

### DIFF
--- a/defmt/src/impls/core_/net.rs
+++ b/defmt/src/impls/core_/net.rs
@@ -1,0 +1,72 @@
+use core::net;
+
+use super::*;
+
+impl Format for net::AddrParseError {
+    fn format(&self, fmt: Formatter) {
+        crate::write!(fmt, "AddrParseError(_)");
+    }
+}
+
+impl Format for net::Ipv4Addr {
+    fn format(&self, fmt: Formatter) {
+        let [a, b, c, d] = self.octets();
+        crate::write!(fmt, "{}.{}.{}.{}", a, b, c, d);
+    }
+}
+
+impl Format for net::SocketAddrV4 {
+    fn format(&self, fmt: Formatter) {
+        crate::write!(fmt, "{}:{}", self.ip(), self.port());
+    }
+}
+
+impl Format for net::Ipv6Addr {
+    fn format(&self, fmt: Formatter) {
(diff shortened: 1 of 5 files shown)


defmt/src/impls/core_/net.rs
```diff
@@ -0,0 +1,69 @@
...
+            self.octets()[1],
+            self.octets()[2],
+            self.octets()[3],
+            self.octets()[4],
+            self.octets()[5],
+            self.octets()[6],
+            self.octets()[7]
+        );
+    }
+}
```
Urhengulas, line 45: `Ipv6Addr::octets(&self)` returns a `[u8; 16]` ([docs](https://doc.rust-lang.org/stable/std/net/struct.Ipv6Addr.html#method.octets)). Are you purposefully just using the first eight?

defmt/src/impls/core_/net.rs
```diff
@@ -0,0 +1,69 @@
...
+impl Format for net::Ipv4Addr {
+    fn format(&self, fmt: Formatter) {
+        crate::write!(
+            fmt,
+            "{}.{}.{}.{}",
+            self.octets()[0],
+            self.octets()[1],
+            self.octets()[2],
+            self.octets()[3]
+        );
```
Urhengulas, line 20: You _could_ also do the following. You don't have to; I just find it pretty.

```suggestion
        let [a, b, c, d] = ip.octets();
        crate::write!(fmt, "{}.{}.{}.{}", a, b, c, d);
```

With `println!` you could even do following, but unfortunately we didn't implement that (yet):

```rust
let [a, b, c, d] = ip.octets();
println!("{a}.{b}.{c}.{d}");
```

Urhengulas (changes requested):
One small question and one suggestion; otherwise, it looks good.

defmt/src/impls/core_/net.rs
newAM: Done!

defmt/src/impls/core_/net.rs
newAM: Whoops, that's a mistake on my part, good catch, thanks!

Urhengulas:
Thank you!

bors r+

PR #740, opened 2023-03-10
Snapshot tests for `core::net`

Urhengulas:
This PR adds snapshot tests for the implements added in #733. While doing that, I also clean up a bit.


newAM (approved):
Looks good to me!  It seems the failures are just from a toolchain mismatch since `ip_in_core` requires nightly.

Urhengulas:
> Looks good to me! It seems the failures are just from a toolchain mismatch since `ip_in_core` requires nightly.

True, they only should be run on nightly.

PR #743, opened 2023-03-20
`defmt-parser`: Simplify tests with `rstest` 

Urhengulas:
This PR uses the `rstest` crate to drastically reduce the amount of testing code while keeping the same tests. Additionally, there is now more granular error reporting.

### DIFF
--- a/parser/src/lib.rs
+++ b/parser/src/lib.rs
@@ -540,7 +540,7 @@ mod tests {
     use rstest::rstest;
 
     #[rstest]
-    #[case::no__param("", None, Type::Format, None)]
+    #[case::noo_param("", None, Type::Format, None)]
     #[case::one_param_type("=u8", None, Type::U8, None)]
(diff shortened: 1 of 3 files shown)


parser/src/lib.rs
```diff
@@ -537,207 +537,55 @@ pub fn parse(format_string: &str, mode: ParserMode) -> Result<Vec<Fragment<'_>>,
...
-            parse_param("1", ParserMode::Strict),
-            Ok(Param {
-                index: Some(1),
-                ty: Type::Format,
-                hint: None,
-            })
+    use rstest::rstest;
+
+    #[rstest]
+    #[case::no__param("", None, Type::Format, None)]
```
BriocheBerlin, line 543: This gives a warning `function `case_1_no__param` should have a snake case name`.

Urhengulas: Damn 😑

Urhengulas: It's solved now

BriocheBerlin (approved):
Looks good!

PR #746, opened 2023-03-29
Update to `bitflags v2`

Urhengulas:
This is a breaking change since users might need to adapt their bitflag definitions.


japaric (approved):

jonathanpallant:
We should save any breaking changes for defmt 1.0, because when we bump the major version, *everyone* in the ecosystem has to move across because you cannot have a defmt vX and a defmt vY dependency in the same tree.

Urhengulas:
This PR is blocked until defmt v1.

jonathanpallant:
defmt v1 will not contain breaking changes. So I'm closing this.

PR #742, opened 2023-03-16
Adds `crate_name` to `Symbol`

BriocheBerlin:
Fixes https://github.com/knurling-rs/defmt/issues/532.

When running e.g. `$ arm-none-eabi-readelf -Ws target/thumbv7em-none-eabihf/debug/hello | rg crate_name` on the `my-app app-template`, it returns:
```
$ arm-none-eabi-readelf -Ws target/thumbv7em-none-eabihf/debug/hello | rg crate_name
165: 00000002     1 OBJECT  GLOBAL DEFAULT    8 {"package":"my-app","tag":"defmt_println","data":"Hello, world!","disambiguator":"1780402918813241646","crate_name":"hello"}
171: 00000001     1 OBJECT  GLOBAL DEFAULT    8 {"package":"defmt","tag":"defmt_prim","data":"{=__internal_Display}","disambiguator":"1668500705691287208","crate_name":"defmt"}
177: 00000003     1 OBJECT  GLOBAL DEFAULT    8 {"package":"panic-probe","tag":"defmt_error","data":"{}","disambiguator":"9448177036453248414","crate_name":"panic_probe"}
```

PR #747, opened 2023-04-04
Bump wire version

therealfrauholle:
The changes introduced in https://github.com/knurling-rs/defmt/pull/742 add the field `crate_name` to the wire format. As far as I understand this requires a bump of the defmt version.

Also, is this a reason to yank version 0.3.5? This problem caused me quite some headache today and I hope this spares others from going through the same pain.


Urhengulas:
Thank you. You are right, this should have happened before the release and I will yank it.

bors r+

Issue #749, opened 2023-04-05
Make bumping the wire format an API breaking change


therealfrauholle:
Hey @Urhengulas after the latest release of the changes in https://github.com/knurling-rs/defmt/pull/747 I am still having problems aligning versions in my setup :sweat:

It appears to me that we missed bumping the dependency `defmt-macros` of `defmt` in https://github.com/knurling-rs/defmt/commit/37892413e3a64797bdc213d6a644754ac8bb2eef and also earlier in https://github.com/knurling-rs/defmt/commit/1cbbb334bf508d13c1d537623fe2a4165dfc5d65, it should now depend on `defmt-macros-0.3.4`.

Following this I tried to wrap my head around how the wire format plays together with the SemVer guarantees, and I think that bumping the wire format is an API breaking change.

This is related to the changes introduced as of https://github.com/knurling-rs/defmt/pull/540, which solved https://github.com/knurling-rs/defmt/issues/287.

Dirbaio:
The original rationale for decoupling the wire format version from the defmt semver version was that doing a defmt semver-major bump is a giant pain. You can't mix defmt versions in a binary (#426), and this seems fundamentally unfixable with the defmt design. So upgrading defmt requires ALL libraries using it to update. And in turn, that update requires a major bump of the library, because it'll break binaries that are still using the older version.

This happened in the defmt 0.2 -> 0.3 upgrade. It was a giant pain, it took MONTHS.

Decoupling the versions allows defmt to evolve while avoiding regularly inflicting this pain into the ecosystem. I really don't think we should walk back on this decision.

However, I remember discussing making defmt-decoder support multiple wire format versions, which it seems it never happened. Currently old defmt-decoder/probe-run versions support only wire `3`, newer ones support only `4`, so you have to keep switching between them if some projects have been updated and some other haven't. 

If the new versions added support for `4` instead of dropping support for `3` the experience would be much better: you upgrade defmt, you get an error from probe-run saying "please update probe-run", and then everything would be working again.

Dirbaio:
It'd be nicer to have a single defmt-decoder version support multiple wire format versions. If it's one-to-one bins ca nhack it by depending on multiple defmt-decoder versions at once, yes, but that work needs to be repeated for each project that uses defmt-decoder, and needs manual work to add new versions, vs just `cargo update`ing to a newer defmt-decoder.

datdenkikniet:
[...]

Issue #412, opened 2021-02-24
defmt-print: misleading error message on defmt version mismatch


japaric:
the error says to install **probe-run** instead of `defmt-print`

japaric:
``` console
$ defmt-print -e target/thumbv7em-none-eabihf/debug/hello
Error: defmt version mismatch: firmware is using 0.2, `probe-run` supports 0.1
suggestion: `cargo install` a different non-git version of `probe-run` that supports defmt 0.2
```

Urhengulas:
I was hoping to get the name of the binary from a cargo env variable during compile time, but none of `CARGO_PKG_NAME`, `CARGO_CRATE_NAME`, `CARGO_BIN_NAME`, `CARGO_PRIMARY_PACKAGE` seems to yield the name of the binary, but only the name of the library or nothing.

jonas-schievink:
We can probably solve this by using a strongly-typed error type in `defmt_decoder` that has a variant that indicates a version mismatch. Then the calling binary can print its own message.

Urhengulas:
> We can probably solve this by using a strongly-typed error type in `defmt_decoder` that has a variant that indicates a version mismatch. Then the calling binary can print its own message.

Yes, I also though about this, but then I tried to be smart 😬 😆

PR #750, opened 2023-04-12
Add support for decoding wire format version 3

jannic:
As suggested in https://github.com/knurling-rs/defmt/issues/749#issuecomment-1500863605, it would be nice if defmt-decoder could handle older wire format versions transparently.

This patch implements support for version 3. A version of `probe-run` compiled against this was verified to work with defmt versions 0.3.2, 0.3.3 and 0.3.4.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -9,7 +9,7 @@ and this project adheres to [Semantic Versioning](http://semver.org/).
 
 - [#750]: Add support for decoding wire format version 3
 
-[#750]: https://github.com/knurling-rs/defmt/pull/748
+[#750]: https://github.com/knurling-rs/defmt/pull/750
 
 ## [v0.3.4] - 2023-04-05
 
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -7,7 +7,9 @@ and this project adheres to [Semantic Versioning](http://semver.org/).
 
 ## [Unreleased]
 
-- Add support for decoding wire format version 3
+- [#750]: Add support for decoding wire format version 3
+
+[#750]: https://github.com/knurling-rs/defmt/pull/748
 
 ## [v0.3.4] - 2023-04-05
 
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -7,6 +7,8 @@ and this project adheres to [Semantic Versioning](http://semver.org/).
 
 ## [Unreleased]
 
+- Add support for decoding wire format version 3
+
 ## [v0.3.4] - 2023-04-05
 
 - [#748]: Release `defmt-v0.3.4`, `defmt-decoder-v0.3.6`, `defmt-print-v0.3.4` and yank previous
--- a/decoder/src/elf2table/mod.rs
+++ b/decoder/src/elf2table/mod.rs
@@ -13,7 +13,7 @@ use std::{
     path::{Path, PathBuf},
 };
 
-use crate::{BitflagsKey, StringEntry, Table, TableEntry, Tag, DEFMT_VERSION, DEFMT_VERSION_3};
+use crate::{BitflagsKey, StringEntry, Table, TableEntry, Tag, DEFMT_VERSIONS};
 use anyhow::{anyhow, bail, ensure};
 use object::{Object, ObjectSection, ObjectSymbol};
 
@@ -218,10 +218,10 @@ pub fn parse_impl(elf: &[u8], check_version: bool) -> Result<Option<Table>, anyh
 
 /// Checks if the version encoded in the symbol table is compatible with this version of the `decoder` crate
 fn check_version(version: &str) -> Result<(), String> {
-    if version != DEFMT_VERSION && version != DEFMT_VERSION_3 {
+    if !DEFMT_VERSIONS.contains(&version) {
         let msg = format!(
             "defmt wire format version mismatch: firmware is using {}, `probe-run` supports {}\nsuggestion: use a newer version of `defmt` or `cargo install` a different version of `probe-run` that supports defmt {}",
-            version, DEFMT_VERSION, version
+            version, DEFMT_VERSIONS.join(", "), version
         );
 
         return Err(msg);
--- a/decoder/src/elf2table/mod.rs
+++ b/decoder/src/elf2table/mod.rs
@@ -13,7 +13,7 @@ use std::{
     path::{Path, PathBuf},
 };
 
-use crate::{BitflagsKey, StringEntry, Table, TableEntry, Tag, DEFMT_VERSION};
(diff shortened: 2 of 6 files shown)


datdenkikniet (commented):
`defmt-print` prints the versions it supports, but this is currently only `defmt_decoder::DEFMT_VERSION`. Would be good to add `DEFMT_VERSION_3` as wel

jannic:
> `defmt-print` prints the versions it supports, but this is currently only `defmt_decoder::DEFMT_VERSION`. Would be good to add `DEFMT_VERSION_3` as well

I thought about it but decided against it to keep the fix minimal.
A proper solution would scale better to more protocol versions (eg. by parsing the number and allowing a range?)

Perhaps the message should just be changed from `supported defmt version: 4` to `latest supported defmt version: 4`?

datdenkikniet:
That is fair and makes sense. Just some indication that it supports more than just the latest version may be nice

decoder/src/lib.rs
```diff
@@ -10,6 +10,7 @@
 #![cfg_attr(docsrs, doc(cfg(unstable)))]
 #![doc(html_logo_url = "https://knurling.ferrous-systems.com/knurling_logo_light_text.svg")]
 
+pub const DEFMT_VERSION_3: &str = "3";
 pub const DEFMT_VERSION: &str = "4";
```
jonathanpallant, line 14: I think I would do:

```suggestion
pub static DEFMT_VERSIONS: &[&'static str] = ["3", "4"];
```

jonathanpallant:
LGTM, but I propose we wait for @japaric or @Urhengulas to return for this one.

decoder/src/lib.rs
jannic: I intentionally didn't include any change that could be viewed as a kind of design decision regarding the way to support multiple protocol versions.
That doesn't mean that I'd be opposed to such a change. I just wanted to leave such decisions to the Ferrous developers. I won't be able to work on this for the next few hours. Feel free to add any changes you like to the pull request.

jannic: @jonathanpallant, would you prefer this variant?
https://github.com/jannic/defmt/commit/b133f771449a8bc050281ae68f499e16deee9189

jonathanpallant: I would, but let's ask Jorge and Johann.

Urhengulas (changes requested):

parser/src/display_hint.rs
```diff
@@ -67,7 +67,15 @@ impl DisplayHint {
...
-                        crate_name: crate_name.into(),
+                        crate_name: Some(crate_name.into()),
+                    });
+                }
+                [bitflags_name, package, disambiguator] => {
+                    return Some(DisplayHint::Bitflags {
+                        name: bitflags_name.into(),
+                        package: package.into(),
+                        disambiguator: disambiguator.into(),
+                        crate_name: None,
```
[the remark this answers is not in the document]
jannic: This variant became a little more complicated due to the fact that `crate_name` needs to be an `Optional`.
I chose to structure it a little bit different to simplify the code. Do you like this variant?
```
            if parts.len() < 3 || parts.len() > 4 {
                return Some(DisplayHint::Unknown(s.into()));
            }
            return Some(DisplayHint::Bitflags {
                name: parts[0].into(),
                package: parts[1].into(),
                disambiguator: parts[2].into(),
                crate_name: parts.get(3).map(|&s| s.to_string()),
            });
```

The part `|| parts.len() > 4` could be omitted, so the code would ignore additional components. That way, it would be possible to add further information in the future, without a breaking change.
However I'm not sure if such an extension is likely, or if it's better to have a stricter parser.

decoder/src/elf2table/mod.rs
```diff
@@ -218,10 +218,10 @@ pub fn parse_impl(elf: &[u8], check_version: bool) -> Result<Option<Table>, anyh
 
 /// Checks if the version encoded in the symbol table is compatible with this version of the `decoder` crate
 fn check_version(version: &str) -> Result<(), String> {
-    if version != DEFMT_VERSION {
+    if !DEFMT_VERSIONS.contains(&version) {
         let msg = format!(
             "defmt wire format version mismatch: firmware is using {}, `probe-run` supports {}\nsuggestion: use a newer version of `defmt` or `cargo install` a different version of `probe-run` that supports defmt {}",
```
jannic, line 223: Just a side note: This function is not only used by `probe-run` but also by other binaries. That can become a little bit confusing if `cargo embed` calls itself `probe-run`: https://github.com/rp-rs/rp2040-project-template/issues/57#issuecomment-1530889493
Any idea for a better wording?

Urhengulas: yeah, this is a bit annoying. we are tracking it here: https://github.com/knurling-rs/defmt/issues/412

While it is ugly, we could just add a second parameter which is the application name.
```rs
fn check_version(version: &str, app_name: &str) -> Result<(), String>
```
```suggestion
            "defmt wire format version mismatch: firmware is using {}, `{app_name}` supports {}\nsuggestion: use a newer version of `defmt` or use a different version of `{app_name}` that supports defmt {}",
```

CHANGELOG.md
```diff
@@ -7,6 +7,8 @@ and this project adheres to [Semantic Versioning](http://semver.org/).
 
 ## [Unreleased]
 
+- Add support for decoding wire format version 3
```
Urhengulas, line 10:
```suggestion
- [#750]: Add support for decoding wire format version 3

[#750]: https://github.com/knurling-rs/defmt/pull/748
```

CHANGELOG.md
```diff
@@ -7,6 +7,10 @@ and this project adheres to [Semantic Versioning](http://semver.org/).
 
 ## [Unreleased]
 
+- [#750]: Add support for decoding wire format version 3
+
+[#750]: https://github.com/knurling-rs/defmt/pull/748
```
Urhengulas, line 12:
```suggestion
[#750]: https://github.com/knurling-rs/defmt/pull/750
```

decoder/src/lib.rs
```diff
@@ -10,7 +10,10 @@
 #![cfg_attr(docsrs, doc(cfg(unstable)))]
 #![doc(html_logo_url = "https://knurling.ferrous-systems.com/knurling_logo_light_text.svg")]
 
-pub const DEFMT_VERSION: &str = "4";
+pub const DEFMT_VERSIONS: &[&str] = &["3", "4"];
+// To avoid a breaking change, still provide `DEFMT_VERSION`.
+#[deprecated = "Please use DEFMT_VERSIONS instead"]
```
Urhengulas, line 15: I wasn't really aware of the deprecated attribute. That's neat!

Urhengulas (approved):
Thank you!

bors r+

PR #753, opened 2023-05-05
add defmt impls for `core::ptr::NonNull` and `fn(...) -> Ret`

### LINKED ISSUE #752: derive support for C-like fn pointers and `core::ptr::NonNull`
When generating a Rust binding to a C struct containing function pointers, `bindgen` outputs type definitions that look roughly like the following

```Rust
pub type c_library_callback_t = ::core::option::Option<
    unsafe extern "C" fn(
        param1: Param1Type,
    ),
>;

#[repr(C)]
#[derive(Debug)]
pub struct c_library_api {
    pub callback: c_library_callback_t 
}
```

with the bindgen `ParseCallbacks::add_derives` interface, I can add `defmt::Format` to all of the bindings without issue, however I currently need to manually make exceptions for any type that includes C function pointers and any struct including those structs.

The std Debug impl is just [printing the function pointer](https://play.rust-lang.org/?version=stable&mode=debug&edition=2021&gist=03a6eb1cfd8b0aa145b4cdf0f7ca73fa). Don't see any reason why defmt shouldn't do exactly the same
image

On a strongly related topic, `core::ptr::NonNull` is a fairly common need when wrapping a generated C-API (at least, from what I've written wrappers for) and doesn't currently have defmt support

Crzyrndm:
Fixes #752

NonNull and fn pointers formatted like a raw pointer. This matches the [behaviour of `Debug`](https://play.rust-lang.org/?version=stable&mode=debug&edition=2021&gist=9ef1d891fe9ae31b46512ded1e5ff5fc) for these types

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -7,6 +7,10 @@ and this project adheres to [Semantic Versioning](http://semver.org/).
 
 ## [Unreleased]
 
+- [#753]: Add `defmt::Format` impls for `core::ptr::NonNull` and `fn(Args...) -> Ret` (up to 12 arguments)
+
+[#753]: https://github.com/knurling-rs/defmt/pull/753
+
 ## [v0.3.5] - 2023-05-05
 
 - [#xxx]: Release `defmt-decoder v0.3.7`, `defmt-macros v0.3.5`, `defmt-parser v0.3.3`, `defmt-print v0.3.7`
--- a/defmt/src/impls/core_/mod.rs
+++ b/defmt/src/impls/core_/mod.rs
@@ -12,6 +12,7 @@ mod cell;
 mod net;
 mod num;
 mod ops;
+mod ptr;
 mod slice;
 
 use super::*;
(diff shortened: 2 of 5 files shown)


Urhengulas (commented):
Thank you for your contribution @Crzyrndm.

> Note: not quite sure where to add tests for the output?

Please add some log statements just before following line line and run `cargo xtask test-snapshot --overwrite` from the repository root: https://github.com/knurling-rs/defmt/blob/d58633f8c89adf02e3ed7f25fad1c0f0f24dbc7d/firmware/qemu/src/bin/log.rs#L735

Please also add this PR to the `CHANGELOG.md`.

Crzyrndm:
Test prints added for NonNull, and fn pointers of 0/1/5 parameters (formatting makes 12 a bit much, it's already noisy enough as is and the limit is about as arbitrary as it gets).

Urhengulas:
Thank you!

bors r+

PR #760, opened 2023-06-14
`defmt-macros`: Upgrade to syn 2

mkroening:
This performs the upgrade to syn 2.

I did this in two steps:

1. I upgraded to syn 2 by changing as little as possible about the code, while still passing all UI tests.
2. I migrated `get_defmt_format_option` to the new `parse_nested_meta` method. This makes the code much more readable while keeping the same behavior.

CC: @Urhengulas, I saw you started working on this on the [`syn2`](https://github.com/knurling-rs/defmt/tree/syn2) branch, after I already started on this.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -7,11 +7,13 @@ and this project adheres to [Semantic Versioning](http://semver.org/).
 
 ## [Unreleased]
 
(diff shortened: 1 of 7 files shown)


Urhengulas (approved):
Thank you @mkroening. LGTM!

PR #764, opened 2023-07-07
Use mermaid for "defmt ecosystem" diagram

Urhengulas:
## What
This PR removes the previous (static) diagram of the defmt ecosystem from the README and replaces it with a mermaid code block.

### Context
Mermaid is a JavaScript based diagramming and charting tool that renders Markdown-inspired text definitions to create and modify diagrams dynamically. It is widely used. By now both GitHub and crates.io do render mermaid diagrams, in GitHub you can even interact with them (zooming in and out mostly, but still).

## Why
This change will make it easier to keep the diagram up to date, since we don't rely on other tooling (draw.io in our case) to modify the diagram.

## Drawbacks
As a grain of salt I have to say that I actually do like the aesthetics of the previous diagram more (see below). But I still think that the advantages outweigh that. It is possible to style the mermaid diagram with CSS, but I did not want to go down that road.

### Before
defmt

### After
mermaid-diagram-2023-07-07-142929


Urhengulas:
> I will do a pre-release on crates.io to see if it renders there as well.

Unfortunately it fails to render. See https://crates.io/crates/defmt/0.3.6-beta.0

crates.io shows a warning popup with "Failed to render mermaid diagram." and in the console it prints "WARN: Could not find the language 'mermaid', did you forget to load/include a language module?".

It could be because I am not using the default renderer.

Urhengulas:
> It could be because I am not using the default renderer.

Nope. Changing back to the default renderer does not change anything.

tshepang:
> As a grain of salt I have to say that I actually do like the aesthetics of the previous diagram more (see below).

hm, I find the mermaid one (at least as rendered here) more pretty

README.md
```diff
@@ -29,7 +29,40 @@ It still might work on older rust versions, but this isn't ensured.
...
+
+    subgraph Legend
+    O[[user facing crate]]
+    P[internal crate]
+    Q(unreleased crate)
+    O -- dependency --> P
+    Q -- pinned dependency --o P
+    end
+
+    J ~~~ Legend
```
tshepang, line 64: should this be here... what does it mean

Urhengulas: It is an invisible link, which is useful to change where the legend is placed. But it is actually not invisible, because the new renderer apparently does ignore quite a few things.

Urhengulas: https://mermaid.js.org/syntax/flowchart.html#an-invisible-link

tshepang: you may get lucky by asking at https://github.com/orgs/mermaid-js/discussions

Urhengulas:
We need to wait for https://github.com/rust-lang/crates.io/issues/6786 to be resolved

Urhengulas:
> We need to wait for [rust-lang/crates.io#6786](https://github.com/rust-lang/crates.io/issues/6786) to be resolved

The bug is fixed and does render on crates.io now! 🎉

README.md
```diff
@@ -29,7 +29,40 @@ It still might work on older rust versions, but this isn't ensured.
 
 The following diagram illustrates the user-facing and internal crates of the defmt framework.
 
-![defmt crates structure](assets/defmt.png)
+```mermaid
+%%{ init: { 'flowchart': { 'curve': 'step', "defaultRenderer": "elk" } } }%%
```
BriocheBerlin, line 33: `"defaultRenderer": "elk"` seems to be the culprit for the `~~~` invisible link not working.

BriocheBerlin (commented):
To my mind the mermaid chart is less clear than the old one because the crossed out links were more obvious. This makes the legend being connected even more confusing.
Not using the elk renderer produces this 
Screenshot 2023-08-03 at 15 58 26
Also not great.

Urhengulas:
@jonathanpallant What is your take? Should we have the mermaid diagram or drop it?

jonathanpallant:
I ripped mermaid out of the training material because I could not get it to reliably render.

I'd prefer doing something in Visio or dot and exporting it as SVG. Reduces the number of variables.

Urhengulas:
Sure, just wanted to check with you before closing the PR.

The only problem I see with creating the diagram in another tool is, that it might be harder to edit the diagram, since not everyone has access.

Issue #602, opened 2021-10-06
`defmt::println` messages have a leading space when timestamps are disabled


japaric:
Steps to reproduce:
- start from `app-template` (`main` branch)
- configure it to use git version of `defmt` crates
- remove the `defmt::timestamp!` item from `src/lib.rs`
- modify `src/bin/hello.rs` to use `println!`

``` rust
defmt::println!("Hello, world!");
```

This outputs:

``` console
 Hello, world!
└─ hello::__cortex_m_rt_main @ src/bin/hello.rs:8
```

Ideally, it should output: (the `H` is exactly on top of the "corner pipe")

``` console
Hello, world!
└─ hello::__cortex_m_rt_main @ src/bin/hello.rs:8
```

to match the other logging macros, e.g. the `levels` example produces:

``` console
ERROR error
└─ levels::__cortex_m_rt_main @ src/bin/levels.rs:12
```

cc @justahero

jonas-schievink:
Problem lies here:

https://github.com/knurling-rs/defmt/blob/4e594035eb6b954ad171c093959ab4bc8ca47525/decoder/src/log.rs#L59

The fix is to make `timestamp` an `Option<String>`in that function and only add the space in that branch when it's `Some`.

PR #765, opened 2023-07-15
decoder: add support for customizable logger formatting

andresovela:
A pain point for me when using `defmt` is that I can't easily customize the format in which the logs are printed.

This PR replaces the current `PrettyLogger` with a new `StdoutLogger` (seemed like a better name to me since the other logger is called `JsonLogger`), that supports an optional format string. If no format string is provided, a default format is used that looks pretty much the same as the current format, with the exception that the location is not "dimmed".

The formatting options I added are completely arbitrary and not set in stone. I just chose what made sense to me. This is from what I wrote in the docs:

```
 /// An optional `log_format` string can be provided to format the way
 /// logs are printed. A format string could look as follows:
 /// "{t} [{L}] Location<{f}:{l}> {s}"
 ///
 /// The arguments between curly braces are placeholders for log metadata.
 /// The following arguments are supported:
 /// - {t} : log timestamp
 /// - {f} : file name (e.g. "main.rs")
 /// - {F} : file path (e.g. "home/app/main.rs")
 /// - {m} : module path (e.g. "foo::bar::some_function")
 /// - {l} : line number
 /// - {L} : log level (e.g. "INFO", "DEBUG", etc)
 /// - {s} : the actual log
 ///
 /// For example, with the log format shown above, a log would look like this:
 /// "23124 [INFO] Location<main.rs:23> Hello, world!"
```

The new format is easily changeable and extendable. Some things that I think could be implemented but I chose not to waste time on before getting community feedback are:

- Support for a new formatting option to set/reset colors
- Support for escaped curly braces (the parser currently does not support random curly braces in the format string)
- Support for field widths (e.g. `{t:8}` or something like that to specify that the timestamp must be at least 8 characters wide)

This PR is also an alternative to #762 (I think)

### DIFF
--- a/decoder/src/log/format.rs
+++ b/decoder/src/log/format.rs
@@ -1,11 +1,11 @@
-use nom::IResult;
+use nom::branch::alt;
 use nom::bytes::complete::{take, take_till1};
 use nom::character::complete::char;
-use nom::combinator::{map_res, map};
+use nom::combinator::{map, map_res};
 use nom::error::{FromExternalError, ParseError};
 use nom::multi::many0;
 use nom::sequence::delimited;
-use nom::branch::alt;
+use nom::IResult;
 use nom::Parser;
 
 #[derive(Debug, PartialEq, Clone)]
@@ -25,35 +25,41 @@ pub enum LogSegment {
 pub struct InvalidArgument;
 
 fn parse_argument<'a, E>(input: &'a str) -> IResult<&'a str, LogSegment, E>
-where E: ParseError<&'a str> + FromExternalError<&'a str, InvalidArgument>, {
+where
+    E: ParseError<&'a str> + FromExternalError<&'a str, InvalidArgument>,
+{
     let parse_enclosed = delimited(char('{'), take(1u32), char('}'));
-    let mut parse_type = map_res(parse_enclosed, move |s| {
-        match s {
-            "t" => Ok(LogSegment::Timestamp),
-            "f" => Ok(LogSegment::FileName),
-            "F" => Ok(LogSegment::FilePath),
-            "m" => Ok(LogSegment::ModulePath),
-            "l" => Ok(LogSegment::LineNumber),
-            "L" => Ok(LogSegment::LogLevel),
-            "s" => Ok(LogSegment::Log),
-            _ => Err(InvalidArgument),
-        }
+    let mut parse_type = map_res(parse_enclosed, move |s| match s {
+        "t" => Ok(LogSegment::Timestamp),
+        "f" => Ok(LogSegment::FileName),
+        "F" => Ok(LogSegment::FilePath),
+        "m" => Ok(LogSegment::ModulePath),
+        "l" => Ok(LogSegment::LineNumber),
+        "L" => Ok(LogSegment::LogLevel),
+        "s" => Ok(LogSegment::Log),
+        _ => Err(InvalidArgument),
     });
 
     parse_type.parse(input)
 }
 
 fn parse_string_segment<'a, E>(input: &'a str) -> IResult<&'a str, LogSegment, E>
-where E: ParseError<&'a str>, {
-    map(take_till1(|c| c == '{'),|s: &str| LogSegment::String(s.to_string())).parse(input)
+where
+    E: ParseError<&'a str>,
+{
+    map(take_till1(|c| c == '{'), |s: &str| {
+        LogSegment::String(s.to_string())
+    })
+    .parse(input)
 }
 
 pub fn parse(input: &str) -> Result<Vec<LogSegment>, String> {
     let mut parse_all = many0(alt((parse_argument::<()>, parse_string_segment)));
 
-    parse_all(input).map(|(_, output)| output).map_err(|e| e.to_string())
+    parse_all(input)
+        .map(|(_, output)| output)
+        .map_err(|e| e.to_string())
 }
-
 
 #[cfg(test)]
 mod tests {
--- a/decoder/src/log/format.rs
+++ b/decoder/src/log/format.rs
@@ -1,0 +1,110 @@
+use nom::IResult;
+use nom::bytes::complete::{take, take_till1};
+use nom::character::complete::char;
+use nom::combinator::{map_res, map};
+use nom::error::{FromExternalError, ParseError};
+use nom::multi::many0;
+use nom::sequence::delimited;
+use nom::branch::alt;
+use nom::Parser;
+
+#[derive(Debug, PartialEq, Clone)]
+#[non_exhaustive]
+pub enum LogSegment {
+    String(String),
+    Timestamp,
+    FileName,
+    FilePath,
+    ModulePath,
+    LineNumber,
+    LogLevel,
+    Log,
+}
+
+#[derive(Debug, PartialEq)]
+pub struct InvalidArgument;
+
+fn parse_argument<'a, E>(input: &'a str) -> IResult<&'a str, LogSegment, E>
+where E: ParseError<&'a str> + FromExternalError<&'a str, InvalidArgument>, {
+    let parse_enclosed = delimited(char('{'), take(1u32), char('}'));
+    let mut parse_type = map_res(parse_enclosed, move |s| {
+        match s {
+            "t" => Ok(LogSegment::Timestamp),
+            "f" => Ok(LogSegment::FileName),
+            "F" => Ok(LogSegment::FilePath),
+            "m" => Ok(LogSegment::ModulePath),
+            "l" => Ok(LogSegment::LineNumber),
+            "L" => Ok(LogSegment::LogLevel),
+            "s" => Ok(LogSegment::Log),
+            _ => Err(InvalidArgument),
+        }
+    });
+
+    parse_type.parse(input)
+}
+
+fn parse_string_segment<'a, E>(input: &'a str) -> IResult<&'a str, LogSegment, E>
+where E: ParseError<&'a str>, {
+    map(take_till1(|c| c == '{'),|s: &str| LogSegment::String(s.to_string())).parse(input)
+}
+
+pub fn parse<'a>(input: &'a str) -> Result<Vec<LogSegment>, String> {
+    let mut parse_all = many0(alt((parse_argument::<()>, parse_string_segment)));
+
+    parse_all(input).map(|(_, output)| output).map_err(|e| e.to_string())
+}
+
+
+#[cfg(test)]
+mod tests {
+    use super::*;
+
+    #[test]
+    fn test_parse_log_template() {
+        let log_template = "{t} [{L}] {s}\n└─ {m} @ {F}:{l}";
+
+        let expected_output = vec![
+            LogSegment::Timestamp,
+            LogSegment::String(" [".to_string()),
+            LogSegment::LogLevel,
+            LogSegment::String("] ".to_string()),
+            LogSegment::Log,
+            LogSegment::String("\n└─ ".to_string()),
+            LogSegment::ModulePath,
+            LogSegment::String(" @ ".to_string()),
+            LogSegment::FilePath,
+            LogSegment::String(":".to_string()),
+            LogSegment::LineNumber,
+        ];
+
+        let result = parse(log_template);
+        assert_eq!(result, Ok(expected_output));
+    }
+
+    #[test]
+    fn test_parse_string_segment() {
+        let result = parse_string_segment::<()>("Log: {t}");
+        let (input, output) = result.unwrap();
+        assert_eq!(input, "{t}");
+        assert_eq!(output, LogSegment::String("Log: ".to_string()));
+    }
+
+    #[test]
+    fn test_parse_empty_string_segment() {
+        let result = parse_string_segment::<()>("");
+        assert!(result.is_err());
+    }
+
+    #[test]
+    fn test_parse_timestamp_argument() {
+        let result = parse_argument::<()>("{t}");
+        assert_eq!(result, Ok(("", LogSegment::Timestamp)));
+    }
+
+    #[test]
+    fn test_parse_invalid_argument() {
+        let result = parse_argument::<()>("{foo}");
+        let result = dbg!(result);
+        assert_eq!(result, Result::Err(nom::Err::Error(())));
+    }
+}
--- a/decoder/src/log/format.rs
+++ b/decoder/src/log/format.rs
@@ -1,60 +1,51 @@
-use nom::branch::alt;
-use nom::bytes::complete::{take, take_till1};
-use nom::character::complete::char;
-use nom::combinator::{map, map_res};
-use nom::error::{FromExternalError, ParseError};
-use nom::multi::many0;
-use nom::sequence::delimited;
-use nom::IResult;
-use nom::Parser;
+use nom::{
+    branch::alt,
+    bytes::complete::{take, take_till1},
+    character::complete::char,
+    combinator::{map, map_res},
+    multi::many0,
+    sequence::delimited,
+    IResult, Parser,
+};
 
 #[derive(Debug, PartialEq, Clone)]
 #[non_exhaustive]
-pub enum LogSegment {
-    String(String),
-    Timestamp,
+pub(super) enum LogSegment {
     FileName,
     FilePath,
-    ModulePath,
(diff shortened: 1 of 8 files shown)


Urhengulas:
Thank you for the PR. I am looking into it

decoder/src/log/format.rs
```diff
@@ -0,0 +1,115 @@
...
+        _ => Err(InvalidArgument),
+    });
+
+    parse_type.parse(input)
+}
+
+fn parse_string_segment<'a, E>(input: &'a str) -> IResult<&'a str, LogSegment, E>
+where
+    E: ParseError<&'a str>,
+{
```
Urhengulas, line 49: Could the error type just be `()`? I don't see a call-site where you are using a different concrete type. Applies to this and the other functions.

```suggestion
fn parse_string_segment<'a, E>(input: &'a str) -> IResult<&'a str, LogSegment, ()> {
```

decoder/src/log/format.rs
```diff
@@ -0,0 +1,115 @@
+use nom::branch::alt;
+use nom::bytes::complete::{take, take_till1};
+use nom::character::complete::char;
+use nom::combinator::{map, map_res};
+use nom::error::{FromExternalError, ParseError};
+use nom::multi::many0;
+use nom::sequence::delimited;
+use nom::IResult;
+use nom::Parser;
```
Urhengulas, line 9: Please merge the imports, like

```rs
use nom::{
    branch::alt,
    bytes::complete::{take, take_till1}
    // ...
}
```

decoder/src/log/format.rs
```diff
@@ -0,0 +1,115 @@
...
+pub enum LogSegment {
+    String(String),
+    Timestamp,
+    FileName,
+    FilePath,
+    ModulePath,
+    LineNumber,
+    LogLevel,
+    Log,
+}
```
Urhengulas, line 22: Why does it need to be `pub`? Isn't it just used for the parser?

Urhengulas: Also please sort the variants alphabetically.

decoder/src/log/stdout_logger.rs
```diff
@@ -0,0 +1,310 @@
...
+
+use super::format;
+use super::{format::LogSegment, DefmtRecord};
+
+enum Record<'a> {
+    Defmt(&'a DefmtRecord<'a>),
+    Host(&'a LogRecord<'a>),
+}
+
+pub(crate) struct StdoutLogger {
```
Urhengulas, line 20: I agree that `StdoutLogger` is a better term. Since the renaming of the file and struct messes up the diff and makes reviewing a bit harder, I will change the name in a separate PR and ask you to please rebase your PR on that.

decoder/src/log/stdout_logger.rs
```diff
@@ -0,0 +1,310 @@
...
+
+use std::{
+    fmt::Write as _,
+    io::{self, StderrLock, StdoutLock},
+    path::Path,
+    sync::atomic::{AtomicUsize, Ordering},
+};
+
+use super::format;
+use super::{format::LogSegment, DefmtRecord};
```
Urhengulas, line 13:
```suggestion
use super::{
    format::{self, LogSegment},
    DefmtRecord,
};
```

Urhengulas (changes requested):
I do like the idea!

Such a mechanism is common practice in many other logging systems, so I don't know why we did not think about it before.

Left a few comments.

decoder/src/log/stdout_logger.rs
```diff
@@ -0,0 +1,310 @@
...
+    }
+
+    fn print_log_level<W: io::Write>(&self, sink: &mut W) -> io::Result<()> {
+        let level = match self.record {
+            Record::Defmt(record) => record.level(),
+            Record::Host(record) => Some(record.level()),
+        };
+
+        let level = if let Some(level) = level {
+            // TODO: Should the color be customizable via the format too?
```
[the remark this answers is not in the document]
andresovela: Yeah I agree, I feel like those extra options are very bikesheddable

andresovela:
I made the requested changes and I added an extra `host_log_format` option. It made sense to implement it after reading some of the issues in `probe-run`.

decoder/src/log/stdout_logger.rs
andresovela: Done

Urhengulas (changes requested):
I played around with the feature and really like it. It gives users all the flexibility they need.

A few topics we need to look into:
- timestamp
    - You are correctly not adding a placeholder in case there is no timestamp. But, after this PR there is one space in front of each log message if the default format is used but there is no timestamp.
        ```console
        $ # before
        $ # after
        $  INFO info
          └─ levels::__cortex_m_rt_main @ src/bin/levels.rs:10
           TRACE trace
          └─ levels::__cortex_m_rt_main @ src/bin/levels.rs:11
           WARN warn
          └─ levels::__cortex_m_rt_main @ src/bin/levels.rs:12
           DEBUG debug
          └─ levels::__cortex_m_rt_main @ src/bin/levels.rs:13
           ERROR error
          └─ levels::__cortex_m_rt_main @ src/bin/levels.rs:14
           <lvl> println
          └─ levels::__cortex_m_rt_main @ src/bin/levels.rs:15
        ````
- println

Urhengulas (changes requested):
I played around with the feature and really like it. It gives users all the flexibility they need.

I've noticed a few details which changed with your PR. Some of these changes might be okay, but I wanted to make sure none of them happen accidentally, and some plainly need to be fixed.

<details>

<summary>Please have a look at these probe-run invocations as context</summary>

```console
$ # before PR without timestamp
$ DEFMT_LOG=trace cargo run -q --bin levels
(HOST) INFO  flashing program (2 pages / 8.00 KiB)
(HOST) INFO  success!
INFO  info
└─ levels::__cortex_m_rt_main @ src/bin/levels.rs:10
TRACE trace
└─ levels::__cortex_m_rt_main @ src/bin/levels.rs:11
WARN  warn
└─ levels::__cortex_m_rt_main @ src/bin/levels.rs:12
DEBUG debug
└─ levels::__cortex_m_rt_main @ src/bin/levels.rs:13
ERROR error
└─ levels::__cortex_m_rt_main @ src/bin/levels.rs:14
println
└─ levels::__cortex_m_rt_main @ src/bin/levels.rs:15
(HOST) INFO  device halted without error

$ # before PR with timestamp
$ DEFMT_LOG=trace cargo run -q --bin levels
0 INFO  info
└─ levels::__cortex_m_rt_main @ src/bin/levels.rs:10
1 TRACE trace
└─ levels::__cortex_m_rt_main @ src/bin/levels.rs:11
2 WARN  warn
└─ levels::__cortex_m_rt_main @ src/bin/levels.rs:12
3 DEBUG debug
└─ levels::__cortex_m_rt_main @ src/bin/levels.rs:13
4 ERROR error
└─ levels::__cortex_m_rt_main @ src/bin/levels.rs:14
5 println
└─ levels::__cortex_m_rt_main @ src/bin/levels.rs:15

$ # after PR without timestamp
$ DEFMT_LOG=trace cargo run -q --bin levels
(HOST) INFO flashing program (2 pages / 8.00 KiB)
(HOST) INFO success!
 INFO info
└─ levels::__cortex_m_rt_main @ src/bin/levels.rs:10
 TRACE trace
└─ levels::__cortex_m_rt_main @ src/bin/levels.rs:11
 WARN warn
└─ levels::__cortex_m_rt_main @ src/bin/levels.rs:12
 DEBUG debug
└─ levels::__cortex_m_rt_main @ src/bin/levels.rs:13
 ERROR error
└─ levels::__cortex_m_rt_main @ src/bin/levels.rs:14
 <lvl> println
└─ levels::__cortex_m_rt_main @ src/bin/levels.rs:15
(HOST) INFO device halted without error

$ # after PR with timestamp
$ DEFMT_LOG=trace cargo run -q --bin levels
0 INFO info
└─ levels::__cortex_m_rt_main @ src/bin/levels.rs:10
1 TRACE trace
└─ levels::__cortex_m_rt_main @ src/bin/levels.rs:11
2 WARN warn
└─ levels::__cortex_m_rt_main @ src/bin/levels.rs:12
3 DEBUG debug
└─ levels::__cortex_m_rt_main @ src/bin/levels.rs:13
4 ERROR error
└─ levels::__cortex_m_rt_main @ src/bin/levels.rs:14
5 <lvl> println
└─ levels::__cortex_m_rt_main @ src/bin/levels.rs:15
```

</details>

## timestamp

With your PR and the default format, if there is no timestamp, the first character is a space. Before this was not the case[^1], but the log line just started with the log level. This should be fixed.

[^1]: Actually there was a similar bug https://github.com/knurling-rs/defmt/issues/602

## println

The idea of `defmt::println` is that the log message is printed without the log level and timestamp. With your PR it emits a `<lvl>` as log level. This should be fixed.

## bold messages

With your PR all messages are bold. Before only log messages were bold, but println and host messages were normal. This is okay if it changes, but I would probably err for making all log messages normal (and the metadata dimmed).

## space between timestamp and log message

Before, there were enough spaces between the log level and the log message so the beginning of the log message was aligned. With the PR there is always just one space. I think this is okay if it changes.

andresovela:
> With your PR and the default format, if there is no timestamp, the first character is a space. Before this was not the case, but the log line just started with the log level. This should be fixed.

Unless I'm misunderstanding the circumstances under which the timestamp would not be available, I'd argue that the previous behavior is not the right behavior in the context of the new formatting system in this PR.

I'm assuming that a timestamp is not available if the user does not provide a `defmt::timestamp!()` implementation. If the format specifies that a timestamp should be printed, I think printing `0` illustrates the issue better than just not printing it.

I also just don't see how I can implement the previous behavior nicely in the current system. If the user does not intend to provide a `defmt::timestamp!()` implementation, I think the right call would be to guide the user to remove the timestamp from the log format.
For example if I want the following format:
```
T1234 | INFO | Hello
```
I think it'd be better to print
```
T0 | INFO | Hello
```
than
```
T| INFO | Hello
```

andresovela:
> The idea of `defmt::println` is that the log message is printed without the log level and timestamp. With your PR it emits a `<lvl>` as log level. This should be fixed.

I've fixed this. I think I hadn't understood the purpose of that `println` function before. I actually think that whole thing should be revamped, and we should instead have `info_raw!()`, `debug_raw!()`, ... macros, that print logs without formatting but preserve log level information for filtering purposes. 

I see the value for a `println!()` macro that always prints regardless of the configured log level, but I still think these logs should have format. I think this is a subject for future discussion and should be addressed in a future PR if there's a desire for such a thing.

andresovela:
> With your PR all messages are bold. Before only log messages were bold, but println and host messages were normal. This is okay if it changes, but I would probably err for making all log messages normal (and the metadata dimmed).

I've fixed this, with the exception of the metadata dimmed. I think further formatting options should be added in future PRs.

andresovela:
> Before, there were enough spaces between the log level and the log message so the beginning of the log message was aligned. With the PR there is always just one space. I think this is okay if it changes.

I didn't understand this comment, but I think we should extend the formatting options with a field width specifier in a future PR. I think that would address this in a better way.

Then you could have something like `"{L} {f:25} {s}"` that would print something like:
```
INFO main.rs                   Hello, world!
INFO decoder.rs                Decoding...
```
without the need to special-case everything internally and allowing for better user-customization.

andresovela:
> space between timestamp and log message

Ah I see what you mean now, but I don't understand why this is not working. The width specifier is [already there](https://github.com/knurling-rs/defmt/pull/765/files#diff-56895c1ad12fe97972cecd2f61bdeab9a13daa1d1ed77d83691b60705e25b2a8R198). Any ideas?

andresovela:
I figured it out.

Urhengulas:
> Unless I'm misunderstanding the circumstances under which the timestamp would not be available, I'd argue that the previous behavior is not the right behavior in the context of the new formatting system in this PR.
> 
> I'm assuming that a timestamp is not available if the user does not provide a `defmt::timestamp!()` implementation. If the format specifies that a timestamp should be printed, I think printing `0` illustrates the issue better than just not printing it.

Or we have a `<time>` default.

> I also just don't see how I can implement the previous behavior nicely in the current system. If the user does not intend to provide a `defmt::timestamp!()` implementation, I think the right call would be to guide the user to remove the timestamp from the log format.

I agree. I am unsure if we easily know if there is a timestamp implementation or not.

Urhengulas:
> I am unsure if we easily know if there is a timestamp implementation or not.

It seems that adding a `defmt_decoder::Table::has_timestamp()` method does the job:

```diff
diff --git a/decoder/src/lib.rs b/decoder/src/lib.rs
index cf7fa12..7da00bf 100644
--- a/decoder/src/lib.rs
+++ b/decoder/src/lib.rs
@@ -271,6 +271,10 @@ impl Table {
     pub fn encoding(&self) -> Encoding {
         self.encoding
     }
+
+    pub fn has_timestamp(&self) -> bool {
+        self.timestamp.is_some()
+    }
 }
 
 // NOTE follows `parser::Type`
```

I testet it with a small project and adding and removing the timestamp implementation.

andresovela:
So what would you like me to do about this?

Urhengulas:
> So what would you like me to do about this?

I propose the following:

| timestamp implementation | timestamp in log_format | action |
| :---: | :---: | :--- |
| yes | yes | just display it |
| yes | no | just not display it |
| no | yes | warn that there is no timestamp implementation and that it should be added or the timestamp should be removed from the log_format |
| no | no | just not display it |

Do you agree?

andresovela:
> It seems that adding a `defmt_decoder::Table::has_timestamp()` method does the job

I really tried to make this work but I kept running into issues. Adding that method means that I have to add the warning in `probe-run`, in a place that has no knowledge of what the logger is at all, much less what its format contains (see [here](https://github.com/knurling-rs/probe-run/blob/4cf82a95eb0b031adb338768a23dbc3bd9545ddf/src/elf.rs#L117)).

I also can't use this function directly in `defmt_decoder` because this crate has no instances of `Table` anywhere.

How do you picture this working out?

andresovela:
I added an info struct to the defmt logger API, and I added the warning in `probe-run`

Urhengulas:
> I agree with the warning, but what do I print with the logs in that case? `""`? `" "`? `"0"`?

`<time>`?

andresovela:
Done

Urhengulas (approved):

ia0:
My understanding is that with this PR, the println format is hard-coded and it is not possible to add the timestamp as it used to be the case (IIRC). Is this by design or it would be possible to let the user customize the println format too? (like for the log and host log format)

andresovela:
Hi @ia0,

According to @Urhengulas this is by design:

> ## println
> The idea of `defmt::println` is that the log message is printed without the log level and timestamp. With your PR it emits a `<lvl>` as log level. This should be fixed.

I also argued that a discussion should be had about what we want `println` to do:

> I've fixed this. I think I hadn't understood the purpose of that `println` function before. I actually think that whole thing should be revamped, and we should instead have `info_raw!()`, `debug_raw!()`, ... macros, that print logs without formatting but preserve log level information for filtering purposes.
> 
> I see the value for a `println!()` macro that always prints regardless of the configured log level, but I still think these logs should have format. I think this is a subject for future discussion and should be addressed in a future PR if there's a desire for such a thing.

ia0:
Thanks for the response! I'm working around by adding the timestamp directly in the `println` call. However, if the goal is for `println` to not show any timestamp information, then target timestamp should not be sent for those messages. When I use `probe-run --json` I can see the `target_timestamp` field being populated on `println` calls.

Urhengulas:
> Thanks for the response! I'm working around by adding the timestamp directly in the `println` call. However, if the goal is for `println` to not show any timestamp information, then target timestamp should not be sent for those messages. When I use `probe-run --json` I can see the `target_timestamp` field being populated on `println` calls.

Hei 👋🏾 

I agree that we need to make the logging story more clear, I am gonna put some time into that soon™️

I would be interested why you cannot use the logging macros (e.g. `info`, `debug`) if you want the timestamp?

ia0:
> I would be interested why you cannot use the logging macros (e.g. `info`, `debug`) if you want the timestamp?

Because that's something I want always printed, regardless of log level.

PR #767, opened 2023-07-25
add serialport support to defmt-print

brandonros:
I got the idea from https://github.com/gauteh/defmt-serial/blob/main/example-pi-pico/pico-elf2uf2-defmt-wrapper.sh except that didn't work for me on Mac OS X Ventura with a FT232RL USB device

Instead of trying to really get into the weeds and debug why that was doing nothing, this seemed easier.

Obviously we'd probably want this behind some feature flag to make this optional (or not at all)

Just figured this might help somebody developing on RP2040 who doesn't have a 2nd RP2040 laying around + doesn't feel like soldering wires to be able to hit the SWD RTT debugger thingie

### DIFF
--- a/print/src/main.rs
+++ b/print/src/main.rs
@@ -82,15 +82,11 @@ impl Source {
                 let n = stdin.read(buf)?;
                 Ok((n, n == 0))
             }
-            Source::Serial(port) => {
-                match port.read(buf) {
-                    Ok(n) => {
-                        Ok((n, n == 0))
-                    },
-                    Err(ref e) if e.kind() == std::io::ErrorKind::TimedOut => Ok((0, false)),
-                    Err(e) => Err(anyhow!(e)),
-                }
-            }
+            Source::Serial(port) => match port.read(buf) {
+                Ok(n) => Ok((n, n == 0)),
+                Err(ref e) if e.kind() == std::io::ErrorKind::TimedOut => Ok((0, false)),
+                Err(e) => Err(anyhow!(e)),
+            },
             Source::Tcp(tcpstream) => Ok((tcpstream.read(buf)?, false)),
         }
     }
@@ -143,7 +139,11 @@ fn main() -> anyhow::Result<()> {
 
     let mut source = match command {
         None | Some(Command::Stdin) => Source::stdin(),
-        Some(Command::Serial { path }) => Source::Serial(serialport::new(path, 115_200).timeout(Duration::from_millis(10)).open()?),
+        Some(Command::Serial { path }) => Source::Serial(
+            serialport::new(path, 115_200)
+                .timeout(Duration::from_millis(10))
+                .open()?,
+        ),
         Some(Command::Tcp { host, port }) => Source::tcp(host, port)?,
     };
 
--- a/print/src/main.rs
+++ b/print/src/main.rs
@@ -35,7 +35,7 @@ struct Opts {
 
 #[derive(Subcommand)]
 enum Command {
-    /// Read defmt frames from stdin (the default)
+    /// Read defmt frames from stdin (default)
     Stdin,
     /// Read defmt frames from serial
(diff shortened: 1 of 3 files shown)


print/src/main.rs
```diff
@@ -114,6 +130,7 @@ fn main() -> anyhow::Result<()> {
 
     let mut source = match command {
         None | Some(Command::Stdin) => Source::stdin(),
+        Some(Command::Serial { path }) => Source::Serial(serialport::new(path, 115_200).timeout(std::time::Duration::from_millis(10)).open().unwrap()),
```
Urhengulas, line 133: Can you please move this into `Source::serial(path)`?
```suggestion
        Some(Command::Serial { path }) => Source::Serial(serialport::new(path, 115_200).timeout(Duration::from_millis(10)).open()?),
```

Urhengulas (changes requested):
Hi @brandonros, thank you for your PR.

I'd be happy to add the serial support. Just a few questions:
1. Can you please share an example of how you use the feature? Just want to understand it better.
2. The library you are introducing, `serialport`, seems to look for new maintainers. I am a bit hesitant to add a library which has no active maintainers. Do you know if the found someone already?

brandonros:
> Hi @brandonros, thank you for your PR.
> 
> I'd be happy to add the serial support. Just a few questions:
> 
> 1. Can you please share an example of how you use the feature? Just want to understand it better.
> 2. The library you are introducing, `serialport`, seems to look for new maintainers. I am a bit hesitant to add a library which has no active maintainers. Do you know if the found someone already?

If you are working on a RP2040 and you don't want to solder/attach a 2nd USB thing to handle `defmt` logs over SWD, you can do 

## .cargo/config.toml

```toml
[build]
target = "thumbv6m-none-eabi"

[target.thumbv6m-none-eabi]
runner = "bash runner.sh"
rustflags = [
  "-C", "link-arg=--nmagic",
  "-C", "link-arg=-Tlink.x",
  "-C", "link-arg=-Tlink-rp.x",
  "-C", "link-arg=-Tdefmt.x",
]

[env]
DEFMT_LOG = "debug"
```

## runner.sh

```shell
#!/bin/bash

set -e

# flash
elf2uf2-rs -d "$1"

# debug uart
defmt-print -e "$1" serial --path /dev/tty.usbserial-A50285BI
```

combined with https://github.com/gauteh/defmt-serial to achieve "compile, flash, get logs" all in one command + one device/solution (no external SWD debugger)

brandonros:
image

I don't think `CI / lint` is failing due to something specifically related to this PR (other than adding a new serialport dependency)

Urhengulas:
> image
> 
> I don't think `CI / lint` is failing due to something specifically related to this PR (other than adding a new serialport dependency)

It fails because of the introduced dependency `serialport`. The relevant part of the error is:
>  could not find system library 'libudev' required by the 'libudev-sys' crate

By adding `serialport` you introduce following dependencies:
```
└── serialport v4.2.2
    ├── bitflags v1.3.2
    ├── cfg-if v1.0.0
    ├── libudev v0.3.0
    │   ├── libc v0.2.147
    │   └── libudev-sys v0.1.4
    │       └── libc v0.2.147
    │       [build-dependencies]
    │       └── pkg-config v0.3.27
    ├── nix v0.26.2
    │   ├── bitflags v1.3.2
    │   ├── cfg-if v1.0.0
    │   ├── libc v0.2.147
    │   └── static_assertions v1.1.0
    └── scopeguard v1.2.0
```

You'd need to install `libudev` in the `lint` job in the CI. Similar to here: https://github.com/knurling-rs/defmt/blob/ca161bf0ab9ea8209fa5b6781e9f4e87f592eb57/.github/workflows/ci.yml#L32

But still, I do not want to add an unmaintained dependency. Are there any alternative crates?

PR #769, opened 2023-08-01
decoder: add support for color, style, width and alignment to format

andresovela:
This PR extends the functionality of the `--log-format` and `--host-log-format` options implemented in #765 by adding support for color, style, width and alignment options.

The format options are pretty much as described in [here](https://github.com/knurling-rs/defmt/pull/765#discussion_r1265257684):

> I can easily imagine assigning static colors, maybe like this "{t:black:bold} [{L:green:dimmed}] Location<{f:yellow:italic}:{l:#5F2A2A:normal}> {s:white:underline}". Note that I A) am using the terms from the colored crate, B) not particularly like that format, it is just what I came up with right now and C) would pick :black:normal as a default.

An example of what the new format could look like is:

```
{t:>8} [{L:severity:bold}] {f:white:underline}:{l:white:3} {s:werror}
```

The `#5F2A2A` variant for colors is not yet supported in this PR - I think the currently implemented options are more than enough.

Something else I still want to implement is support for nested format options or something similar that would allow users to align logs like this:

```
main.rs:15          This is a log
decoder.rs:345      This is another log
```

With the current formatting options this is not possible. I think something like `{{f}:{l}:20:white:underline}` would be a nice syntax to support formatting groups of segments at once.  I'll likely wait for this PR to be merged before working on that though.

One open question is how to implement support for styling `String` log segments, i.e. how do we allow the user to have something like `(HOST) [{L}] {s}` where the `(HOST)` part could be blue, dimmed, etc. It's probably not very important and I can't think of many use cases for this.

### DIFF
--- a/decoder/src/log/format.rs
+++ b/decoder/src/log/format.rs
@@ -53,7 +53,7 @@ pub(super) struct LogSegment {
 pub(super) struct LogFormat {
     pub(super) width: Option<usize>,
     pub(super) color: Option<LogColor>,
-    pub(super) style: Option<colored::Styles>,
+    pub(super) style: Option<Vec<colored::Styles>>,
     pub(super) alignment: Option<Alignment>,
 }
 
@@ -85,8 +85,10 @@ impl LogSegment {
     }
 
     #[cfg(test)]
-    const fn with_style(mut self, style: colored::Styles) -> Self {
-        self.format.style = Some(style);
+    fn with_style(mut self, style: colored::Styles) -> Self {
+        let mut styles = self.format.style.unwrap_or_default();
+        styles.push(style);
+        self.format.style = Some(styles);
         self
     }
 
@@ -188,16 +190,26 @@ fn parse_log_segment(input: &str) -> IResult<&str, LogSegment, ()> {
         match item {
             IntermediateOutput::Metadata(m) if metadata.is_none() => metadata = Some(m),
             IntermediateOutput::Color(c) if color.is_none() => color = Some(c),
-            IntermediateOutput::Style(s) if style.is_none() => style = Some(s),
+            IntermediateOutput::Style(s) => {
+                let mut styles: Vec<colored::Styles> = style.unwrap_or_default();
+
+                // A format with repeated style specifiers is not valid
+                if styles.contains(&s) {
+                    return Err(nom::Err::Failure(()));
+                }
+
+                styles.push(s);
+                style = Some(styles);
+            }
             IntermediateOutput::WidthAndAlignment(w) if width_and_alignment.is_none() => {
                 width_and_alignment = Some(w)
             }
-            _ => return Err(nom::Err::Error(())),
+            _ => return Err(nom::Err::Failure(())),
         }
     }
 
     let Some(metadata) = metadata else {
-        return Err(nom::Err::Error(()));
+        return Err(nom::Err::Failure(()));
     };
 
     let (width, alignment) = width_and_alignment
@@ -380,5 +392,26 @@ mod tests {
 
         let result = parse(log_template);
         assert_eq!(result, Ok(expected_output));
+    }
+
+    #[test]
+    fn test_parse_log_with_multiple_different_styles() {
+        let log_template = "{s:bold:underline:italic:dimmed}";
+
+        let expected_output = vec![LogSegment::new(LogMetadata::Log)
+            .with_style(colored::Styles::Bold)
+            .with_style(colored::Styles::Underline)
+            .with_style(colored::Styles::Italic)
+            .with_style(colored::Styles::Dimmed)];
+
+        let result = parse(log_template);
+        assert_eq!(result, Ok(expected_output));
+    }
+
+    #[test]
+    fn test_parse_log_with_repeated_styles() {
+        let log_template = "{s:bold:bold}";
+        let result = parse(log_template);
+        assert!(result.is_err());
     }
 }
--- a/decoder/src/log/format.rs
+++ b/decoder/src/log/format.rs
@@ -8,6 +8,8 @@ use nom::{
     IResult, Parser,
 };
 
+use std::str::FromStr;
+
 #[derive(Debug, PartialEq, Clone)]
 #[non_exhaustive]
 pub(super) enum LogMetadata {
@@ -23,22 +25,8 @@ pub(super) enum LogMetadata {
 
 #[derive(Debug, PartialEq, Clone, Copy)]
 pub(super) enum LogColor {
-    Black,
-    Red,
-    Green,
-    Yellow,
-    Blue,
-    Magenta,
-    Cyan,
-    White,
-    BrightBlack,
-    BrightRed,
-    BrightGreen,
-    BrightYellow,
-    BrightBlue,
-    BrightMagenta,
-    BrightCyan,
-    BrightWhite,
+    /// User-defined color
+    Color(colored::Color),
 
     /// Color matching the default color for the log level
     SeverityLevel,
@@ -46,15 +34,6 @@ pub(super) enum LogColor {
     /// Color matching the default color for the log level,
     /// but only if the log level is WARN or ERROR
     WarnError,
-}
-
-#[derive(Debug, PartialEq, Clone, Copy)]
-pub(super) enum LogStyle {
-    Bold,
-    Italic,
-    Underline,
-    Strikethrough,
-    Dimmed,
 }
 
 #[derive(Debug, PartialEq, Clone, Copy)]
@@ -74,7 +53,7 @@ pub(super) struct LogSegment {
 pub(super) struct LogFormat {
     pub(super) width: Option<usize>,
     pub(super) color: Option<LogColor>,
-    pub(super) style: Option<LogStyle>,
+    pub(super) style: Option<colored::Styles>,
     pub(super) alignment: Option<Alignment>,
 }
 
@@ -83,7 +62,7 @@ enum IntermediateOutput {
     Metadata(LogMetadata),
     WidthAndAlignment((usize, Option<Alignment>)),
     Color(LogColor),
-    Style(LogStyle),
+    Style(colored::Styles),
 }
 
 impl LogSegment {
@@ -106,7 +85,7 @@ impl LogSegment {
     }
 
     #[allow(dead_code)]
-    const fn with_style(mut self, style: LogStyle) -> Self {
+    const fn with_style(mut self, style: colored::Styles) -> Self {
         self.format.style = Some(style);
         self
     }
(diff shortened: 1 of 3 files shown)


Urhengulas (changes requested):
Thank you for your PR. I like it and will play around a bit more with it and see how it behaves.

One thing I am wondering is, if we can rely a bit more on the functionality from the `colored` crate to make the code a bit more simple on our side. E.g. we could use the `FromStr` implementation from `colored::Color` to simplify the parsing code. Or we could use `colored::Styles` instead of our own `LogStyle`.
I would be against exposing `colored` types in our public API, but as long as they are only internal I think it makes sense to use them and not hand-roll everything by ourselves.

decoder/src/log/format.rs
```diff
@@ -21,25 +21,252 @@ pub(super) enum LogSegment {
...
+    BrightCyan,
+    BrightWhite,
+
+    /// Color matching the default color for the log level
+    SeverityLevel,
+
+    /// Color matching the default color for the log level,
+    /// but only if the log level is WARN or ERROR
+    WarnError,
+}
```
Urhengulas, line 49: Maybe `LogColor`, could be simplified to following? I didn't think it fully through, but I would be happy if it works out.

```suggestion
pub(super) enum LogColor {
    Color(colored::Color),

    /// Color matching the default color for the log level
    SeverityLevel,

    /// Color matching the default color for the log level,
    /// but only if the log level is WARN or ERROR
    WarnError,
}
```

decoder/src/log/format.rs
```diff
@@ -21,25 +23,213 @@ pub(super) enum LogSegment {
...
+            format: LogFormat {
+                color: None,
+                style: None,
+                width: None,
+                alignment: None,
+            },
+        }
+    }
+
+    #[allow(dead_code)]
```
Urhengulas, line 81:
```suggestion
    #[cfg(test)]
```

decoder/src/log/format.rs
```diff
@@ -21,25 +23,213 @@ pub(super) enum LogSegment {
...
+    const fn with_color(mut self, color: LogColor) -> Self {
+        self.format.color = Some(color);
+        self
+    }
+
+    #[allow(dead_code)]
```
Urhengulas, line 87:
```suggestion
    #[cfg(test)]
```

decoder/src/log/format.rs
```diff
@@ -21,25 +23,213 @@ pub(super) enum LogSegment {
...
+    const fn with_style(mut self, style: colored::Styles) -> Self {
+        self.format.style = Some(style);
+        self
+    }
+
+    #[allow(dead_code)]
```
Urhengulas, line 93:
```suggestion
    #[cfg(test)]
```

decoder/src/log/format.rs
```diff
@@ -21,25 +23,213 @@ pub(super) enum LogSegment {
...
+    const fn with_width(mut self, width: usize) -> Self {
+        self.format.width = Some(width);
+        self
+    }
+
+    #[allow(dead_code)]
```
Urhengulas, line 99:
```suggestion
    #[cfg(test)]
```

decoder/src/log/format.rs
```diff
@@ -21,25 +23,213 @@ pub(super) enum LogSegment {
...
+    let mut parse_type = map_res(take_while(char::is_alphabetic), move |s| match s {
+        "f" => Ok(IntermediateOutput::Metadata(LogMetadata::FileName)),
+        "F" => Ok(IntermediateOutput::Metadata(LogMetadata::FilePath)),
+        "l" => Ok(IntermediateOutput::Metadata(LogMetadata::LineNumber)),
+        "s" => Ok(IntermediateOutput::Metadata(LogMetadata::Log)),
+        "L" => Ok(IntermediateOutput::Metadata(LogMetadata::LogLevel)),
+        "m" => Ok(IntermediateOutput::Metadata(LogMetadata::ModulePath)),
+        "t" => Ok(IntermediateOutput::Metadata(LogMetadata::Timestamp)),
+        _ => Err(()),
+    });
```
Urhengulas, line 116:
```suggestion
    let mut parse_type = map_res(take_while(char::is_alphabetic), move |s| {
        let meta = match s {
            "f" => LogMetadata::FileName,
            "F" => LogMetadata::FilePath,
            "l" => LogMetadata::LineNumber,
            "s" => LogMetadata::Log,
            "L" => LogMetadata::LogLevel,
            "m" => LogMetadata::ModulePath,
            "t" => LogMetadata::Timestamp,
            _ => return Err(()),
        };
        Ok(IntermediateOutput::Metadata(meta))
    });
```

decoder/src/log/format.rs
```diff
@@ -21,25 +23,213 @@ pub(super) enum LogSegment {
...
+
+    parse_type.parse(input)
+}
+
+fn parse_color(input: &str) -> IResult<&str, IntermediateOutput, ()> {
+    let mut parse_type = map_res(take_while(char::is_alphabetic), move |s| match s {
+        "severity" => Ok(IntermediateOutput::Color(LogColor::SeverityLevel)),
+        "werror" => Ok(IntermediateOutput::Color(LogColor::WarnError)),
+        s => colored::Color::from_str(s).map(|c| IntermediateOutput::Color(LogColor::Color(c))),
+    });
```
Urhengulas, line 126:
```suggestion
    let mut parse_type = map_res(take_while(char::is_alphabetic), move |s| {
        let color = match s {
            "severity" => LogColor::SeverityLevel,
            "werror" => LogColor::WarnError,
            s => match colored::Color::from_str(s) {
                Ok(c) => LogColor::Color(c),
                Err(()) => return Err(()),
            },
        };
        Ok(IntermediateOutput::Color(color))
    });
```

decoder/src/log/format.rs
```diff
@@ -21,25 +23,213 @@ pub(super) enum LogSegment {
...
+
+fn parse_style(input: &str) -> IResult<&str, IntermediateOutput, ()> {
+    let mut parse_type = map_res(take_while(char::is_alphabetic), move |s| match s {
+        "bold" => Ok(IntermediateOutput::Style(colored::Styles::Bold)),
+        "italic" => Ok(IntermediateOutput::Style(colored::Styles::Italic)),
+        "underline" => Ok(IntermediateOutput::Style(colored::Styles::Underline)),
+        "strike" => Ok(IntermediateOutput::Style(colored::Styles::Strikethrough)),
+        "dimmed" => Ok(IntermediateOutput::Style(colored::Styles::Dimmed)),
         _ => Err(()),
     });
```
Urhengulas, line 139:
```suggestion
    let mut parse_type = map_res(take_while(char::is_alphabetic), move |s| {
        let style = match s {
            "bold" => colored::Styles::Bold,
            "italic" => colored::Styles::Italic,
            "underline" => colored::Styles::Underline,
            "strike" => colored::Styles::Strikethrough,
            "dimmed" => colored::Styles::Dimmed,
            _ => return Err(()),
        };
        Ok(IntermediateOutput::Style(style))
    });
```

decoder/src/log/format.rs
```diff
@@ -21,25 +23,213 @@ pub(super) enum LogSegment {
...
+                    return Err(nom::Err::Error(()));
+                }
+                style = Some(s);
+            }
+            IntermediateOutput::WidthAndAlignment(w) => {
+                if width_and_alignment.is_some() {
+                    return Err(nom::Err::Error(()));
+                }
+                width_and_alignment = Some(w);
+            }
```
Urhengulas, line 200:
```suggestion
            IntermediateOutput::Metadata(m) if metadata.is_none() => metadata = Some(m),
            IntermediateOutput::Color(c) if color.is_none() => color = Some(c),
            IntermediateOutput::Style(s) if style.is_none() => style = Some(s),
            IntermediateOutput::WidthAndAlignment(w) if width_and_alignment.is_none() => {
                width_and_alignment = Some(w)
            }
            _ => return Err(nom::Err::Error(())),
```

decoder/src/log/stdout_logger.rs
```diff
@@ -181,36 +184,50 @@ impl<'a> Printer<'a> {
             Record::Host(_) => String::from("<time>"),
         };
 
-        write!(sink, "{timestamp:>0$}", self.min_timestamp_width,)
-    }
+        let timestamp = ColoredString::from(timestamp.as_str());
+        let timestamp = apply_color(timestamp, format.color, self.record_log_level());
+        let timestamp = apply_style(timestamp, format.style);
+        let alignment = format.alignment.unwrap_or(Alignment::Left);
+        let width = format.width.unwrap_or(self.min_timestamp_width);
```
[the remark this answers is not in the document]
Urhengulas: Or maybe they could move into `write_string` since it seems it is always called directly afterwards.

decoder/src/log/stdout_logger.rs
```diff
@@ -338,3 +391,70 @@ fn color_for_log_level(level: Level) -> Color {
...
+                    let color = color_for_log_level(level.unwrap());
+                    s.color(color)
+                } else {
+                    s
+                }
+            }
+        }
+    } else {
+        s
+    }
```
Urhengulas, line 422:
```suggestion
    match log_color {
        Some(color) => match color {
            LogColor::Color(c) => s.color(c),
            LogColor::SeverityLevel => match level {
                Some(level) => s.color(color_for_log_level(level)),
                None => s,
            },
            LogColor::WarnError => match level {
                Some(level @ (Level::Warn | Level::Error)) => s.color(color_for_log_level(level)),
                _ => s,
            },
        },
        None => s,
    }
```

decoder/src/log/stdout_logger.rs
```diff
@@ -338,3 +391,70 @@ fn color_for_log_level(level: Level) -> Color {
...
+            Styles::Strikethrough => s.strikethrough(),
+            Styles::Dimmed => s.dimmed(),
+            Styles::Clear => s.clear(),
+            Styles::Reversed => s.reversed(),
+            Styles::Blink => s.blink(),
+            Styles::Hidden => s.hidden(),
+        }
+    } else {
+        s
+    }
```
Urhengulas, line 440:
```suggestion
    match log_style {
        Some(log_style) => match log_style {
            Styles::Bold => s.bold(),
            Styles::Italic => s.italic(),
            Styles::Underline => s.underline(),
            Styles::Strikethrough => s.strikethrough(),
            Styles::Dimmed => s.dimmed(),
            Styles::Clear => s.clear(),
            Styles::Reversed => s.reversed(),
            Styles::Blink => s.blink(),
            Styles::Hidden => s.hidden(),
        },
        None => s,
    }
```

decoder/src/log/stdout_logger.rs
```diff
@@ -338,3 +391,70 @@ fn color_for_log_level(level: Level) -> Color {
...
+    match alignment {
+        Alignment::Left => {
+            write!(sink, "{s:<0$}", width)
+        }
+        Alignment::Center => {
+            write!(sink, "{s:^0$}", width)
+        }
+        Alignment::Right => {
+            write!(sink, "{s:>0$}", width)
+        }
```
Urhengulas, line 458:
```suggestion
        Alignment::Left => write!(sink, "{s:<0$}", width),
        Alignment::Center => write!(sink, "{s:^0$}", width),
        Alignment::Right => write!(sink, "{s:>0$}", width),
```

Urhengulas (changes requested):

decoder/src/log/stdout_logger.rs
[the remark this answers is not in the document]
andresovela: I refactored most of it, feel free to make more suggestions :)

andresovela:
Hi @Urhengulas, is there any news about this PR?

Urhengulas:
Hey, I've been sick and busy with other things recently. Will come back to it next week.

andresovela:
Get well soon :)

Urhengulas (changes requested):
Looks good! The happy path works nicely.

I just think the error reporting needs to be slightly improved. I found following cases, which both result in the warning and erroneous output below. I assume it is due to a parsing error.

The first case should be rejected, because it is a mistage. I think abort is fine, since the parsing of the log_format will happen at the start of the program.

The second case should be supported in the long run, but it is fine if we error for now. 

- `{t:bold:bold} [{L}] {f}:{l} {s}`
- `{t:bold:underline} [{L}] {f}:{l} {s}`
```
$ DEFMT_LOG=info cargo rb levels
    Finished dev [optimized + debuginfo] target(s) in 0.04s
     Running `/home/urhengulas/Documents/github.com/knurling-rs/krate/./probe-run --chip nRF52840_xxAA --log-format '{t:bold:bold} [{L}] {f}:{l} {s}' target/thumbv7em-none-eabihf/debug/levels`
(HOST) WARN  `defmt::timestamp!` implementation was found, but timestamp is not part of the log format; consider adding the timestamp `{t}` argument to the log format



printlnnn
(HOST) INFO  program has used at least 0.23/254.93 KiB (0.1%) of stack space
(HOST) INFO  device halted without error
```

Urhengulas:
Also I just realized that `"--host-log-format", "{t} [{L}] {f}:{l} {s}",` results in `<time> [INFO ] mod.rs:131  device halted without error` (note the `<time>`. But this should not be handled in this PR. In theory it should error, but in the long run we might want to add a host timestamp and then it should work.

andresovela:
Thanks for your feedback :)

I believe I have fixed the issues with the duplicated style specifiers. I also added some tests for that.

Urhengulas (approved):
🎉

PR #771, opened 2023-08-07
Ignore empty items in DEFMT_LOG

GrantM11235:
There are a several ways that the DEFMT_LOG environment variable could have an empty item, for example:
- `DEFMT_LOG=`
- `DEFMT_LOG=info,`
- `DEFMT_LOG=info,,app::noisy=error`

Previously, an empty item would be incorrectly parsed as a module path, causing the macro to panic.

With this patch, empty items are simply ignored.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -7,6 +7,10 @@ and this project adheres to [Semantic Versioning](http://semver.org/).
 
 ## [Unreleased]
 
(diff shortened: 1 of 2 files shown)


Urhengulas (approved):
I first was a bit unsure if we should maybe throw an explicit error in case a user accidentally adds a spare `,`, but I like the "it just works" approach more.

PR #773, opened 2023-08-17
Doc-hide macro-generated idents

mamins1376:
When running `cargo doc` on a project using defmt, I saw a lot of automatically-generated identifiers in the resulting documentation. Cargo will by default emit documentation for private items in a crate, so items like `DEFMT_LOG_STATEMENT` or `CHECK` or `check()` are apparent there too, which are clearly an implementation detail and should not be exposed to library user (documentation bloat aside).

Hopefully this tiny PR fixes that.

We love defmt, thanks for your great work :)

### DIFF
--- a/macros/src/function_like/internp.rs
+++ b/macros/src/function_like/internp.rs
@@ -16,6 +16,7 @@ pub(crate) fn expand(args: TokenStream) -> TokenStream {
         quote!({ defmt::export::fetch_add_string_index() as u16 })
     } else {
         quote!({
+            #[doc(hidden)]
             #[cfg_attr(target_os = "macos", link_section = #section_for_macos)]
             #[cfg_attr(not(target_os = "macos"), link_section = #section)]
             #[export_name = #sym_name]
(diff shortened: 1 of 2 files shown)


Urhengulas:
Hi @mamins1376, I am trying to reproduce that `cargo doc` emits documentation for the `defmt`-internals. I am trying it with a project created from our `app-template`, but don't see documentation for these items. Could you please provide me with a repository and the exact command you are using to see this? But yes, I agree that the `defmt`-internals should be hidden from the output.

mamins1376:
@Urhengulas Interesting. When tried with latest stable, I get the same results as you do but with `nightly-2023-04-12` toolchain the problem does manifest for me.

Urhengulas:
> @Urhengulas Interesting. When tried with latest stable, I get the same results as you do but with `nightly-2023-04-12` toolchain the problem does manifest for me.

That's weird. Can you please test with a more recent nightly?

jonathanpallant:
@mamins1376 did you have a chance to retry this with the latest nightly?

Urhengulas:
@mamins1376 I am gonna close the PR for now. Please re-open it if the issue still persists

PR #775, opened 2023-08-26
decoder: Ignore AArch64 mapping symbols

paulmenage:
The AArch64 ELF spec includes "mapping symbols" for identifying contiguous data/code regions in a binary:
https://github.com/ARM-software/abi-aa/blob/main/aaelf64/aaelf64.rst#mapping-symbols

Ignore these if we encounter them in the ".defmt" section rather than returning an error such as
`Error: failed to demangle defmt symbol '$d.3': expected value at line 1 column 1`

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -7,8 +7,10 @@ and this project adheres to [Semantic Versioning](http://semver.org/).
 
 ## [Unreleased]
 
+- [#775]: `defmt-decoder`: Ignore AArch64 mapping symbols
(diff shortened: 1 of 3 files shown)


Urhengulas:
Can you please share the error you encounter without the PR?

paulmenage:
I get an error like `Error: failed to demangle defmt symbol $d.3: expected value at line 1 column 1` when calling `defmt_decoder::Table::parse()` on the contents of the aarch64 binary. (Added the error to the PR description).

Urhengulas (approved):

PR #778, opened 2023-10-01
decoder: add support for nested log formatting

andresovela:
This PR is a follow-up for #769.

This adds support for nesting formats, allowing users to specify formatting for _groups_ of log segments at once.
For example, `"{[{L:bold}]%underline} {s}"` prints `"[ERROR] Something went wrong"` where `ERROR` is bold, and `[ERROR]` is underlined from `[`to `]`.

This nesting feature also enables aligning groups of log segments at once, which allows users to format their logs like this:

```
   54251 [INFO ] main.rs:543   Something happened
  400241 [ERROR] parser.rs:86  Failed to parse input 
```

I think the formatting feature is pretty much complete with this PR, and I don't have plans to extend it any further at this point.

That said, I do intend to submit a PR to propose a new default format after this is merged.

I'm looking forward to finally have this in defmt :)

### DIFF
--- a/decoder/src/log/format.rs
+++ b/decoder/src/log/format.rs
@@ -75,13 +75,14 @@
 //! The alignment can be specified to be left (`<`), right (`>`), or center-aligned (`^`).
 //! If no alignment is specified, left alignment is used by default.
 //!
-//! The minimum width is specified after the alignment
-//!
+//! The minimum width is specified after the alignment.
 //! For example, "{L} {f:>10}: {s}" is printed as follows:
 //!
 //! ```text
 //! [ERROR]    main.rs: hello
 //! ```
+//! 
+//! The log level format specifier is printed with a minimum width of 5 by default.
 //!
 //! ## Nested formatting
 //!
--- a/decoder/src/log/format.rs
+++ b/decoder/src/log/format.rs
@@ -4,8 +4,9 @@ use nom::{
     character::complete::{char, digit1, one_of},
     combinator::{map, map_res, opt},
     multi::{many0, separated_list1},
-    sequence::delimited,
-    IResult, Parser,
+    sequence::{delimited, preceded},
+    error::{Error, ErrorKind, ParseError},
+    Err, IResult, Parser,
 };
 
 use std::str::FromStr;
@@ -21,6 +22,7 @@ pub(super) enum LogMetadata {
     ModulePath,
     String(String),
     Timestamp,
+    NestedLogSegments(Vec<LogSegment>),
 }
 
 #[derive(Debug, PartialEq, Clone, Copy)]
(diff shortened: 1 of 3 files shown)


decoder/src/log/format.rs
```diff
@@ -1,48 +1,233 @@
...
+//! like this: `"{L:bold:5} {f:white:<10} {s}"`.
+//!
+//! ### Color
+//!
+//! A log segment can be specified to be colored by providing a color in the format parameters.
+//!
+//! There are three different options for coloring a log segment:
+//! - a string that can be parsed by the FromStr implementation of [colored::Color].
+//! - `severity` colors the log segment using the predefined color for the log level of log.
+//! - `werror` is similar to `severity`, but it only applies the color if the log level is WARN or ERROR.
```
Urhengulas, line 53: I am just wondering why we actually need `werror`. Isn't `severity` enough?

decoder/src/log/format.rs
```diff
@@ -1,48 +1,233 @@
+//! Formatting for `defmt` log frames.
```
Urhengulas, line 1: Thank you for writing this specification. I think that we should make it publicly accessible, which it is not atm, since the `format` module is private. Either we add it to the [defmt book](https://defmt.ferrous-systems.com), or to the `log` module. What do you think?

Urhengulas (approved):
Thank you. I have two questions, but otherwise it seems good to me.

decoder/src/log/format.rs
andresovela: I personally find that coloring warning and error logs is very useful, but coloring info, debug, and trace logs is way too noisy for me. Hence I added an option to color _just_ warnings and errors.

decoder/src/log/format.rs
andresovela: I was looking into adding this into the book too. I also realized after I opened this PR that rustdoc doesn't include this when building documentation since it's a private module.

andresovela: If you don't mind, I can do the book thing in a separate PR.

Urhengulas: Yes, that sounds good

decoder/src/log/format.rs
Urhengulas: Oki

decoder/src/log/format.rs
```diff
@@ -1,48 +1,233 @@
...
+//!
+//! Format parameters are provided by adding the parameters after the format specifier, separated by colons (`:`),
+//! like this: `"{L:bold:5} {f:white:<10} {s}"`.
+//!
+//! ### Color
+//!
+//! A log segment can be specified to be colored by providing a color in the format parameters.
+//!
+//! There are three different options for coloring a log segment:
+//! - a string that can be parsed by the FromStr implementation of [colored::Color].
```
Urhengulas, line 51:
```suggestion
//! - a string that can be parsed by the FromStr implementation of [colored::Color] (`red`, `green` etc.).
```

Issue #779, opened 2023-10-02
Add a way to specify the log level of a module in code


andresovela:
There is currently no way to specify the log level per module without passing the filter via the `DEFMT_LOG` environment variable. This approach gets unwieldy very quickly. Even the example given in the [book](https://defmt.ferrous-systems.com/filtering#modules) (which specifies only two modules) is already quite hard to visually parse:

```
DEFMT_LOG=app::important=info,app::noisy=error cargo run --bin app
```

It'd be great if we could have something like this:

```
#[defmt::level = debug]
mod app {
    fn foo() {
        // This prints because info < debug
        defmt::info!("hello");
    }
}
```

A potential crate-level attribute would also be a great addition:

```
#![defmt::global_level = debug]
```

Or something like that.

PR #782, opened 2023-10-02
Raise the default log level from error to info if DEFMT_LOG was not defined

andresovela:
I realize this might be controversial, but I've personally _**never**_ had any use for the default behavior of printing only error-level logs. Not only with `defmt`, but generally with any logger I've ever used in embedded systems.

Every time I run a project, any of them, I always have to run it with at least `DEFMT_LOG=info` or `debug`.

I think might be one of the reasons why people default to using `defmt::println`, because `defmt` prints nothing by default. Most people likely do not test `defmt` by writing `defmt::error!("hello world");`.

Note that this wouldn't be an issue if we had proper module-level filtering with something like the proposal in #779

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -7,12 +7,14 @@ and this project adheres to [Semantic Versioning](http://semver.org/).
 
 ## [Unreleased]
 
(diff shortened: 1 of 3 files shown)


andresovela:
I'm gonna give the filtering thing a go properly, so I'll close this PR.

PR #780, opened 2023-10-02
Add support for zero-padded timestamps and Unix-style timestamps

andresovela:
This PR adds a new format specifier to format the timestamp like this "01:42:56.643".

This PR also extends the parser to support parsing format parameters like `"{t:>08"}`, to zero-pad timestamps.

I also removed the `min_timestamp_width`/`timing_align` stuff that was previously needed for formatting timestamps.

PR #783, opened 2023-10-03
decoder: formatting improvements

andresovela:
Besides refactoring the log formatting code, this PR introduces a few new features:

#### `{c}` metadata specifier
Now it's possible to add `{c}` to the log format in order to print the crate name. I'm currently parsing this out of the module path, there may be better ways to do this.

#### More detailed `{f}` metadata specifiers
I found the `{f}` specifier to be not very useful in most cases, mainly because a lot of files are called `mod.rs`. Now it's possible to increase how detailed to print the path to the file by adding `{ff}`, `{fff}`, etc... to the log format. With `{ff}` the Formatter would print `format/mod.rs` instead of just `mod.rs`, which I think is much more useful.

#### Zero-padding support
I added the zero-padding functionality from #780 to this PR. Now it's possible to zero-pad metadata specifiers, such as the timestamp or the line number.

### DIFF
--- a/decoder/src/log/format/mod.rs
+++ b/decoder/src/log/format/mod.rs
@@ -473,11 +473,7 @@ impl InternalFormatter {
             let path_iter = Path::new(file).iter();
             let number_of_components = path_iter.clone().count();
 
-            let number_of_components_to_join = if number_of_components > level_of_detail as usize {
-                level_of_detail as usize
-            } else {
-                number_of_components
-            };
+            let number_of_components_to_join = number_of_components.min(level_of_detail as usize);
 
             let number_of_elements_to_skip =
                 number_of_components.saturating_sub(number_of_components_to_join);
--- a/decoder/src/log/format/mod.rs
+++ b/decoder/src/log/format/mod.rs
@@ -239,14 +239,52 @@ enum Record<'a> {
     Host(&'a LogRecord<'a>),
 }
 
+#[derive(Debug)]
 pub enum FormatterFormat<'a> {
     Default { with_location: bool },
     Custom(&'a str),
 }
 
+impl Default for FormatterFormat<'_> {
+    fn default() -> Self {
+        FormatterFormat::Default {
+            with_location: false,
+        }
+    }
+}
+
+#[derive(Debug, Default)]
 pub struct FormatterConfig<'a> {
     pub format: FormatterFormat<'a>,
     pub is_timestamp_available: bool,
+}
+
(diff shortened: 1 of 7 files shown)


Urhengulas:
I am actually a bit unsure how to go about it. Yes, the default format logic is duplicated in https://github.com/knurling-rs/probe-run/pull/425, but it is actually a bit different than in `probe-run` and also different than in this PR. Yes they check if there is a timestamp or not, but they also have a `--no-location` flag. So the code on that side would not necessarily get more simple by this PR. Also note that they are not using `log` and therefore not `defmt_decoder::log::init_logger`.

andresovela:
>  Yes they check if there is a timestamp or not, but they also have a --no-location flag.

I'd then propose to add a no-location flag to `DefmtLoggerConfig`. I don't see why not, it seems like it would be a fairly common configuration option.

> Also note that they are not using log and therefore not defmt_decoder::log::init_logger

I think all of this logic would end up in the `Formatter` introduced in #781, so I don't see this as an issue.

Urhengulas:
Makes sense

andresovela:
I did a relatively big refactor of all the formatting stuff.

Urhengulas:
> I did a relatively big refactor of all the formatting stuff.

Cool, thank you! I will hand over the project to a colleague this week. He will come back to you.

andresovela:
Hi @Urhengulas, do you know if your colleague will have time for some reviews soon?

jonathanpallant:
I'll take a look now.

jonathanpallant:
Looks good to me. It'll need a rebase to fix the CHANGELOG, and perhaps we could squash the first three commits, just for neatness.

andresovela:
Rebased now :)

andresovela:
Ping @jonathanpallant

decoder/src/log/format/mod.rs
```diff
@@ -0,0 +1,731 @@
...
+
+        let s = if let Some(file) = file {
+            let path_iter = Path::new(file).iter();
+            let number_of_components = path_iter.clone().count();
+
+            let number_of_components_to_join = if number_of_components > level_of_detail as usize {
+                level_of_detail as usize
+            } else {
+                number_of_components
+            };
```
Urhengulas, line 480:
```suggestion
            let number_of_components_to_join = number_of_components.min(level_of_detail as usize);
```

Urhengulas (approved):

PR #785, opened 2023-10-10
Add log format documentation to the book

andresovela:
This PR adds a chapter to the `defmt` book about customizing logs using the log format functionality. This PR shouldn't be merged until #783 is merged.

### DIFF
--- a/book/src/custom-log-output.md
+++ b/book/src/custom-log-output.md
@@ -175,7 +175,7 @@ This option can be passed to the printer in `.cargo/config.toml`, but due to lim
 
 runner = [
   "probe-rs",
-  "run"
+  "run",
   "--chip",
   "nRF52840_xxAA",
   "--log-format",
--- a/book/src/custom-log-output.md
+++ b/book/src/custom-log-output.md
@@ -1,0 +1,171 @@
+# Customizing the log output
+
+The way a printer outputs logs can be customized by providing a format string to `defmt`. The format string takes a set of metadata and format specifiers which can be used to include or exclude certain information when printing the logs.
+
+## Basics
+
+> The following log will be used as reference in the examples below: `defmt::debug!("hello");`
+
+ The simplest format string is `"{s}"`. This prints the log and nothing else:
+
+ ```text
+ hello
+ ```
+
+ Arbitrary text can be added to the format string, which will be printed as specified with each log.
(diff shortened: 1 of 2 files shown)


Urhengulas:
Just mention that it can be passed as `--log-format` to all three of those. And that due to a cargo config parsing restriction, `runner` needs to be an array of string instead of just a string.

book/src/custom-log-output.md
```diff
@@ -0,0 +1,186 @@
...
+
+```text
+[DEBUG] hello
+```
+
+## Metadata specifiers
+
+There are several metadata specifiers available that can be used in a format string.
+
+#### Log - `{s}`
```
jonathanpallant, line 32:
```suggestion
### Log - `{s}`
```

I think a 3-# heading is correct here?

book/src/custom-log-output.md
```diff
@@ -0,0 +1,186 @@
...
+```ignore
+// crate: my_crate
+mod foo {
+    fn bar() {
+        defmt::info!("hello");
+    }
+}
+```
+
+#### Timestamp - `{t}`
```
jonathanpallant, line 71: I think #789 means we need some more entries in here now?

book/src/custom-log-output.md
```diff
@@ -0,0 +1,186 @@
...
+The format string can be passed to `probe-rs`, `probe-run` and `defmt-print` using the `--log-format` option.
+
+This option can be passed to the printer in `.cargo/config.toml`, but due to limitations in `cargo`, the command has to be split as follows:
+
+```toml
+# .cargo/config.toml
+
+runner = [
+  "probe-rs",
+  "run"
```
jonathanpallant, line 178:
```suggestion
  "run",
```

jonathanpallant:
This looks great, thank you so much! Just a couple of minor notes.

book/src/custom-log-output.md
andresovela: I think I made these level-4 intentionally because it looks better in the book.

book/src/custom-log-output.md
andresovela: Not really, the `{t}` specifier prints whatever `defmt::timestamp!` prints

Urhengulas (approved):

PR #796, opened 2023-11-15
risc-v support for defmt-test

t-moe:
I'm trying to add risc-v support to defmt-test. I'm happy to discuss it... Thank you

TODOs:
- [x] Add risc-v semihosting support to probe-rs https://github.com/probe-rs/probe-rs/pull/1870
- [ ] Update app-template, to include `use cortex_m_rt as _;`  (as it is no longer included in defmt-test)
- [ ] Update defmt-test readme with example for risc-v

Optional:
- [ ] Add risc-v support to panic-probe
- [ ] Add option to ignore additional arguments to probe-rs (similar to how it was with probe-run) (so that rust analyzer's run test functionality is working again)
- [ ] Extend probe-rs to support passing arguments to the target via semihosting


t-moe:
Hei @japaric,
Can you look at this draft? 

To my knowledge, riscv has no such thing as an `udf` instruction, to manually trigger a hardfault.
I think it would be better to use `semihosting::process::exit` or `semihosting::process::abort` to signal test success/failure That way we can get support for a lot more platforms (see https://github.com/taiki-e/semihosting#platform-support).

Would it be ok for you, if I switch to using semihosting instead of hardfaults in panic-probe? (Currently I only use semihosting for riscv, and for cortex-m I still use `udf`). I would probably remove the `pub fn hard_fault` altogether or rename it to abort.

BriocheBerlin:
Sorry this is taking so long. Unfortunately we can't look at this before the week of December 11.

firmware/panic-probe/Cargo.toml
```diff
@@ -1,26 +1,35 @@
...
+description = "Panic handler that exits `probe-rs` with an error code"
 edition = "2021"
-keywords = ["knurling", "panic-impl", "defmt", "probe-run"]
+keywords = ["knurling", "panic-impl", "defmt", "probe-rs"]
 license = "MIT OR Apache-2.0"
 name = "panic-probe"
 readme = "README.md"
 repository = "https://github.com/knurling-rs/defmt"
-version = "0.3.1"
+version = "0.4.0"
```
jonathanpallant, line 11: I think we'll save the version bump for when we actually do a release.

firmware/panic-probe/README.md
```diff
@@ -1,10 +1,16 @@
 # `panic-probe`
 
-> Panic handler that exits [`probe-run`] with an error code
+> Panic handler that exits [`probe-rs`] with an error code using semihosting::process::abort.
 
-[`probe-run`]: https://github.com/knurling-rs/probe-run
+[`probe-rs`]: https://github.com/probe-rs/probe-rs
 
-`panic-probe` can optionally log the panic message using the [`defmt`] logging framework.
+`panic-probe` can optionally log the panic message. Enabled one of the following features for that:
```
jonathanpallant, line 7:
```suggestion
`panic-probe` can optionally log the panic message. Enable one of the following features for that:

```

firmware/panic-probe/src/lib.rs
```diff
@@ -64,26 +67,22 @@ mod imp {
...
-            core::ptr::write_volatile(SHCSR, shcsr);
-        }
-    }
+pub fn abort() -> ! {
+    semihosting::process::abort();
+}
 
-    cortex_m::asm::udf();
+#[cfg(target_os = "none")]
+pub fn disable_isr() {
```
jonathanpallant, line 79: I think I would prefer `fn cortex_m_disable_isr()` and `fn riscv32_disable_isr()` and then use the `[cfg]` to `use crate::cortex_m_disable_isr as disable_isr;`.

Then if the neither cfg is `true`, you get a missing function error rather than a silent miscompilation that doesn't disable interrupts.

Also, I wonder if we should be using the critical-section crate, because disabling interrupts is insufficient on a multi-core system.

jonathanpallant:
Hi! Sorry for the delay. The draft looks good in principle - just a few comments.

SergioGasquez:
Hi! What is the status on this PR?

t-moe:
@SergioGasquez 
I'll probably abandon this, as I have created embedded-test now which is much more under our control. Also I feel that the testing framework shouldn't be coupled to whether you use defmt or not.

I'll also create panic_semihosting_abort which does what I tried to do in this PR with panic_probe.

Issue #794, opened 2023-11-06
feature: defmt-print detect executable changing


d3zd3z:
When debugging using the JLink debugger, it remains attached, and is able to (eventually) detect that the target has reset. However, if using defmt-print tcp to this target, it doesn't notice that the executable has changed, and will continue trying to decode messages from the old symbols. I'm not sure the best way to handle this, but just wanted to think about how it could be handled.

jonathanpallant:
We might be able to watch the file and reload it if it changes? You could also do something similar using a package like inotify-tools in the interim.

Urhengulas:
Fixed by https://github.com/knurling-rs/defmt/pull/807/.

PR #807, opened 2024-01-28
Adds a watch flag to watch the elf file and reloads the file if it changed

JomerDev:
This PR adds a --watch_elf/-w flag that watches the elf file for changes and reloads defmt-print on file modify or create.

To make this work I had to add tokio as dependency, since `stdin.read` otherwise blocked until some more serial data is received at which point it is too late to make the reload (if you do you end up with broken frames for the first few messages).

I also had to add `alterable_logger` to defmt-decoder to allow for the global logger to be overwritten after it was already set.

I'm open to making the flag and/or the logger changes depending on a feature.

This PR would fix issue #794

In verbose mode the logger also catches some logs from the notify crate used for the file watcher. This can probably be worked around if need be

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -13,7 +13,7 @@ and this project adheres to [Semantic Versioning](http://semver.org/).
 - [#843]: `defmt`: Sort IDs of log msgs by severity to allow runtime filtering by severity
 - [#822]: `CI`: Run `cargo semver-checks` on every PR
 - [#855]: `defmt-print`: Now uses tokio to make tcp and stdin reads async (in preparation for a `watch elf` flag)
-- [#807]: `defmt-print`: Add watch_elf flag to allow elf reload without restarting `defmt-print`
+- [#807]: `defmt-print`: Add `watch_elf` flag to allow ELF file reload without restarting `defmt-print`
 
 [#852]: https://github.com/knurling-rs/defmt/pull/852
 [#847]: https://github.com/knurling-rs/defmt/pull/847
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -13,6 +13,7 @@ and this project adheres to [Semantic Versioning](http://semver.org/).
 - [#843]: `defmt`: Sort IDs of log msgs by severity to allow runtime filtering by severity
 - [#822]: `CI`: Run `cargo semver-checks` on every PR
 - [#855]: `defmt-print`: Now uses tokio to make tcp and stdin reads async (in preparation for a `watch elf` flag)
+- [#807]: `defmt-print`: Add watch_elf flag to allow elf reload without restarting `defmt-print`
 
 [#852]: https://github.com/knurling-rs/defmt/pull/852
 [#847]: https://github.com/knurling-rs/defmt/pull/847
@@ -20,6 +21,7 @@ and this project adheres to [Semantic Versioning](http://semver.org/).
 [#843]: https://github.com/knurling-rs/defmt/pull/843
 [#822]: https://github.com/knurling-rs/defmt/pull/822
 [#855]: https://github.com/knurling-rs/defmt/pull/855
+[#807]: https://github.com/knurling-rs/defmt/pull/807
 
 ## [v0.3.8] - 2024-05-17
 
--- a/decoder/Cargo.toml
+++ b/decoder/Cargo.toml
@@ -43,7 +43,7 @@ object = { version = "0.35", default-features = false, features = [
 serde = { version = "1", features = ["derive"] }
 serde_json = { version = "1", features = ["arbitrary_precision"] }
 regex = "1"
-alterable_logger = "1.0.0"
+alterable_logger = "1"
 
 [features]
 # WARNING: API and wire format subject to change.
--- a/decoder/Cargo.toml
+++ b/decoder/Cargo.toml
@@ -43,6 +43,7 @@ object = { version = "0.35", default-features = false, features = [
 serde = { version = "1", features = ["derive"] }
 serde_json = { version = "1", features = ["arbitrary_precision"] }
 regex = "1"
+alterable_logger = "1.0.0"
 
 [features]
 # WARNING: API and wire format subject to change.
--- a/print/Cargo.toml
+++ b/print/Cargo.toml
@@ -18,7 +18,6 @@ clap = { version = "4.0", features = ["derive", "env"] }
 defmt-decoder = { version = "=0.3.11", path = "../decoder", features = [
     "unstable",
 ] }
-futures = "0.3"
 log = "0.4"
 notify = "6.1"
 tokio = { version = "1.38", features = ["full"] }
--- a/print/Cargo.toml
+++ b/print/Cargo.toml
@@ -20,3 +20,5 @@ defmt-decoder = { version = "=0.3.11", path = "../decoder", features = [
 ] }
 log = "0.4"
 tokio = { version = "1.38", features = ["full"] }
+notify = "6.1.1"
+futures = "0.3"
--- a/print/Cargo.toml
+++ b/print/Cargo.toml
@@ -18,7 +18,7 @@ clap = { version = "4.0", features = ["derive", "env"] }
 defmt-decoder = { version = "=0.3.11", path = "../decoder", features = [
     "unstable",
 ] }
-log = "0.4"
-tokio = { version = "1.38", features = ["full"] }
-notify = "6.1.1"
 futures = "0.3"
+log = "0.4"
+notify = "6.1"
+tokio = { version = "1.38", features = ["full"] }
--- a/print/src/main.rs
+++ b/print/src/main.rs
@@ -134,7 +134,7 @@ async fn has_file_changed(rx: &mut Receiver<Result<Event, notify::Error>>, path:
     true
 }
 
-async fn run_and_watch(opts: Opts, mut source: &mut Source) -> anyhow::Result<()> {
+async fn run_and_watch(opts: Opts, source: &mut Source) -> anyhow::Result<()> {
     let (tx, mut rx) = tokio::sync::mpsc::channel(1);
 
     let mut watcher = RecommendedWatcher::new(
@@ -151,7 +151,7 @@ async fn run_and_watch(opts: Opts, mut source: &mut Source) -> anyhow::Result<()
 
     loop {
         select! {
-            r = run(opts.clone(), &mut source) => r?,
+            r = run(opts.clone(), source) => r?,
             _ = has_file_changed(&mut rx, &path) => ()
         }
     }
--- a/print/src/main.rs
+++ b/print/src/main.rs
@@ -47,7 +47,7 @@ struct Opts {
     #[arg(short = 'V', long)]
     version: bool,
 
-    #[arg[short, long]]
+    #[arg(short, long)]
     watch_elf: bool,
 
     #[command(subcommand)]
--- a/print/src/main.rs
+++ b/print/src/main.rs
@@ -1,8 +1,9 @@
 use std::{
     env,
-    path::{Path, PathBuf},
+    path::{Path, PathBuf}, time::Duration,
 };
 
+use notify::{Config, Event, RecommendedWatcher, RecursiveMode, Watcher};
 use anyhow::anyhow;
 use clap::{Parser, Subcommand};
 use defmt_decoder::{
@@ -16,11 +17,11 @@ use defmt_decoder::{
 use tokio::{
     fs,
     io::{self, AsyncReadExt, Stdin},
-    net::TcpStream,
+    net::TcpStream, select, sync::mpsc::Receiver,
 };
 
 /// Prints defmt-encoded logs to stdout
-#[derive(Parser)]
+#[derive(Parser, Clone)]
 #[command(name = "defmt-print")]
 struct Opts {
     #[arg(short, required = true, conflicts_with("version"))]
@@ -44,11 +45,14 @@ struct Opts {
     #[arg(short = 'V', long)]
     version: bool,
 
+    #[arg[short, long]]
+    watch_elf: bool,
+
     #[command(subcommand)]
     command: Option<Command>,
 }
 
-#[derive(Subcommand)]
+#[derive(Subcommand, Clone)]
 enum Command {
     /// Read defmt frames from stdin (default)
     Stdin,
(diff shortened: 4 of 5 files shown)


Urhengulas:
Could this not be achieved with `cargo watch`?

JomerDev:
In theory yes, however afaik you'd usually run defmt-print by piping data from a serial port into it and I doubt `cargo watch` sends any through `stdin` received data to the inner command

CHANGELOG.md
```diff
@@ -11,11 +11,13 @@ and this project adheres to [Semantic Versioning](http://semver.org/).
 - [#803]: `CI`: Disable nightly qemu-snapshot tests
 - [#789]: `defmt`: Add support for new time-related display hints
 - [#783]: `defmt-decoder`: Move formatting logic to `Formatter`
+- [#807]: `defmt-print`: Add watch_elf flrag to allow elf reload without restarting `defmt-print`
```
jonathanpallant, line 14:
```suggestion
- [#807]: `defmt-print`: Add `watch_elf` flag to allow ELF file reload without restarting `defmt-print`
```

decoder/Cargo.toml
```diff
@@ -42,6 +42,7 @@ object = { version = "0.32", default-features = false, features = [
 ] }
 serde = { version = "1", features = ["derive"] }
 serde_json = { version = "1", features = ["arbitrary_precision"] }
+alterable_logger = "1.0.0"
```
jonathanpallant, line 45: What was this change for?

jonathanpallant (commented):
This generally looks OK to me, modulo a few comments.

I am a *little* bit worried about adding tokio to the defat-print dependency tree, but maybe someone can persuade me that this is better than running the TCP stream read in a thread with a 500ms timeout that checks some shared shutdown flag.

jonathanpallant:
I tested this on macOS, and I the file watching didn't appear to work.

I also wonder if some extra logging is useful, to let people know the ELF has been changed and reloaded.

print/Cargo.toml
```diff
@@ -19,3 +19,6 @@ defmt-decoder = { version = "=0.3.9", path = "../decoder", features = [
     "unstable",
 ] }
 log = "0.4"
+notify = "6.1.1"
+tokio = { version = "1", features = ["full"] }
```
Urhengulas, line 23: I share the concern of @jonathanpallant to add tokio as a dependency. Especially with full features. You probably can limit the features.

But otherwise I think a background thread to watch for changes sounds like a more lightweight solution.

print/src/main.rs
```diff
@@ -14,9 +12,18 @@ use defmt_decoder::{
...
+use notify::{Config, Event, RecommendedWatcher, RecursiveMode, Watcher};
 
+use tokio::io::{self, AsyncReadExt};
+use tokio::{
+    fs::{self},
+    io::Stdin,
+    net::TcpStream,
+    select,
+    sync::mpsc::Receiver,
+};
```
Urhengulas, line 24:
```suggestion
use tokio::{
    fs,
    io::{self, AsyncReadExt, Stdin},
    net::TcpStream,
    select,
    sync::mpsc::Receiver,
};
```

print/src/main.rs
```diff
@@ -40,11 +47,14 @@ struct Opts {
     #[arg(short = 'V', long)]
     version: bool,
 
+    #[arg[short, long]]
```
Urhengulas, line 50: Let's be consistent
```suggestion
    #[arg(short, long)]
```

Urhengulas (changes requested):

decoder/Cargo.toml
JomerDev: It's for lines 140 and 141 in `decoder/src/log/mod.rs`, since what is in this PR the `run` method (it was `main` before) is calling `init_logger` on each restart - meaning ELF file change. But since a logger is already set after the first time calling `init_logger` will crash after the first reload

print/Cargo.toml
JomerDev: Fair enough. I think my issue with using a background thread is that in the case of an ELF file change, I'll need to stop the read call on source (`let (n, eof) = source.read(&mut buf)?;`) as soon as possible to be able to reload the ELF file. 
I'm not sure that that is possible with using a timeout

Urhengulas:
Maybe it actually does make sense to use async. In the end `defmt-print` does a lot of waiting either on stdin or tcp. But I'd like if we first port the current behaviour of `defmt-print` to async in a separate PR and see if that makes sense. And then we add the watch flag after. 

@JomerDev Would you be willing to do that work?

JomerDev:
Done, see #855

Urhengulas:
Please rebase on main. I will review next week.

Urhengulas:
@JomerDev ping

JomerDev:
Sorry, I didn't have much time last week,. I'll try and get it done tomorrow

print/src/main.rs
```diff
@@ -94,20 +100,74 @@ const READ_BUFFER_SIZE: usize = 1024;
...
+    };
+
+    if opts.watch_elf {
+        let (tx, mut rx) = tokio::sync::mpsc::channel(1);
+
+        let mut watcher = RecommendedWatcher::new(
+            move |res| {
+                futures::executor::block_on(async {
+                    tx.send(res).await.unwrap();
+                })
```
Urhengulas, line 122: Does this execute on the tokio runtime or does it spawn a new one?

print/Cargo.toml
```diff
@@ -20,3 +20,5 @@ defmt-decoder = { version = "=0.3.11", path = "../decoder", features = [
 ] }
 log = "0.4"
 tokio = { version = "1.38", features = ["full"] }
+notify = "6.1.1"
```
Urhengulas, line 23: Should we disable crossbeam? See https://docs.rs/notify/latest/notify/index.html#crossbeam-channel--tokio

Urhengulas (changes requested):
First round of review, still need to actually test it

print/src/main.rs
Urhengulas: And would [tokio::runtime::Handle.current()](https://docs.rs/tokio/latest/tokio/runtime/struct.Handle.html#method.current) work here? Then we could get rid of the `futures` dependency.
```suggestion
                tokio::runtime::Handle::current().block_on(async {
                    tx.send(res).await.unwrap();
                })
```

Urhengulas, line 122: Would [tokio::runtime::Handle.current()](https://docs.rs/tokio/latest/tokio/runtime/struct.Handle.html#method.current) work here? Then we could get rid of the `futures` dependency.
```suggestion
                tokio::runtime::Handle::current().block_on(async {
                    tx.send(res).await.unwrap();
                })
```

Urhengulas: According to the docs it runs on the current thread.

> There is also a convenience function [block_on](https://docs.rs/futures/latest/futures/executor/fn.block_on.html) for simply running a future to completion on the current thread.

Urhengulas (changes requested):

print/src/main.rs
JomerDev: It's probably even easier to use `tx.blocking_send` instead, that way there is no need for any async context. I'll try that later today

print/Cargo.toml
JomerDev: Seeing as nobody could reliably confirm the issue and since std has been using the crossbeam implementation since 1.67 anyway I don't see a reason to do so.
See also the last comment on the issue: https://github.com/notify-rs/notify/issues/380#issuecomment-2186929524

JomerDev:
So, I don't know why the lint is complaining, I tried to apply it's suggestion but the rust-analyzer complains. I also don't get the lint when I run clippy locally

Urhengulas:
> So, I don't know why the lint is complaining, I tried to apply it's suggestion but the rust-analyzer complains. I also don't get the lint when I run clippy locally

I do see the suggestion locally and also my RA is happy with it. I am `clippy 0.1.80 (0514789 2024-07-21)`.

The diff is:

```diff
diff --git a/print/src/main.rs b/print/src/main.rs
index 88b0cfd..0125a27 100644
--- a/print/src/main.rs
+++ b/print/src/main.rs
@@ -134,7 +134,7 @@ async fn has_file_changed(rx: &mut Receiver<Result<Event, notify::Error>>, path:
     true
 }
 
-async fn run_and_watch(opts: Opts, mut source: &mut Source) -> anyhow::Result<()> {
+async fn run_and_watch(opts: Opts, source: &mut Source) -> anyhow::Result<()> {
     let (tx, mut rx) = tokio::sync::mpsc::channel(1);
 
     let mut watcher = RecommendedWatcher::new(
@@ -151,7 +151,7 @@ async fn run_and_watch(opts: Opts, mut source: &mut Source) -> anyhow::Result<()
 
     loop {
         select! {
-            r = run(opts.clone(), &mut source) => r?,
+            r = run(opts.clone(), source) => r?,
             _ = has_file_changed(&mut rx, &path) => ()
         }
     }
```

print/src/main.rs
```diff
@@ -94,20 +100,74 @@ const READ_BUFFER_SIZE: usize = 1024;
...
+            },
+            Config::default(),
+        )?;
+        let mut directory_path = opts.elf.clone().unwrap();
+        directory_path.pop(); // We want the elf directory instead of the elf, since some editors remove and recreate the file on save which will remove the notifier
+        watcher.watch(directory_path.as_ref(), RecursiveMode::NonRecursive)?;
```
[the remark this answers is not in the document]
Urhengulas: Nice!

Urhengulas:
I just tried it locally, but unfortunately could not get it to work.

# My assumption

I think the path filter is not correct.

# What I did

First, I patched `qemu-run` to not decode the defmt frames but just dump it raw.[^1]

I take `dbg.rs` as an example
- the source is at `firmware/qemu/src/bin/dbg.rs` 
- the binary is at `firmware/target/thumbv7m-none-eabi/debug/dbg`
- the defmt frames are at `firmware/qemu/src/bin/dbg.out` 

Then from `print/` I execute `defmt-print`, which prints the frames once:

```console
$ bat ../firmware/qemu/src/bin/dbg.out | cargo run -q -- -e ../firmware/target/thumbv7m-none-eabi/debug/dbg --watch-elf
TRACE x + 1 = 43
TRACE x - 1 = 41
INFO  the answer is 41
TRACE x - 2 = 40
TRACE x + 2 = 44
TRACE 
TRACE x = 42
```

But when I re-compile with (executed in repository root) `cargo xtask test-snapshot dbg --overwrite`, it does not seem to restart

I added some debug statements to  `has_file_changed` (see below) and the path filter blocks all the events. Probably it is because of absolute and relative path, or because of the `..`.

```text
[print/src/main.rs:124:13] &path = "../firmware/target/thumbv7m-none-eabi/debug/dbg"
[print/src/main.rs:124:13] &event = Event {
    kind: Remove(
        File,
    ),
    paths: [
        "/home/urhengulas/Documents/github.com/knurling-rs/defmt/print/../firmware/target/thumbv7m-none-eabi/debug/dbg",
    ],
    attr:tracker: None,
    attr:flag: None,
    attr:info: None,
    attr:source: None,
}
[print/src/main.rs:125:16] event.paths.contains(path) = false
[print/src/main.rs:124:13] &path = "../firmware/target/thumbv7m-none-eabi/debug/dbg"
[print/src/main.rs:124:13] &event = Event {
    kind: Create(
        File,
    ),
    paths: [
        "/home/urhengulas/Documents/github.com/knurling-rs/defmt/print/../firmware/target/thumbv7m-none-eabi/debug/dbg",
    ],
    attr:tracker: None,
    attr:flag: None,
    attr:info: None,
    attr:source: None,
}
[print/src/main.rs:125:16] event.paths.contains(path) = false
[print/src/main.rs:124:13] &path = "../firmware/target/thumbv7m-none-eabi/debug/dbg"
[print/src/main.rs:124:13] &event = Event {
    kind: Access(
        Close(
            Write,
        ),
    ),
    paths: [
        "/home/urhengulas/Documents/github.com/knurling-rs/defmt/print/../firmware/target/thumbv7m-none-eabi/debug/.cargo-lock",
    ],
    attr:tracker: None,
    attr:flag: None,
    attr:info: None,
    attr:source: None,
}
[print/src/main.rs:125:16] event.paths.contains(path) = false
```

```rs
async fn has_file_changed(rx: &mut Receiver<Result<Event, notify::Error>>, path: &PathBuf) -> bool {
    loop {
        if let Some(Ok(event)) = rx.recv().await {
            dbg!(&path, &event); // 124
            if dbg!(event.paths.contains(path)) {
                dbg!(&event); // 126
                match event.kind {
                    notify::EventKind::Create(_) | notify::EventKind::Modify(_) => {
                        dbg!(event); // 129
                        break;
                    }
                    _ => (),
                }
            }
        }
    }
    true
}
``` 

[^1]: I can send you a branch if you want it for testing.

Urhengulas (changes requested):
I just tried it locally, but unfortunately could not get it to work.

# My assumption

I think the path filter is not correct.

# What I did

First, I patched `qemu-run` to not decode the defmt frames but just dump it raw.[^1]

I take `dbg.rs` as an example
- the source is at `firmware/qemu/src/bin/dbg.rs` 
- the binary is at `firmware/target/thumbv7m-none-eabi/debug/dbg`
- the defmt frames are at `firmware/qemu/src/bin/dbg.out` 

Then from `print/` I execute `defmt-print`, which prints the frames once:

```console
$ bat ../firmware/qemu/src/bin/dbg.out | cargo run -q -- -e ../firmware/target/thumbv7m-none-eabi/debug/dbg --watch-elf
TRACE x + 1 = 43
TRACE x - 1 = 41
INFO  the answer is 41
TRACE x - 2 = 40
TRACE x + 2 = 44
TRACE 
TRACE x = 42
```

But when I re-compile with (executed in repository root) `cargo xtask test-snapshot dbg --overwrite`, it does not seem to restart

I added some debug statements to  `has_file_changed` (see below) and the path filter blocks all the events. Probably it is because of absolute and relative path, or because of the `..`.

```text
[print/src/main.rs:124:13] &path = "../firmware/target/thumbv7m-none-eabi/debug/dbg"
[print/src/main.rs:124:13] &event = Event {
    kind: Remove(
        File,
    ),
    paths: [
        "/home/urhengulas/Documents/github.com/knurling-rs/defmt/print/../firmware/target/thumbv7m-none-eabi/debug/dbg",
    ],
    attr:tracker: None,
    attr:flag: None,
    attr:info: None,
    attr:source: None,
}
[print/src/main.rs:125:16] event.paths.contains(path) = false
[print/src/main.rs:124:13] &path = "../firmware/target/thumbv7m-none-eabi/debug/dbg"
[print/src/main.rs:124:13] &event = Event {
    kind: Create(
        File,
    ),
    paths: [
        "/home/urhengulas/Documents/github.com/knurling-rs/defmt/print/../firmware/target/thumbv7m-none-eabi/debug/dbg",
    ],
    attr:tracker: None,
    attr:flag: None,
    attr:info: None,
    attr:source: None,
}
[print/src/main.rs:125:16] event.paths.contains(path) = false
[print/src/main.rs:124:13] &path = "../firmware/target/thumbv7m-none-eabi/debug/dbg"
[print/src/main.rs:124:13] &event = Event {
    kind: Access(
        Close(
            Write,
        ),
    ),
    paths: [
        "/home/urhengulas/Documents/github.com/knurling-rs/defmt/print/../firmware/target/thumbv7m-none-eabi/debug/.cargo-lock",
    ],
    attr:tracker: None,
    attr:flag: None,
    attr:info: None,
    attr:source: None,
}
[print/src/main.rs:125:16] event.paths.contains(path) = false
```

```rs
async fn has_file_changed(rx: &mut Receiver<Result<Event, notify::Error>>, path: &PathBuf) -> bool {
    loop {
        if let Some(Ok(event)) = rx.recv().await {
            dbg!(&path, &event); // 124
            if dbg!(event.paths.contains(path)) {
                dbg!(&event); // 126
                match event.kind {
                    notify::EventKind::Create(_) | notify::EventKind::Modify(_) => {
                        dbg!(event); // 129
                        break;
                    }
                    _ => (),
                }
            }
        }
    }
    true
}
``` 

[^1]: I can send you a branch if you want it for testing.

JomerDev:
Hmm. A possible fix could be to just remove every instance of `../` or `./` we find in the given path (since we only check that the event paths contains the given path), though that could potentially cause issues. A better solution might be to use `std::fs::canonicalize` on the given path, I would assume that `event.paths` only contains canonicalized paths (but I do not actually know if that is true)

Urhengulas:
@JomerDev I played a bit around with it and for it seems it is enough if we canonicalize the ELF path the user provides. Can you please test if that works for you?

Urhengulas:
@Urhengulas said:
> @JomerDev I played a bit around with it and for it seems it is enough if we canonicalize the ELF path the user provides. Can you please test if that works for you?

@jonathanpallant Can you please test if that also works on MacOS?

jonathanpallant:
On macOS:

Console 1: 

```console
$ nc -k -l 8000 | cargo run --bin defmt-print --release -- -e /Users/jonathan/Documents/ferrous-systems/rust-exercises/nrf52-code/radio-app/target/thumbv7em-none-eabihf/release/hello --show-skipped-frames -w
    Finished `release` profile [optimized + debuginfo] target(s) in 0.05s
     Running `target/release/defmt-print -e /Users/jonathan/Documents/ferrous-systems/rust-exercises/nrf52-code/radio-app/target/thumbv7em-none-eabihf/release/hello --show-skipped-frames -w`
Hello, world
Hello, test
```

Console 2:
```console
$ echo -n -e "\x02\x00" | nc localhost 8000
$ $EDITOR src/bin/hello.rs
$ cargo build --release
$ echo -n -e "\x02\x00" | nc localhost 8000
```

Seems to work as expected.

jonathanpallant:
I also tested it with `../../ferrous-systems/rust-exercises/nrf52-code/radio-app/target/thumbv7em-none-eabihf/release/hello` and `~/Documents/ferrous-systems/rust-exercises/nrf52-code/radio-app/target/thumbv7em-none-eabihf/release/hello` and both were fine.

Urhengulas (approved):
@jonathanpallant thank you for testing

@JomerDev thank you for the work

PR #808, opened 2024-02-05
Release defmt v0.3.6 (and others)

Urhengulas:
Release order:
- defmt-parser 0.3.4
- defmt-macros 0.3.7
- defmt 0.3.6
- defmt-decoder 0.3.10
- defmt-print 0.3.11

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -7,6 +7,8 @@ and this project adheres to [Semantic Versioning](http://semver.org/).
 
 ## [Unreleased]
 
+## [v0.3.6] - 2024-02-05
+
 - [#804]: `CI`: Remove mdbook strategy
 - [#803]: `CI`: Disable nightly qemu-snapshot tests
 - [#789]: `defmt`: Add support for new time-related display hints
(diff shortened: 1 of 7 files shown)


CHANGELOG.md
```diff
@@ -7,6 +7,8 @@ and this project adheres to [Semantic Versioning](http://semver.org/).
 
 ## [Unreleased]
 
+## [v0.3.6] - 2024-02-05
+
```
BriocheBerlin, line 11: Doesn't the release PR itself (#808) need to go in here as well?

BriocheBerlin (commented):
Thank you!

Not a problem of this PR, but maybe we can fix it here, it seem like https://github.com/knurling-rs/defmt/pull/787 is not showing up anywhere?

CHANGELOG.md
Urhengulas: We sometimes did and sometimes don't. It does not add a lot of value to include it, but often I added it for completenes. Will add it here as well.

Urhengulas: Decided to not add it, to reduce noise.

PR #827, opened 2024-04-07
Bumped rtt-target dependency to version 0.5

daniel-dbg-ginsburg:
When panic-probe depends on rtt-target 0.4 while the binary depends on version 0.5, the panics will fail to print. Hence the bump.


Urhengulas:
@daniel-dbg-ginsburg Can you please update the CHANGELOG as well?

Urhengulas:
Updated in #832

Issue #837, opened 2024-05-15
`cargo::rustc-check-cfg` breaks build for older Cargo versions


arctic-alpaca:
Hi,

[this commit](https://github.com/knurling-rs/defmt/commit/16d5eac499597a6a4cfd08bcb48aec9030c98f26) added cfg checking via `cargo::rustc-check-cfg`. Older versions of Cargo will fail with the following error:
```
error: unsupported output in build script of `defmt v0.3.7`: `cargo::rustc-check-cfg=cfg(c_variadic)`
Found a `cargo::key=value` build directive which is reserved for future use.
Either change the directive to `cargo:key=value` syntax (note the single `:`) or upgrade your version of Rust.
See https://doc.rust-lang.org/cargo/reference/build-scripts.html#outputs-of-the-build-script for more information about build script outputs.
```

Changing this to `cargo:rustc-check-cfg` fixes the issue.

PR #838, opened 2024-05-15
Switch to Cargo instruction compatible with older versions of Cargo

arctic-alpaca:
Fixes https://github.com/knurling-rs/defmt/issues/837

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -7,6 +7,10 @@ and this project adheres to [Semantic Versioning](http://semver.org/).
 
 ## [Unreleased]
 
+- [#805]: `defmt`: Switch to Cargo instruction compatible with older versions of Cargo
+
+[#838]: https://github.com/knurling-rs/defmt/pull/838
+
 ## defmt-macros v0.3.9
 
 - [#835]: `macros`: Fix some `defmt` crate name usage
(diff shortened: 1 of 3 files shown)


Urhengulas (changes requested):
@arctic-alpaca Will this work forever or will this syntax be deprecated in favor of the new one at some point? If it is gonna be deprecated we should add a not or a tracking issue.

arctic-alpaca:
This is indeed deprecated, but required for older versions of cargo to function:
`
Note: The old invocation prefix cargo: (one colon only) is deprecated and won’t get any new features. To migrate, use two-colons prefix cargo::, which was added in Rust 1.77. If you were using cargo:KEY=VALUE for arbitrary links manifest key-value pairs, it is encouraged to switch to cargo::metadata=KEY=VALUE. Stick to cargo: only if the support of Rust version older than 1.77 is required.
`
([source](https://doc.rust-lang.org/cargo/reference/build-scripts.html#outputs-of-the-build-script))

This PR brings the newly added `cargo:rustc-check-cfg` instruction in line with other instructions already present, that all use the deprecated invocation prefix.

For backwards compatibility while migrating to the new prefix, a feature could be added to switch to the old prefix for those stuck on older toolchain versions.

Urhengulas (approved):
I am fine with using the old syntax for now, also because it is consistent with the other invocations as you pointed out. We need to re-evaluate as soon as we need a new feature which only supports `::`. Until then I think we are fine like this.

PR #845, opened 2024-06-01
decoder: fix println!() records being printed with formatting

### LINKED ISSUE #844: defmt::println!() prints a <lvl> prefix
I don't think this used to be the case. Is this a mis-configuration from my part?

I don't have a small repro, but a big repro would be:
```
git clone https://github.com/google/wasefire.git
cd wasefire
./script/setup.sh
cargo xtask applet rust hello runner nordic --log=debug
```

which outputs:

```
DEBUG Runner starts.
└─ runner_nordic::__cortex_m_rt_main @ crates/runner-nordic/src/main.rs:118 
DEBUG Runner is initialized.
└─ runner_nordic::__cortex_m_rt_main @ crates/runner-nordic/src/main.rs:195 
DEBUG Loading applet.
└─ wasefire_scheduler::{impl#6}::run @ crates/scheduler/src/lib.rs:231 
DEBUG Schedule thread init[].
└─ wasefire_scheduler::{impl#6}::call @ crates/scheduler/src/lib.rs:421 
DEBUG Thread is done.
└─ wasefire_scheduler::{impl#6}::process_answer @ crates/scheduler/src/lib.rs:435 
DEBUG Schedule thread main[].
└─ wasefire_scheduler::{impl#6}::call @ crates/scheduler/src/lib.rs:421 
DEBUG Calling Debug(Println(Sig))
└─ wasefire_scheduler::{impl#6}::process_applet @ crates/scheduler/src/lib.rs:408 
<lvl> 0.095169: hello world
└─ wasefire_board_api::debug::Api::println @ crates/board/src/debug.rs:27  
DEBUG Thread is done.
└─ wasefire_scheduler::{impl#6}::process_answer @ crates/scheduler/src/lib.rs:435 
```

We can see that `defmt::debug!()` prints `DEBUG` while `defmt::println!()` prints `<lvl>`. In the past, it would look like this (no `<lvl>` prefix, the line would start with the timestamp, at least as far as I remember):

```
0.095169: hello world
└─ wasefire_board_api::debug::Api::println @ crates/board/src/debug.rs:27  
```

andresovela:
Fixes #844

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -7,6 +7,10 @@ and this project adheres to [Semantic Versioning](http://semver.org/).
 
 ## [Unreleased]
 
+- [#845]: `decoder`: fix println!() records being printed with formatting
+
+[#845]: https://github.com/knurling-rs/defmt/pull/845
+
 ## [v0.3.8] - 2024-05-17
 
 - [#840]: `defmt`: Support pre-1.77
--- a/decoder/src/log/format/mod.rs
+++ b/decoder/src/log/format/mod.rs
@@ -129,7 +129,7 @@ pub(super) struct LogSegment {
     pub(super) format: LogFormat,
 }
 
-#[derive(Debug, PartialEq, Clone)]
+#[derive(Debug, PartialEq, Clone, Default)]
 pub(super) struct LogFormat {
     pub(super) width: Option<usize>,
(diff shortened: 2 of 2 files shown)


Urhengulas (approved):

Urhengulas:
Thank you for the fix!

PR #849, opened 2024-06-06
Release `v0.5.0`

Urhengulas:
From now on all defmt crates will be released in lock-step. This means they will all always have the same version number. This means that sometimes crates will be released without changes.

This Release contains breaking changes in `defmt`, `defmt-decoder` and `defmt-parser`.


andresovela:
It'd be nice if this release included #848 :)

jonathanpallant:
The title says 0.3.13 but the defmt crate version has been bumped to 0.5.0 which will cause widespread breakage. Please can you clarify what version we're intending to release here.

jonathanpallant (changes requested):
Need to agree on process for handling breaking changes

PR #855, opened 2024-06-15
Make defmt-print async

JomerDev:
As discussed in #807 I've split out the work to make defmt-print async into a separate PR.
It changes very little, but please do test it to see if everything still works.
As for the tokio features, I've added all of the features needed instead of using `full` but afaik it wouldn't have made much of a difference as most features were needed anyway.

### DIFF
--- a/print/Cargo.toml
+++ b/print/Cargo.toml
@@ -19,4 +19,4 @@ defmt-decoder = { version = "=0.3.11", path = "../decoder", features = [
     "unstable",
 ] }
 log = "0.4"
-tokio = { version = "1.38.0", features = ["net", "macros", "time", "rt", "io-util", "fs", "io-std", "sync", "rt-multi-thread"] }
+tokio = { version = "1.38", features = ["full"] }
--- a/print/Cargo.toml
+++ b/print/Cargo.toml
@@ -19,3 +19,4 @@ defmt-decoder = { version = "=0.3.11", path = "../decoder", features = [
     "unstable",
 ] }
 log = "0.4"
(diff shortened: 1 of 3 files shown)


jonathanpallant:
Build time (main): `57.09s user 2.33s system 215% cpu 27.534 total`

Build time (PR): `69.60s user 2.75s system 226% cpu 31.895 total`

So an extra 4 seconds to compile, or +15%, on this macBook. I think that's fine.

jonathanpallant (approved):
Thank you for this PR. The cognitive overhead of switching to Tokio here seems small, so this looks good to me.

print/src/main.rs
```diff
@@ -15,6 +13,13 @@ use defmt_decoder::{
     DecodeError, Frame, Locations, Table, DEFMT_VERSIONS,
 };
 
+use tokio::io::{self, AsyncReadExt};
+use tokio::{
+    fs::{self},
+    io::Stdin,
+    net::TcpStream,
+};
```
Urhengulas, line 21:
```suggestion
use tokio::{
    fs,
    io::{self, AsyncReadExt, Stdin},
    net::TcpStream,
};
```

print/Cargo.toml
```diff
@@ -19,3 +19,4 @@ defmt-decoder = { version = "=0.3.11", path = "../decoder", features = [
     "unstable",
 ] }
 log = "0.4"
+tokio = { version = "1.38.0", features = ["net", "macros", "time", "rt", "io-util", "fs", "io-std", "sync", "rt-multi-thread"] }
```
Urhengulas, line 22: You convinced me

```suggestion
tokio = { version = "1.38", features = ["full"] }
```

Urhengulas (changes requested):

Urhengulas (approved):
Thank you!

PR #856, opened 2024-06-28
Add an impl Format for PanicInfo

dflemstr:
This impl is useful for panic handlers that use `defmt` to report panics, for example, `panic_probe` with the `defmt` feature.

The default `core::fmt` implementation for `PanicInfo` pulls in number formatting, padding support, `dyn Debug/Display` dependencies, etc. which can cause a lot of unnecessary code to be pulled in just to format `PanicInfo`.

### DIFF
--- a/defmt/src/impls/core_/panic.rs
+++ b/defmt/src/impls/core_/panic.rs
@@ -4,14 +4,9 @@ use super::*;
 
 impl<'a> Format for panic::PanicInfo<'a> {
     fn format(&self, f: Formatter) {
-        crate::write!(f, "panicked at {}:", self.location());
+        crate::write!(f, "panicked at {}", self.location());
         // TODO: consider supporting self.message() once stabilized, or add a crate feature for
         // conditional support
-
-        #[allow(deprecated)] // Deprecated on nightly
-        if let Some(msg) = self.payload().downcast_ref::<&'static str>() {
-            crate::write!(f, "\n{=str}", msg);
-        }
     }
 }
 
--- a/defmt/src/impls/core_/panic.rs
+++ b/defmt/src/impls/core_/panic.rs
@@ -1,0 +1,28 @@
+use core::panic;
+
+use super::*;
+
+impl<'a> Format for panic::PanicInfo<'a> {
+    fn format(&self, f: Formatter) {
+        crate::write!(f, "panicked at {}:", self.location());
+        // TODO: consider supporting self.message() once stabilized, or add a crate feature for
+        // conditional support
+
+        #[allow(deprecated)] // Deprecated on nightly
+        if let Some(msg) = self.payload().downcast_ref::<&'static str>() {
+            crate::write!(f, "\n{=str}", msg);
+        }
+    }
+}
+
+impl<'a> Format for panic::Location<'a> {
+    fn format(&self, f: Formatter) {
+        crate::write!(
+            f,
+            "{=str}:{=u32}:{=u32}",
+            self.file(),
+            self.line(),
+            self.column()
+        );
+    }
+}
--- a/defmt/src/impls/core_/panic.rs
+++ b/defmt/src/impls/core_/panic.rs
@@ -4,7 +4,11 @@ use super::*;
 
 impl<'a> Format for panic::PanicInfo<'a> {
     fn format(&self, f: Formatter) {
-        crate::write!(f, "panicked at {}", self.location());
+        if let Some(location) = self.location() {
+            crate::write!(f, "panicked at {}", location);
+        } else {
+            crate::write!(f, "panicked");
+        }
         // TODO: consider supporting self.message() once stabilized, or add a crate feature for
         // conditional support
     }
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -17,6 +17,7 @@ and this project adheres to [Semantic Versioning](http://semver.org/).
 - [#845]: `decoder`: fix println!() records being printed with formatting
 - [#843]: `defmt`: Sort IDs of log msgs by severity to allow runtime filtering by severity
 - [#822]: `CI`: Run `cargo semver-checks` on every PR
+- [#856]: `defmt`: Add a `Format` impl for `PanicInfo` and related types.
 
 [#859]: https://github.com/knurling-rs/defmt/pull/859
 [#858]: https://github.com/knurling-rs/defmt/pull/858
@@ -28,6 +29,7 @@ and this project adheres to [Semantic Versioning](http://semver.org/).
 [#845]: https://github.com/knurling-rs/defmt/pull/845
 [#843]: https://github.com/knurling-rs/defmt/pull/843
 [#822]: https://github.com/knurling-rs/defmt/pull/822
+[#856]: https://github.com/knurling-rs/defmt/pull/856
 
 ## [v0.3.8] - 2024-05-17
 
--- a/defmt/src/impls/core_/mod.rs
+++ b/defmt/src/impls/core_/mod.rs
@@ -12,6 +12,7 @@ mod cell;
 mod net;
 mod num;
 mod ops;
+mod panic;
 mod ptr;
 mod slice;
 
--- a/firmware/qemu/src/bin/panic_info.out
+++ b/firmware/qemu/src/bin/panic_info.out
@@ -1,0 +1,1 @@
+INFO PanicInfo: panicked at qemu/src/bin/panic_info.rs:14:5
--- a/firmware/qemu/src/bin/panic_info.rs
+++ b/firmware/qemu/src/bin/panic_info.rs
@@ -1,0 +1,23 @@
+#![no_std]
+#![no_main]
+
+use core::panic;
+use cortex_m_semihosting::debug;
+
+use defmt_semihosting as _; // global logger
+
+#[cortex_m_rt::entry]
+fn main() -> ! {
+    // Note: this test is a bit brittle in that the line/column number of the following panic is
+    // included in the test snapshot.  Hence, be mindful to update the snapshot if you want to
+    // add any additional code to this file above the following line!
+    panic!("aaah!")
+}
+
+#[panic_handler]
+fn panic(panic_info: &panic::PanicInfo) -> ! {
+    defmt::info!("PanicInfo: {=?}", panic_info);
+    loop {
+        debug::exit(debug::EXIT_SUCCESS)
+    }
+}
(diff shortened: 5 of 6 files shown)


jonathanpallant:
The deprecation notice says:

> this never returns anything useful

Do you have some example output of a panic handler using this formatter, where it did in fact return something useful?

Going forward, I wonder how we design a handler that works on both <= 1.80 and >= 1.81

jonathanpallant:
I also asked this on the wg-embedded Zulip channel.

dflemstr:
I tried to mirror the `core` code for stable in this case: https://github.com/rust-lang/rust/blob/129f3b9964af4d4a709d1383930ade12dfe7c081/library/core/src/panic/panic_info.rs#L160-L178

From what I can see, the deprecation note was added when `payload()` started to literally not return anything useful, ie the `NoPayload` struct: https://github.com/rust-lang/rust/blob/c3774be7411722d3695de2ab1da9a358d0d5c82c/library/core/src/panic/panic_info.rs#L107-L113

m-ou-se:
> when payload() started to literally not return anything useful

A `core::panic::PanicHandler` from a panic_handler has never returned anything useful, that behavior not a new thing. Trying to downcast a panic payload from a (no_std) panic_handler has simply always failed.

jonathanpallant:
@m-ou-se is their a recommendation for how people should render a `PanicInfo` object without using the `Debug` impl? In a way that works across Rust versions?

m-ou-se:
The only stable way today is to Display the PanicInfo itself. That will print the standard "panicked at {location}:\n{message}" message. That will still work in the future.

The only thing we're adding in the future is adding a .message() to get the `{message}` part separately. (We're not removing any functionality.)

dflemstr:
To me the root reason for making this change is to remove a dependency on `core::fmt` for padding and number formatting, which increases binary size significantly for some embedded applications, e.g. I have size constraints to build a firmware that fits in 30 kiB, and number formatting/padding takes at least 3 kiB, which ends up being 10% of the binary.

@m-ou-se any ideas how one could make it possible to implement efficient `defmt` encoding of `PanicInfo` given these constraints? Since it is definitely not possible to add a `defmt` feature to `core`, I suppose some sort of public API for this would be a requirement?

m-ou-se:
Yes, that's exactly why I was working on adding core::panic::PanicInfo::message(). ^^

That method will be stable in 1.81, most likely.

dflemstr:
Haha, well that makes sense! Great!  I suppose one has to just patiently wait then...

I guess the reason why I didn't make the connection immediately is that `PanicMessage` still encapsulates a `core::fmt::Arguments` which might still pull in the things I mentioned, but do you have any data on whether dead code elimination works better in this case? (Assuming that there are no panics in the binary that need to format integers of course)

Regarding this PR, I guess I should remove the downcast code, but should I also add a nightly feature flag that prints `self.message()`?

m-ou-se:
> I guess the reason why I didn't make the connection immediately is that PanicMessage still encapsulates a core::fmt::Arguments which might still pull in the things I mentioned

I'm confused, you want to be able to format the panic message without pulling in the code required to format that panic message?

That fmt::Arguments only pulls in whatever is used by the panic message. If that panic message is just a static string literal, it will not pull in any padding/formatting code etc.

But yes, if that panic message contains a formatted number or something (e.g. `panic!("{}", my_i32)`), that fmt::Arguments will contain a function pointer to `<i32 as Display>::fmt`.

jonathanpallant:
Yes, because this is *defmt*, a deferred formatting mechanism where strings are interned into debug sections and replaced with small integers in the code. What then goes over the logging transport is the numeric ID of the interned string, and a simple byte-serialisation of any fundamental types used as arguments. Where the argument is not a fundamental type, we implement the `defmt::Format` trait, which explains how to print a type using strings (interned) and fundamental types (copied byte wise).

Thus the whole point of this crate is to bypass the expensive `core::fmt` machinery.

m-ou-se:
> I guess if the panic message is `panic!("x = {x}")` we probably can't avoid it, but if the message is `panic!("Oh no, not again")` then we'd like to.

Yeah, that's my point. Display'ing PanicMessage is fine then. fmt::Arguments for a static string doesn't pull in any formatting code.

For the `panic!("x = {x}")` case, that can be addressed by https://github.com/rust-lang/rust/pull/104525 (once I find some time to finish it ^^').

dflemstr:
Hey, great pointers on the nuances between how static msgs are handled vs when they need integer interpolation.

To throw another idea on the stack: could `defmt` provide an implementation of `core::fmt::Write` to intercept the integer formatting? Then when you call `Display::fmt` on an `Arguments` object that uses integers in its format string, `defmt` could intercept it at that point to move over the integer formatting to the host too?

Or would that not be worth it, since we wouldn't get a monomorphized version of that `core::fmt::Write` struct, so all integer formatting calls would map to the same `defmt` symbol?

dflemstr:
@m-ou-se do you have any leads for how one would best offload the number formatting from `core::fmt` here? I'm happy to give this some more thoughts too but would appreciate a decent starting point for what might or might not be possible to land in the `core` lib, or what might or might not be possible to do with current stable/public APIs.

What we'd want to avoid is to have the number formatting code on the actual device; instead `defmt` will send the data to a host machine (e.g. in the case of a `u16` it would literally send two bytes in binary form to the host machine) and perform the formatting there.

As mentioned above, I see that `core::fmt::write` takes a `Write` impl but the `Write` trait only has `write_str`, `write_char`, and `write_fmt` of course.  That means we'd either need to find a way to intercept how specific types are formatted via some trait (with a different trait other than `Debug` or `Display`), or else get access to the parts of the `core::fmt::Arguments`  struct where one could discriminate based on the type of each argument part.

We have the `core` implementation details, where `core::fmt::rt::Placeholder` and `core::fmt::rt::Argument` might have the data we'd be looking for, but of course those aren't available in public APIs.

I'd imagine wanting to be able to write a function at some point that does something like:

```rust
fn defmt_write_args(f: defmt::Formatter, args: core::fmt::Arguments) {
  for part in args.parts() {
    match part {
      Part::LitStr(s) => defmt::write!("{=str}", s),
      Part::Placeholder(p, Value::U8(v)) => {
        // TODO: if we imagine that `p` has some type similar to `core::fmt::rt::Placeholder`, then
        // params like p.position, p.align, p.precision, p.width could also be sent along to the host here
        // to be used to influence formatting.
        defmt::write!("{=u8}", v);
      }
      Part::Placeholder(p, Value::U16(v)) => {
        defmt::write!("{=u16}", v);
      }
      //...
    }
  }
}
```

defmt/src/impls/core_/panic.rs
```diff
@@ -0,0 +1,28 @@
...
+impl<'a> Format for panic::PanicInfo<'a> {
+    fn format(&self, f: Formatter) {
+        crate::write!(f, "panicked at {}:", self.location());
+        // TODO: consider supporting self.message() once stabilized, or add a crate feature for
+        // conditional support
+
+        #[allow(deprecated)] // Deprecated on nightly
+        if let Some(msg) = self.payload().downcast_ref::<&'static str>() {
+            crate::write!(f, "\n{=str}", msg);
+        }
```
Urhengulas, line 14: Should we write something else in case the payload is _not_ a static string? Something like `<panic info>` maybe?

Urhengulas (changes requested):
Looks good.

Can you please also add some examples to the snapshot tests? The file is `firmware/qemu/src/bin/log.rs` and you can update the output with `cargo xtask test-snapshot --overwrite`.

dflemstr:
@Urhengulas thanks for the PR review! As per the above conversation I consider this PR to *not be ready to merge* since we are waiting to learn more from `libcore`/waiting for a stable release of `PanicInfo::message`.

I'm happy to make the changes that you requested once we have some clarity on the above points.

Urhengulas:
Am I seeing it correctly that this PR is 

> @Urhengulas thanks for the PR review! As per the above conversation I consider this PR to _not be ready to merge_ since we are waiting to learn more from `libcore`/waiting for a stable release of `PanicInfo::message`.
> 
> I'm happy to make the changes that you requested once we have some clarity on the above points.

Ah okay, I understood that the current version is not ideal but better than nothing. And then once `.message` gets stabilized we can improve it.

I will convert it to a draft PR. Please un-draft as soon as it is ready for merging again.

dflemstr:
@Urhengulas ah sure, we can do it that way, to add `.message()` support at a later point.

For now I completely removed the `.payload()` code because as mentioned above the downcast will never succeed anyways.

I didn't find a way to update the snapshot tests because it isn't possible to construct a `PanicInfo` without actually panicking, and I didn't find a reliable way to use `catch_unwind` or similar.  Please advise if you can think of another way of constructing `PanicInfo`.

As I think this is as good as things can get right now without waiting for the `libcore` changes, I will again unmark the PR as draft, as in I'm happy to get further review feedback :smile:

Urhengulas:
@dflemstr said:
> I didn't find a way to update the snapshot tests because it isn't possible to construct a `PanicInfo` without actually panicking, and I didn't find a reliable way to use `catch_unwind` or similar. Please advise if you can think of another way of constructing `PanicInfo`.

One (slightly involved) way I see, is to add a new snapshot test binary (probablu `firmware/qemu/src/bin/panic_info.rs`) and enable the `panic_internals` feature for it. Then you can use the `PanicInfo::internal_constructor` (https://doc.rust-lang.org/1.80.0/src/core/panic/panic_info.rs.html#42) to create a panic info and print it.

To add a new snapshot test, just create the binary, add it to the `NIGHTLY_SNAPSHOT_TESTS` https://github.com/knurling-rs/defmt/blob/5d467c2ef0b8e3d0c6894d5c0a4e643fc073a011/xtask/src/snapshot.rs#L30 and add the required feature (`panic_internals`) to https://github.com/knurling-rs/defmt/blob/5d467c2ef0b8e3d0c6894d5c0a4e643fc073a011/xtask/src/snapshot.rs#L79-L83

And then you can run it and create the output with `cargo xtask test-snapshot panic_info --overwrite`.

dflemstr:
@Urhengulas thanks for the pointers! I had gone down the route of trying to enable the compiler-internal feature flag, but even when I tried to disable all the safeguards:

```
#![allow(internal_features)]
#![feature(allow_internal_unstable, panic_internals)]

#[allow_internal_unstable(panic_internals)]
fn test_panic_info() {
  //...
}
```
...it still wouldn't compile.  However it was helpful that you explained how to add a separate snapshot test, which made it possible for now to have the test panic exactly once and succeed on panic.

dflemstr:
For the record, I don't know why the semver check is failing, and I think it is unrelated to this PR.

Urhengulas (approved):

PR #857, opened 2024-07-11
Support octal display hint

wetheredge:
Updated tests and docs based on the binary display hint.

### DIFF
--- a/firmware/qemu/src/bin/hints.rs
+++ b/firmware/qemu/src/bin/hints.rs
@@ -17,6 +17,8 @@ fn main() -> ! {
     defmt::info!("hex alt {=u8:#x}", x);
     defmt::info!("HEX     {=u8:X}", x);
     defmt::info!("HEX alt {=u8:#X}", x);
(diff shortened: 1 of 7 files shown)


firmware/qemu/src/bin/hints.rs
```diff
@@ -17,6 +17,8 @@ fn main() -> ! {
     defmt::info!("hex alt {=u8:#x}", x);
     defmt::info!("HEX     {=u8:X}", x);
     defmt::info!("HEX alt {=u8:#X}", x);
+    defmt::info!("octal   {=u8:o}", x);
```
jonathanpallant, line 20: Could we add the tests at the end, to minimise the diff on the expected output file?

jonathanpallant: Oh, actually, that probably isn't going to help. Never mind.

jonathanpallant (approved):
This looks good to me. Thank you for the PR!

PR #858, opened 2024-07-13
Implement "passthrough" Debug and Display instances for wrappers

wyager:
Useful for if you're writing code that switches between defmt::whatever and log::whatever or core::whatever using macros

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -7,6 +7,7 @@ and this project adheres to [Semantic Versioning](http://semver.org/).
 
 ## [Unreleased]
 
+- [#858]: `defmt`: Implement "passthrough" trait impls for *2Format wrappers
 - [#852]: `CI`: Update mdbook to v0.4.40
 - [#847]: `decoder`: Fix log format width specifier not working as expected
 - [#845]: `decoder`: fix println!() records being printed with formatting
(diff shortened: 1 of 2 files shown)


jonathanpallant:
Do you have a short example of where this is useful?

wyager:
@jonathanpallant Sure - when writing code that will be used on embedded *or* desktop, it's pretty common to use a macro adapter like https://github.com/embassy-rs/embassy/blob/ac06ca2fa0aa061a0053336433fa9cbfad5212ca/embassy-sync/src/fmt.rs which switches `whatever!` between `log::whatever!` and `defmt::whatever!`

If the object you are logging is non-defmt (e.g. `NorFlashErrorKind`), you must wrap it in a `Debug2Format` to use it with defmt. However, if you do that, now the `whatever!` macro will fail when you compile for desktop.

This PR fixes that issue. Now, something wrapped in `Debug2Format` will compile whether using `defmt` or `log` as your macro implementation.

jonathanpallant:
Oh! Right, that makes perfect sense.

jonathanpallant (approved):

jonathanpallant:
@Urhengulas any thoughts on this one?

Urhengulas (approved):
Sounds useful to me. Thank you for the contribution @wyager!

PR #871, opened 2024-10-09
Set MSRV to 1.76

Urhengulas:
This PR documents and tests Rust 1.76 as our MSRV. This is the version Ferrocene 24.05 is based on. Before we only guaranteed to support the latest stable.

I tried to store the MSRV in an env variable in the workflow file, but unfortunately env variables are not available in the strategy matrix, so it has to be repeated.

### DIFF
--- a/README.md
+++ b/README.md
@@ -22,7 +22,7 @@ To include `defmt` in your existing project, follow our [Application Setup guide
 
 ## MSRV
 
-The minimum supported Rust version is 1.76, which is Ferrocene 24.05.
+The minimum supported Rust version is 1.76 (or Ferrocene 24.05). `defmt` is tested against the latest stable Rust version and the MSRV.
 
 ## defmt ecosystem
 
--- a/README.md
+++ b/README.md
@@ -21,9 +21,8 @@ To include `defmt` in your existing project, follow our [Application Setup guide
 [Application Setup guide]: https://defmt.ferrous-systems.com/setup.html
 
 ## MSRV
-`defmt` always compiles on the [latest `stable` rust release](https://github.com/rust-lang/rust/releases/latest). This is enforced by our CI building and testing against this version.
 
-It still might work on older rust versions, but this isn't ensured.
+The minum supported Rust version is 1.77.
 
 ## defmt ecosystem
 
--- a/README.md
+++ b/README.md
@@ -22,7 +22,7 @@ To include `defmt` in your existing project, follow our [Application Setup guide
 
 ## MSRV
 
-The minum supported Rust version is 1.77.
+The minimum supported Rust version is 1.76, which is Ferrocene 24.05.
 
 ## defmt ecosystem
 
--- a/firmware/defmt-semihosting/src/lib.rs
+++ b/firmware/defmt-semihosting/src/lib.rs
@@ -6,6 +6,9 @@
 //! log frames so don't use those APIs.
 
 #![no_std]
+// nightly warns about static_mut_refs, but 1.76 (our MSRV) does not know about
+// that lint yet, so we ignore it it but also ignore if it is unknown
+#![allow(unknown_lints, static_mut_refs)]
 
 use core::sync::atomic::{AtomicBool, Ordering};
 
--- a/xtask/src/main.rs
+++ b/xtask/src/main.rs
@@ -7,6 +7,7 @@ use std::sync::Mutex;
 
 use anyhow::anyhow;
 use clap::{Parser, Subcommand};
+use utils::rustc_is_msrv;
 
 use crate::{
     snapshot::{test_snapshot, Snapshot},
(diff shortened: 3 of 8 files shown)


xtask/src/main.rs
```diff
@@ -136,8 +137,13 @@ fn test_cross(deny_warnings: bool) {
         false => vec![],
     };
 
+    let mut features = vec!["", "alloc"];
```
Hoverbear, line 140: 😬 I don't love the empty string but I see why it exists below, so it's fine. It just... stinks.

xtask/src/snapshot.rs
```diff
@@ -25,15 +28,19 @@ pub(crate) fn all_snapshot_tests() -> Vec<&'static str> {
         "hints",
         "hints_inner",
         "dbg",
-        "net",
         "panic_info",
     ];
     const NIGHTLY_SNAPSHOT_TESTS: &[&str] = &["alloc"];
+    const POST_MSRV_SNAPSHOT_TESTS: &[&str] = &["net"];
```
Hoverbear, line 34: It's unclear to me what "post msrv snapshot tests" are. If we raise the msrv should this change?

firmware/defmt-semihosting/src/lib.rs
```diff
@@ -6,6 +6,9 @@
 //! log frames so don't use those APIs.
 
 #![no_std]
+// nightly warns about static_mut_refs, but 1.76 (our MSRV) does not know about
+// that lint yet, so we ignore it it but also ignore if it is unknown
+#![allow(unknown_lints, static_mut_refs)]
```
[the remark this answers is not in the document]
Urhengulas: Saame

xtask/src/utils.rs
```diff
@@ -94,3 +94,9 @@ pub fn rustc_is_nightly() -> bool {
     let out = run_capturing_stdout(Command::new("rustc").args(["-V"])).unwrap();
     out.contains("nightly")
 }
+
+pub fn rustc_is_msrv() -> bool {
+    const MSRV: &str = "1.76";
+    let out = run_capturing_stdout(Command::new("rustc").args(["-V"])).unwrap();
```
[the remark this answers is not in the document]
Urhengulas: I copied it from `fn rustc_is_nightly` above, which does the same. That one contains the comment `if this crashes the system is not in a good state, so we'll not pretend to be able to recover`. I can add that comment if you like. But I don't see how we should handle that error at a higher level except also panicing/aborting there.

xtask/src/main.rs
Urhengulas: It does. In the past I tested `vec![None, Some("alloc")]`, which feels more clean but adds quite a bit of boilerplate.

xtask/src/snapshot.rs
Urhengulas: Yes. We have three brackets of tests
1. tests that work on the msrv, stable and nightly
2. tests that work on stable and nightly
3. tests that work on nightly only

Regarding 2: The `default` defmt features support the msrv, but some of the optional features require a newer version, e.g. `ip_in_core`. These tests are the ones in `POST_MSRV_TESTS`. Maybe `OPTIONAL_FEATURE_TESTS` is better?

README.md
```diff
@@ -21,9 +21,8 @@ To include `defmt` in your existing project, follow our [Application Setup guide
 [Application Setup guide]: https://defmt.ferrous-systems.com/setup.html
 
 ## MSRV
-`defmt` always compiles on the [latest `stable` rust release](https://github.com/rust-lang/rust/releases/latest). This is enforced by our CI building and testing against this version.
 
-It still might work on older rust versions, but this isn't ensured.
+The minimum supported Rust version is 1.76, which is Ferrocene 24.05.
```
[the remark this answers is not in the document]
Urhengulas: What do you think about this?

```suggestion
The minimum supported Rust version is 1.76 (Ferrocene 24.05). `defmt` is tested against the latest stable Rust version and the MSRV.
```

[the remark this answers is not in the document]
Hoverbear: Looks good, could we say `(or Ferrocene 24.05)` to really ephasize either works?

xtask/src/utils.rs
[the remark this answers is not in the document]
Hoverbear: Okay, I see. Let's keep as is.

xtask/src/snapshot.rs
Hoverbear: So `POST_MSRV` means "any version above our MSRV"? If so, a little comment would be really nice:

```rust
// Tests which run on all Rustc versions later than our MSRV
const POST_MSRV_SNAPSHOT_TESTS: &[&str] = &["net"];
```

Hoverbear (approved):
No blockers. :)

xtask/src/snapshot.rs
Urhengulas: Well it's not all versions. It's some version in `MSRV..=STABLE`.

Issue #867, opened 2024-09-27
feature req: defmt::expect! macro like unwrap!


dvdsk:
The `unwrap!` macro got implemented [see](https://github.com/knurling-rs/defmt/issues/161) , in that issue a `expect!` macro is also proposed and an implementation sketched out . Only the `unwrap!` macro got implemented however. Is there a reason for this?

To me an `expect!` macro would be very useful .

Urhengulas:
It would be useful indeed. Are you interested in implementing it?

dvdsk:
I'll have a go at it, not familiar with the internals of defmt however since there is a copypaste-ready implementation and I can have a look at `unwrap!` I can probably figure it out. Then we can hammer out the details in the PR

dvdsk:
So I had a quick look at noticed that `unwrap!` already implements the behavior of `expect!`. As in it allows a second argument that is a message. So `expect!` is implemented as `unwrap!`. See the docs here: https://docs.rs/defmt/latest/defmt/macro.unwrap.html

So this seems to be more an issue of discoverability then missing features. What to do there. I propose we make `expect` an alias for the current `unwrap`-macro. That way users that know of `unwrap!` can discover `expect!`. It might also be worth mentioning `unwrap!` in the defmt book. If I am not mistaken its not in there atm.

Urhengulas:
> So this seems to be more an issue of discoverability then missing features. What to do there. I propose we make `expect` an alias for the current `unwrap`-macro. That way users that know of `unwrap!` can discover `expect!`.

Agree

> It might also be worth mentioning `unwrap!` in the defmt book. If I am not mistaken its not in there atm.

and agree

dvdsk:
PR https://github.com/knurling-rs/defmt/pull/872 for the alias

PR #872, opened 2024-10-09
Improve discoverability with expect! as alias for unwrap!

dvdsk:
See related issue: https://github.com/knurling-rs/defmt/issues/867 Its easy to assume unwrap! can not have a message like std's expect.

Adding an expect macro will help both autocomplete guide programmers looking for one and those who discount the unwrap entry in the defmt api docs based on their knowledge that std's unwrap can not have a message.

The documentation of the expect alias highlights that unwrap works with a message to and points to the unwrap macro docs for details.

### DIFF
--- a/defmt/src/lib.rs
+++ b/defmt/src/lib.rs
@@ -190,6 +190,19 @@ pub use defmt_macros::panic_ as panic;
 /// [the manual]: https://defmt.ferrous-systems.com/macros.html
 pub use defmt_macros::unwrap;
 
+/// This is an alias for defmt's [`unwrap`] macro which supports messages like std's except.
+/// ```
+/// use defmt::expect;
+///
+/// # let arg = ();
+/// let x = value.expect(&format!("text {:?}", arg));
+/// let x = expect!(value, "text {:?}", arg); // arg must be implement `Format`
+/// ```
+///
+/// For the complete documentation see that of defmt's *unwrap* macro.
+// note: Linking to unwrap is broken as of 2024-10-09, it links back to expect
+pub use defmt_macros::unwrap as expect;
+
 /// Overrides the panicking behavior of `defmt::panic!`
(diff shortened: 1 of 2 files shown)


dvdsk:
pretty sure CI breaking is unrelated since this PR is only 1 use line + some lines of comments

dvdsk:
possible discussion point: because its an alias expect can be used without a message. Therefore you can not easily forbid unwrap's without a message by checking for the `unwrap` string.

Urhengulas:
> pretty sure CI breaking is unrelated since this PR is only 1 use line + some lines of comments

That's true. Please wait for #871 which fixes the nightly tests.

I will mark this PR with "breaking change", which is not true, but the CI check compares against the latest published version and not against main and we have some breaking changes in the main branch already.

defmt/src/lib.rs
```diff
@@ -190,6 +190,19 @@ pub use defmt_macros::panic_ as panic;
...
+/// ```
+/// use defmt::expect;
+///
+/// # let arg = ();
+/// let x = value.expect(&format!("text {:?}", arg));
+/// let x = expect!(value, "text {:?}", arg); // arg must be implement `Format`
+/// ```
+///
+/// For the complete documentation see that of defmt's *unwrap* macro.
+// note: Linking to unwrap is broken as of 2024-10-09, it links back to expect
```
Urhengulas, line 203: What does that mean?

Urhengulas (approved):

defmt/src/lib.rs
dvdsk: ahh that means that link to unwrap written as: `[unwrap](defmt::unwrap)` does not go to the unwrap item but to the new expect introduced in this PR. That is why the word unwrap is italic and not a link in line 202.

PR #878, opened 2024-10-28
Hex-encode defmt symbols to avoid compatibility issues.

Dirbaio:
Hex-encode defmt symbols to avoid compatibility issues.

defmt currently puts json as-is in symbol names, which can contain special
characters not normally found in symbol names like quotes `"`, braces `{}`
and spaces ` `.

This can cause compatibility issues with language features or tools that don't
expect this. For example it breaks `sym` in `asm!`.

This is a *breaking change* of the wire format, so this PR increases the
format numbre. `defmt-decoder` is updated to be able to decode the new format,
while keeping ability to decode older formats.

- **decoder: improve version mismatch error message, don't mention probe-run.**
- **Hex-encode defmt symbols to avoid compatibility issues.**


Dirbaio:
the `cargo-semver-check` ci error seems bogus? it's from a previous already-merged PR

jonathanpallant:
I've pinged the team internally so we can work out how to fix those tests. We're on version 4 of the wire format already, so we must have a process for bumping it.

decoder/src/elf2table/symbol.rs
```diff
@@ -38,7 +41,11 @@ pub(super) enum SymbolTag {
 
 impl Symbol {
     pub fn demangle(raw: &str) -> anyhow::Result<Self> {
-        serde_json::from_str(raw)
+        let mut raw = Cow::from(raw);
+        if let Some(s) = raw.strip_prefix("__defmt_") {
```
jonathanpallant, line 45: This string exists twice (once here, once in the macro), and it doesn't say a lot about the encoding used.

* Perhaps it should be `__defmt_hex_`, allowing for some future `__defmt_base64_` or similar.
* Is there somewhere it can be defined once and imported in the other place? If not, we should document the canonical form in a third place (like the book) and have the other two places refer to it.

Dirbaio: > Perhaps it should be __defmt_hex_, allowing for some future __defmt_base64_ or similar.

changed!

> Is there somewhere it can be defined once and imported in the other place? If not, we should document the canonical form in a third place (like the book) and have the other two places refer to it.

there's no crate that's a shared dep of the macro and the decoder. I've added a section to the book.

book/src/interning.md
```diff
@@ -73,3 +73,19 @@ static SYM: u8 = 0;
 
 The next thing to note is that each interned string symbol is one byte in size (because `static SYM` has type `u8`).
 Thanks to this the addresses of the symbols are consecutive: 0, 1, 2, etc.
+
+## Encoding
+
+Storing strings as-is in symbol names can cause compatibility problems, since it can contain any arbitrary character. For example:
+
+- The `'@'` character can't be used in symbol names because it's reserved for denoting symbol versions/
```
jonathanpallant, line 81: What's with the trailing `/`?

book/src/interning.md
```diff
@@ -73,3 +73,19 @@ static SYM: u8 = 0;
...
+- The double-quote character `'"'` causes issues with escaping if you use it with `sym` inside an `asm!()` call.
+
+To deal with this, strings are encoded to bytes as UTF-8, and then the bytes are hex-encoded.
```
jonathanpallant, line 84:
```suggestion
To deal with this, as of Wire Format Version 5, strings are encoded to bytes as UTF-8, and then the bytes are hex-encoded.
```

Maybe we should start to explain that multiple wire versions exist?

Dirbaio: done

book/src/interning.md
Dirbaio: I can't keyboard

Urhengulas:
> * the `cargo-semver-check` ci error seems bogus? it's from a previous already-merged PR

Yes, that's a known problem. I marked the PR as breaking change, which is not true, but skips the check.

>     * How to fix backcompat CI tests? it's expected that they fail, this is a wire format version bump.

Just bump the commit hash https://github.com/knurling-rs/defmt/blob/7d74e8b2fcf74ddc21bc3d38f30c5c1ba98c121b/xtask/src/backcompat.rs#L47.

book/src/interning.md
```diff
@@ -73,3 +73,19 @@ static SYM: u8 = 0;
...
+
+## Encoding
+
+Storing strings as-is in symbol names can cause compatibility problems, since it can contain any arbitrary character. For example:
+
+- The `'@'` character can't be used in symbol names because it's reserved for denoting symbol versions.
+- The double-quote character `'"'` causes issues with escaping if you use it with `sym` inside an `asm!()` call.
+
+To deal with this, as of Wire Format Version 5, strings are encoded to bytes as UTF-8, and then the bytes are hex-encoded.
+The symbol is prefixed with `__defmt_hex_` to denote it's is hex-encoded, and to allow for future expansion.
```
jonathanpallant, line 85:
```suggestion
The symbol is prefixed with `__defmt_hex_` to denote that it's hex-encoded, and to allow for future expansion.
```

jonathanpallant (commented):
One nit, and one thought.

By hex encoding the symbols, you stop people from scanning the output of `objdump -t` and seeing the log strings, modules, etc.

Is there some other encoding which would solve the problem of invalid characters (`"`, `{`, `@`, etc), but which retains the ability for a human to read the text? I'm thinking of something like [C trigraphs](https://en.wikibooks.org/wiki/C_Programming/C_trigraph) perhaps.

decoder/src/elf2table/symbol.rs
Urhengulas: > there's no crate that's a shared dep of the macro and the decoder. I've added a section to the book.

There is! It's the `defmt-parser`. I think it makes sense to put the magic string there.

book/src/interning.md
jonathanpallant: :)

decoder/src/elf2table/symbol.rs
jonathanpallant: Huh.

```console
➜  defmt git:(main) cargo tree -p defmt-macros | grep parser
├── defmt-parser v0.3.4 (/Users/jonathan/Documents/knurling-rs/defmt/parser)

➜  defmt git:(main) cargo tree -p defmt-decoder | grep parser
├── defmt-parser v0.3.4 (/Users/jonathan/Documents/knurling-rs/defmt/parser)
```

jonathanpallant:
We should probably fix the *input section* name at the same time (i.e. the `link_section = "...."). Currently this is JSON, unless you are on macOS when we hex encode it. We should probably always hex encode it, or hash it, or use the disambiguator. I *think* we do still need to put somewhere there (i.e. don't throw them all into `.defmt.error` et al) so that the garbage collection can throw out symbols that aren't referenced (which happens at the input section level).

jonathanpallant:
@Dirbaio did you have any thoughts on whether there's a more readable encoding than plain hex encoding here?

jonathanpallant:
Yes, good point, we'll need to Mangle All The Things.

> pipe their symbol names through a demangler will also see fully demangled defmt data

I'm mentally uprating my idea from "bizarre" to "huh, actually..."

Dirbaio:
closing because my motivation for doing this is gone. https://github.com/knurling-rs/defmt/pull/879#issuecomment-3312522725

I understand there's other motivations, like making it Just Compile on linux/windows/mac. If someone wants to explore those, feel free to pick up the code here and open a new PR.

PR #883, opened 2024-11-06
Mark decoder and parser not as unstable anymore

Urhengulas:
This PR marks the `defmt-decoder` and `defmt-parser` as _not_ unstable anymore. We did uphold semver for years (except one accidental breakage), so making that explicit just makes sense.

Users now do not need to pass `unstable` anymore in order to use this crate. To be precise they are not allowed to pass it anymore, therefore this is a breaking change, as it requires users to update their `Cargo.toml`. It will be released together with two other breaking changes. For `probe-rs` we gonna change it in https://github.com/probe-rs/probe-rs/pull/2809.

### DIFF
--- a/.github/workflows/cargo-semver-check.yml
+++ b/.github/workflows/cargo-semver-check.yml
@@ -14,15 +14,10 @@ jobs:
 
     steps:
       - uses: actions/checkout@v4
-      - name: Semver check crates without "unstable" feature
+      - name: Semver check host crates
         uses: obi1kenobi/cargo-semver-checks-action@v2
         with:
           feature-group: default-features
-      - name: Semver check crates with "unstable" feature
-        uses: obi1kenobi/cargo-semver-checks-action@v2
-        with:
-          exclude: defmt, defmt-json-schema
-          features: unstable
       - name: Semver check firmware crates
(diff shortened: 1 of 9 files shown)


jonathanpallant:
Would it be better to leave the feature flag, but as a no-op, and mark it as deprecated in the documentation? Then it isn't a breaking change.

We can take it out later, but let's give people a nice notice period so they can upgrade at their leisure.

Urhengulas:
@jonathanpallant Added it back

.github/workflows/cargo-semver-check.yml
```diff
@@ -14,15 +14,10 @@ jobs:
 
     steps:
       - uses: actions/checkout@v4
-      - name: Semver check crates without "unstable" feature
+      - name: Semver check host crates
         uses: obi1kenobi/cargo-semver-checks-action@v2
         with:
           feature-group: default-features
-      - name: Semver check crates with "unstable" feature
```
jonathanpallant, line 21: Maybe we should leave this in, just in case we add an unstable feature in the future? Not sure. Feel free to argue against that.

Urhengulas: I would add such a check when we add such a crate. There could be any other feature as well.

PR #884, opened 2024-11-06
Upgrade dependencies

Urhengulas:


decoder/Cargo.toml
```diff
@@ -31,11 +31,11 @@ defmt-json-schema = { version = "0.1", path = "./defmt-json-schema" }
 
 # elf2table
 anyhow = "1.0.65"
-gimli = { version = "0.29", default-features = false, features = [
```
jonathanpallant, line 34: Unless we rely on a feature in 0.31, we don't need this. Cargo will resolve to the latest anyway and there's no use on over-specifying the dependencies in case users have conflicting requirements.

decoder/Cargo.toml
```diff
@@ -31,11 +31,11 @@ defmt-json-schema = { version = "0.1", path = "./defmt-json-schema" }
...
+gimli = { version = "0.31", default-features = false, features = [
     "read",
     "std",
 ] }
-object = { version = "0.35", default-features = false, features = [
+object = { version = "0.36", default-features = false, features = [
```
jonathanpallant, line 38: Unless we rely on a feature in 0.36, we don't need this. Cargo will resolve to the latest anyway and there's no use on over-specifying the dependencies in case users have conflicting requirements.

macros/Cargo.toml
```diff
@@ -26,4 +26,4 @@ syn = { version = "2", features = ["full", "extra-traits"] }
 [dev-dependencies]
 maplit = "1"
 pretty_assertions = "1"
-rstest = { version = "0.19", default-features = false }
+rstest = { version = "0.23", default-features = false }
```
jonathanpallant, line 29: Unless we rely on a feature in 0.23, we don't need this. Cargo will resolve to the latest anyway and there's no use on over-specifying the dependencies in case users have conflicting requirements.

parser/Cargo.toml
```diff
@@ -10,10 +10,10 @@ repository = "https://github.com/knurling-rs/defmt"
 version = "0.3.4"
 
 [dependencies]
-thiserror = "1.0"
+thiserror = "2"
 
 [dev-dependencies]
-rstest = { version = "0.19", default-features = false }
+rstest = { version = "0.23", default-features = false }
```
jonathanpallant, line 16: Unless we rely on a feature in 0.23, we don't need this. Cargo will resolve to the latest anyway and there's no use on over-specifying the dependencies in case users have conflicting requirements.

jonathanpallant:
The two major bumps are OK I guess (not that we seem to require them). I think we need to justify the minor bumps though.

Urhengulas:
> The two major bumps are OK I guess (not that we seem to require them). I think we need to justify the minor bumps though.

No need for them. Will remove them.

jonathanpallant:
notify 7 bumps its MSRV to 1.72 (https://github.com/notify-rs/notify/releases/tag/notify-7.0.0) but we're offering 1.75 anyway.

thiserror 2 has an MSRV of 1.61, so that's fine.

CHANGELOG.md
```diff
@@ -7,6 +7,7 @@ and this project adheres to [Semantic Versioning](http://semver.org/).
 
 ## [Unreleased]
 
+- [#884]: Upgrade dependencies
```
jonathanpallant, line 10:
```suggestion
- [#884]: Upgrade dependencies: notify is now at v7, thiserror is now at v2 
```

jonathanpallant (approved):

PR #894, opened 2024-11-27
Allow bitflags v1 and v2 to coexist

### LINKED ISSUE #877: Allow bitflags version one and two to coexist
There is a `defmt::bitflags!` macro which depends on the `bitflags::bitflags!` macro from the `bitflags v1` crate. Since it's introduction there was a `bitflags v2`, but unfortunately we cannot just bump the dependency, since the `defmt::bitflags!` macro leaks implementation details of the `bitflags::bitflags!` macro and therefore the dependency update could break the code of defmt users.

Therefore we want to add two new cargo features to defmt: `bitflags1` and `bitflags2`. They each gate a `defmt::bitflags` macro which depends on `bitflags v1` and `bitflags v2` respectively. To not break any existing code, the `bitflags1` feature should be enabled by default. A `compile_error` when both are enabled sounds useful as well.

Note that there is an existing PR to update to bitflags v2: https://github.com/knurling-rs/defmt/pull/746.

manushT:
The patch breaks cargo's additive features 

Fixes #877.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -48,6 +48,7 @@ We have several packages which live in this repository. Changes are tracked sepa
 [defmt-v0.1.0]: https://github.com/knurling-rs/defmt/releases/tag/defmt-v0.1.0
 
 ### [defmt-next]
+
 * [#894] Add optional support for bitflags v2
 
 * [#938] Emit `option_env!("DEFMT_LOG")` so rustc depinfo can track it
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -48,6 +48,7 @@ We have several packages which live in this repository. Changes are tracked sepa
 [defmt-v0.1.0]: https://github.com/knurling-rs/defmt/releases/tag/defmt-v0.1.0
 
 ### [defmt-next]
+* [#894] Add optional support for bitflags v2
 
 * [#938] Emit `option_env!("DEFMT_LOG")` so rustc depinfo can track it
 * [#935] Add note in book's setup chapter to clarify staticlib setup
@@ -845,6 +846,7 @@ Initial release
 [#901]: https://github.com/knurling-rs/defmt/pull/901
 [#899]: https://github.com/knurling-rs/defmt/pull/899
 [#897]: https://github.com/knurling-rs/defmt/pull/897
+[#894]: https://github.com/knurling-rs/defmt/pull/894
 [#889]: https://github.com/knurling-rs/defmt/pull/889
 [#887]: https://github.com/knurling-rs/defmt/pull/887
 [#884]: https://github.com/knurling-rs/defmt/pull/884
--- a/defmt/src/export/mod.rs
+++ b/defmt/src/export/mod.rs
@@ -8,6 +8,9 @@ use crate::{Format, Formatter, Str};
 pub use self::integers::*;
 pub use bitflags::bitflags;
 
+#[cfg(feature = "bitflagsv2")]
+pub use bitflagsv2::bitflags as bitflagsv2;
+
 pub trait UnsignedInt {}
 impl UnsignedInt for u8 {}
 impl UnsignedInt for u16 {}
--- a/defmt/src/lib.rs
+++ b/defmt/src/lib.rs
@@ -370,7 +370,7 @@ pub use defmt_macros::bitflags;
 /// Generates a bitflags structure that can be formatted with defmt (using version 2 of the bitflags crate)
 ///
 
-/// Users of the defmt crate can enable the bitflagsv2 feature by adding defmt = { features = ["bitflagsv2"] }
+/// Users of the defmt crate can enable the bitflagsv2 feature by adding `defmt = {version = "1", features = ["bitflagsv2"] }`
 /// to their Cargo.toml. Bitflags version 2 introduces significant improvements over version 1, including a safer
 /// API with the replacement of the unsafe from_bits_unchecked method by the safe from_bits_retain method
 /// and enhanced serialization support via an optional serde feature.
--- a/defmt/src/lib.rs
+++ b/defmt/src/lib.rs
@@ -361,11 +361,21 @@ pub use defmt_macros::timestamp;
 ///         const ABC = Self::A.bits | Self::B.bits | Self::C.bits;
 ///     }
 /// }
-///
+/// #[cfg(feature = "bitflagsv2")]
+/// defmt::bitflagsv2! {
+///     struct Flags: u32 {
+///         const A = 0b00000001;
+///         const B = 0b00000010;
+///         const C = 0b00000100;
+///         const ABC = Self::A.bits() | Self::B.bits() | Self::C.bits()
+///     }
+/// }
 /// defmt::info!("Flags::ABC: {}", Flags::ABC);
 /// defmt::info!("Flags::empty(): {}", Flags::empty());
 /// ```
 pub use defmt_macros::bitflags;
+#[cfg(feature = "bitflagsv2")]
+pub use defmt_macros::bitflagsv2;
 
 #[doc(hidden)] // documented as the `Format` trait instead
 pub use defmt_macros::Format;
--- a/defmt/src/lib.rs
+++ b/defmt/src/lib.rs
@@ -337,7 +337,7 @@ pub use defmt_macros::timestamp;
 
 /// Generates a bitflags structure that can be formatted with defmt.
 ///
-/// This macro is a wrapper around the [`bitflags!`] crate, and provides an (almost) identical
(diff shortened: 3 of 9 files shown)


jonathanpallant:
Unfortunately this doesn't work because Library A might use defmt to describe some data type and expect the bitflags v1 API whilst Library B might use defmt but expect the bitflags v2 API (if we add it).

You could have a feature flag for bitflags v2 so that we keep the dependency list as short as possible, but it has to be additional and exported as a different macro with a different name.

jonathanpallant:
The CHANGELOG was reformatted. Can you rebase?

jonathanpallant:
The format check failed, I'm afraid.

CHANGELOG.md
```diff
@@ -27,6 +27,7 @@ We have several packages which live in this repository. Changes are tracked sepa
 > A highly efficient logging framework that targets resource-constrained devices, like microcontrollers
 
 [defmt-next]: https://github.com/knurling-rs/defmt/compare/defmt-v0.3.9...main
+[defmt-next]: https://github.com/knurling-rs/defmt/pull/894
```
jonathanpallant, line 30: We already have a markdown link for `[defmt-next]` which was correct. You can't add another.

The change you've made needs to be listed under `## [defmt-next]` heading.

CHANGELOG.md
```diff
@@ -47,6 +47,7 @@ We have several packages which live in this repository. Changes are tracked sepa
 [defmt-v0.1.0]: https://github.com/knurling-rs/defmt/releases/tag/defmt-v0.1.0
 
 ### [defmt-next]
+* [#894] Add optional support for bitflags v2
```
jonathanpallant, line 50: There is another change here already in `main`. You'll need to rebase.

manushT: done

manushT: sorry for the force pushes :)

CHANGELOG.md
```diff
@@ -48,6 +48,7 @@ We have several packages which live in this repository. Changes are tracked sepa
...
```
jonathanpallant, line 51: Markdown headings always have a blank line after.

```suggestion

* [#894] Add optional support for bitflags v2
```

defmt/src/lib.rs
```diff
@@ -361,11 +361,21 @@ pub use defmt_macros::timestamp;
 ///         const ABC = Self::A.bits | Self::B.bits | Self::C.bits;
 ///     }
 /// }
-///
+/// #[cfg(feature = "bitflagsv2")]
```
jonathanpallant, line 364: I think we need some more documentation here to explain why we have two bitflags macros now and how users should pick between them. And actually, currently `bitflagsv2` is completely undocumented because the new docs were added to the block for `bitflags`. You can check with `cargo doc -p defmt --features=bitflagsv2 --open`.

image

image

macros/src/items/bitflags.rs
```diff
@@ -69,11 +71,21 @@ fn codegen_flag_statics(input: &Input) -> Vec<TokenStream2> {
...
             let sym_name = construct::mangled_symbol_name(
-                "bitflags_value",
+                _tag,
                 &format!("{}::{i}::{}", input.ident(), flag.ident()),
             );
 
+            let bits_access = if is_v2 {
+                quote! { bits() }
+            } else {
+                quote! { bits}
```
jonathanpallant, line 87:
```suggestion
                quote! { bits }
```

jonathanpallant (changes requested):

jonathanpallant:
@manushT - we still have an outstanding comment on this one.

defmt/src/export/mod.rs
```diff
@@ -8,6 +8,9 @@ use crate::{Format, Formatter, Str};
 pub use self::integers::*;
 pub use bitflags::bitflags;
 
+#[cfg(feature = "bitflagsv2")]
+pub use bitflagsv2::bitflags as bitflagsv2;
```
vic1707, line 12: may I ask for the `bitflags` traits (`Bits` & `Flags`) to also be re-exported.
I need them, and other users might want them too, for a project and would prefer relying on the version installed by `defmt` instead of listing it manually.
Refers to #923

defmt/src/lib.rs
```diff
@@ -361,6 +361,31 @@ pub use defmt_macros::timestamp;
 ///         const ABC = Self::A.bits | Self::B.bits | Self::C.bits;
 ///     }
 /// }
+///
+/// defmt::info!("Flags::ABC: {}", Flags::ABC);
+/// defmt::info!("Flags::empty(): {}", Flags::empty());
+/// ```
+pub use defmt_macros::bitflags;
+
+/// Generates a bitflagsv2 structure that can be formatted with defmt. Users of the defmt crate can
```
LeonMatthesKDAB: This makes it sound like `bitflagsv2` is a different crate to `bitflags`, which is not the case, it's the same `bitflags` crate, but just using version 2.X.X.
Please try to formulate this clearer.

defmt/src/lib.rs
```diff
@@ -367,6 +367,42 @@ pub use defmt_macros::timestamp;
 /// ```
 pub use defmt_macros::bitflags;
 
+/// Users of the defmt crate can enable the bitflagsv2 feature by adding defmt = { features = ["bitflagsv2"] } 
+/// to their Cargo.toml. Bitflags version 2 introduces significant improvements over version 1, including a safer 
+/// API with the replacement of the unsafe from_bits_unchecked method by the safe from_bits_retain method 
+/// and enhanced serialization support via an optional serde feature.
+///
+/// This macro is a wrapper around the [`bitflags!`][bitflagsv2] macro of the bitflags! version 2 crate, and provides an (almost) identical
```
LeonMatthesKDAB, line 375: Nit pick: the macro name can have an `!` in it, the crate should not. So please change `bitflags! version 2 crate` to `bitflags version 2 crate`.
The `bitflags! macro` can keep the `!`

defmt/src/lib.rs
```diff
@@ -367,6 +367,42 @@ pub use defmt_macros::timestamp;
 /// ```
 pub use defmt_macros::bitflags;
 
+/// Users of the defmt crate can enable the bitflagsv2 feature by adding defmt = { features = ["bitflagsv2"] }
```
LeonMatthesKDAB, line 370: Nit: On the overview page, Rustdocs will list the first paragraph as the high-level description.

This paragraph is a little bit long for such a description, please add a short description like there is for the v1 macro
e.g.:
```
/// Generates a bitflags structure that can be formatted with defmt (using version 2 of the bitflags crate)
///
```

LeonMatthesKDAB (commented):
Mostly looks good to me, only two nitpicks.

defmt/src/export/mod.rs
jonathanpallant: That's reasonable - perhaps for v2 import the whole crate like:

```rust
pub use bitflags::{bitflags, Bits, Flags};

#[cfg(feature = "bitflagsv2")]
pub use bitflagsv2;
```

defmt/src/lib.rs
```diff
@@ -367,6 +367,44 @@ pub use defmt_macros::timestamp;
 /// ```
 pub use defmt_macros::bitflags;
 
+/// Generates a bitflags structure that can be formatted with defmt (using version 2 of the bitflags crate)
+///
+
+/// Users of the defmt crate can enable the bitflagsv2 feature by adding defmt = { features = ["bitflagsv2"] }
```
jonathanpallant, line 373:
```suggestion
/// Users of the defmt crate can enable the bitflagsv2 feature by adding `defmt = { version = "1", features = ["bitflagsv2"] }`
```

jonathanpallant (commented):
Minor nit on some formatting, and I like the proposal for re-exporting some useful types from the crate (to save people having to go and pull in the exact one that defmt is using).

I think this requires a Wire Format Version bump? In which case I think it's blocked until after 1.0. Maybe we should do a Version 5 with the encoded symbols and a bunch of other stuff.

defmt/src/export/mod.rs
jonathanpallant: @manushT any thoughts on this one?

jonathanpallant:
There's still an outstanding question around importing and re-exporting the bitflags API so that people don't have to try and include the same version of bitflags in their application that defmt was using. I think that's a good idea.

vic1707:
Any updates @manushT?

konstin:
This would be great for us to have to deduplicate the bitflags dependency. In uv we currently have 19 (transitive) crates depend on bitflags 2, with only defmt (through jiff) still pulling in bitflags 1.

PR #895, opened 2024-11-27
Fix defmt-semihosting metadata.

jonathanpallant:
Without it, you can't publish.


skade (approved):
Adds general metadata.

PR #905, opened 2024-12-02
Fix address of logo on docs.rs

### LINKED ISSUE #904: The logo on docs.rs/defmt is missing
You just get a missing-image box at docs.rs/defmt.

Urhengulas:
The logo is not available at the previous address anymore, since we deprecated the knurling website some time ago. This PR changes the address to the logo from the defmt book.

`cargo doc` now shows the logo again, but this will only solve it for new releases. The old releases will keep the old address. If we want to fix them we have to host the logo at the previous address again.

Fix #904.

### DIFF
--- a/defmt/src/lib.rs
+++ b/defmt/src/lib.rs
@@ -12,7 +12,7 @@
 
 #![cfg_attr(not(feature = "unstable-test"), no_std)]
 // NOTE if you change this URL you'll also need to update all other crates in this repo
-#![doc(html_logo_url = "https://knurling.ferrous-systems.com/knurling_logo_light_text.svg")]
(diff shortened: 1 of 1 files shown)


Urhengulas:
Maybe it is a better idea to include the logo locally, so we do not depend on the address of the defmt book?

jonathanpallant (approved):

jonathanpallant:
> Maybe it is a better idea to include the logo locally, so we do not depend on the address of the defmt book?

Yes, we probably should. It's not that large.

Urhengulas:
> Yes, we probably should. It's not that large.

Okay, gonna close in favor of a local image.

Issue #860, opened 2024-07-31
`panic-probe` "only supports Cortex-M targets" (could be mentioned in the docs)


lure23:
Tried using `panic-probe` on a RISC V target, and faced [this](https://github.com/knurling-rs/defmt/blob/5d467c2ef0b8e3d0c6894d5c0a4e643fc073a011/firmware/panic-probe/src/lib.rs#L24):

```
error: `panic-probe` only supports Cortex-M targets (thumbvN-none-eabi[hf])
```

I think this could be mentioned on the [documentation page](https://github.com/knurling-rs/defmt/tree/main/firmware/panic-probe).

PR #907, opened 2024-12-02
Add note to panic-probe readme about supported targets

BenFordTytherington:
Fix https://github.com/knurling-rs/defmt/issues/860.


Urhengulas (approved):
Thank you! But please always include the ticket your PR is related to.

PR #915, opened 2024-12-16
Introduce feature to disable blocking mode

mdrssv:
Hi,

I've added the feature `disable-blocking-mode` which prevents the host from putting rtt into blocking mode, by which I've got bitten a couple of times because I tend to just unplug my devices. However with blocking mode they will just hang once the buffer is full. Since I do not log too much blocking is not of great benefit to me whereas not having the mcu hang is.

### DIFF
--- a/firmware/defmt-rtt/src/channel.rs
+++ b/firmware/defmt-rtt/src/channel.rs
@@ -27,6 +27,7 @@ impl Channel {
         // the host-connection-status is only modified after RAM initialization while the device is
         // halted, so we only need to check it once before the write-loop
         let write = match self.host_is_connected() {
+            _ if cfg!(feature = "disable-blocking-mode") => Self::nonblocking_write,
             true => Self::blocking_write,
             false => Self::nonblocking_write,
         };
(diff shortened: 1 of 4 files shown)


firmware/defmt-rtt/src/channel.rs
```diff
@@ -27,8 +27,8 @@ impl Channel {
         // the host-connection-status is only modified after RAM initialization while the device is
         // halted, so we only need to check it once before the write-loop
         let write = match self.host_is_connected() {
-            true => Self::blocking_write,
-            false => Self::nonblocking_write,
+            true if cfg!(not(feature = "disable-blocking-mode")) => Self::blocking_write,
+            _ => Self::nonblocking_write,
```
Urhengulas, line 31:
```suggestion
            _ if cfg!(feature = "disable-blocking-mode") => Self::nonblocking_write,
            true => Self::blocking_write,
            false => Self::nonblocking_write,
```

Urhengulas (changes requested):
Looks generally good to me. But I want @jonathanpallant to have a look before merging to ensure it does not interfere with defmt 1.0.

jonathanpallant:
This seems fine to me - it adds a new feature and takes nothing away.

jonathanpallant (approved):

jonathanpallant:
@Urhengulas can you clear your blocking marker?

PR #917, opened 2025-01-08
feat: add teardown fn for defmt-test

mhatzl:
This PR adds the possibility to specify a teardown function in a defmt-test module. 

Similar to the before_each and after_each attributes, the teardown function can only be specified once and may take the state from the init function as parameter. 

In contrast to the *_each functions, the teardown function is called only once at the end of the test run directly before exiting. 

This function is needed, for example to handle cleanup code that must only be run at the end of a test run.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -741,6 +741,7 @@ Initial release
 > A test harness for embedded devices
 
 [defmt-test-next]:  https://github.com/knurling-rs/defmt/compare/defmt-test-v0.3.2...main
+[defmt-test-v0.3.3]:  https://github.com/knurling-rs/defmt/compare/defmt-test-v0.3.2...main
 [defmt-test-v0.3.2]:  https://github.com/knurling-rs/defmt/releases/tag/defmt-test-v0.3.2
 [defmt-test-v0.3.1]:  https://github.com/knurling-rs/defmt/releases/tag/defmt-test-v0.3.1
 [defmt-test-v0.3.0]:  https://github.com/knurling-rs/defmt/releases/tag/defmt-test-v0.3.0
@@ -752,6 +753,8 @@ Initial release
 [defmt-test-v0.1.0]:  https://github.com/knurling-rs/defmt/releases/tag/defmt-test-v0.1.0
 
 ### [defmt-test-next]
+
+### [defmt-test-v0.3.3] (unreleased)
 
 ### [defmt-test-v0.3.2] (2024-03-05)
 
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -741,7 +741,6 @@ Initial release
 > A test harness for embedded devices
 
 [defmt-test-next]:  https://github.com/knurling-rs/defmt/compare/defmt-test-v0.3.2...main
-[defmt-test-v0.3.3]:  https://github.com/knurling-rs/defmt/compare/defmt-test-v0.3.2...main
 [defmt-test-v0.3.2]:  https://github.com/knurling-rs/defmt/releases/tag/defmt-test-v0.3.2
 [defmt-test-v0.3.1]:  https://github.com/knurling-rs/defmt/releases/tag/defmt-test-v0.3.1
 [defmt-test-v0.3.0]:  https://github.com/knurling-rs/defmt/releases/tag/defmt-test-v0.3.0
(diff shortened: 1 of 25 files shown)


firmware/defmt-test/macros/tests/ui/teardown-has-invalid-function-signature.stderr
```diff
@@ -0,0 +1,5 @@
+error: `#[teardown]` function must have signature `fn(state: &mut Type)` (parameter is optional)
```
Urhengulas, line 1: maybe we can change this to:
```suggestion
error: `#[teardown]` function must have signature `fn()` or `fn(state: &mut T)`
```

Urhengulas (changes requested):
Hi Manuel 👋🏾  thank you for your contribution. Looks very good to me, only a few minor nits.

CHANGELOG.md
```diff
@@ -753,6 +754,8 @@ Initial release
 
 ### [defmt-test-next]
 
+### [defmt-test-v0.3.3] (unreleased)
```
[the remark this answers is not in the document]
mhatzl: Ah perfect. I made those changes to pass the CI. 

will revert them 😊

firmware/defmt-test/README.md
```diff
@@ -190,6 +190,45 @@ $ cargo test -p testsuite
...
+
+    // This function is called after all test cases passed.
+    // It accesses the state created in `init`,
+    // though like with `test`, state access is optional.
+    #[teardown]
+    fn teardown(state: &mut super::MyState) {
+        defmt::println!("State flag in teardown is {}", state.flag);
+    }
+
+    // but this test does
```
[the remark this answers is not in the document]
mhatzl: ah it was a leftover from copy pasting. my bad

firmware/defmt-test/macros/src/lib.rs
```diff
@@ -472,6 +565,11 @@ struct Init {
     state: Option<Box<Type>>,
 }
 
+struct Teardown {
+    func: ItemFn,
+    input: Option<Input>,
```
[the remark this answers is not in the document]
mhatzl: correct. it has the same behavior as before/after_each and test functions

firmware/defmt-test/macros/tests/ui/teardown-has-invalid-function-signature.stderr
mhatzl: I copied the message from before_each. after_each and test also have the same message. should I change all of them?

mhatzl:
Hi Johann,
thanks for the review.
I will revert the version and changelog changes.

for the error msg I wait on how you decide

firmware/defmt-test/macros/tests/ui/teardown-has-invalid-function-signature.stderr
Urhengulas: yes. if we use the same message (in parts) in all of them maybe we can store them in a constant?

mhatzl: I went with a macro to concatenate the string literals.

The concat!() macro only works with string literals. That‘s why I chose the macro instead of a constant.

firmware/defmt-test/macros/src/lib.rs
```diff
@@ -15,6 +15,16 @@ pub fn tests(args: TokenStream, input: TokenStream) -> TokenStream {
...
 
+macro_rules! fn_state_signature_msg {
+    ($name:literal) => {
+        concat!(
+            "`#[",
+            $name,
+            "]` function must have signature `fn()` or `fn(state: &mut T)`"
+        )
+    };
+}
```
Urhengulas, line 26: Could this also be a function? If not, it is okay as well. I just think functions are usually easier to understand.

```suggestion
fn fn_state_signature_msg(name: &str) -> String {
    format!("`#[{name}]` function must have signature `fn()` or `fn(state: &mut T)`")
}
```

Urhengulas (approved):
Looks amazing. One question around the macro, if you don't like it I will merge.

firmware/defmt-test/macros/src/lib.rs
mhatzl: not sure how Rust optimizes format!() in this case, because String is heap allocated while the macro is compile time (so constant).

Thats why I went with a macro.

For defmt-test to be usable in no_std, it is not possible to use String.

Urhengulas: > For defmt-test to be usable in no_std, it is not possible to use String.

True, makes sense

Issue #861, opened 2024-08-18
Object safe Format


EdBuilds:
Hello, 
Would it be possible to make the `Format` trait object safe? I would like to pass messages around by using 
```
&dyn defmt::Format 
```
But the static function in the trait prevents that.
Cheers!

jonathanpallant:
I looked into this for you.

The serialiser needs to report what kind of *thing* is being serialised so the deserialiser can unpack it. This is done using an interned string:

```rust=
pub fn fmt<T: Format + ?Sized>(f: &T) {
    istr(&T::_format_tag());
    f._format_data();
}

pub fn fmt_slice<T: Format>(values: &[T]) {
    usize(&values.len());
    istr(&T::_format_tag());
    for value in values {
        value._format_data();
    }
}
```

If `_format_tag()` takes `&self`, then you can't call `T::_format_tag()` when serialising a slice, and we can't guarantee that the slice has non-zero length, so we can't call `values[0]._format_tag()` either. So the trait has to include a static method, and that means we can't have dynamic trait objects.

So with the current design, no, it's not possible to have a `&dyn defmt::Format`. Can you elaborate on why you would like to make `&dyn` references to things that can be formatted with defmt? What would you wish to do with those reference?

jordens:
One could change the slice format and omit the inner format tag in case of an empty slice. The data is already there to recognize it on the decoder (zero length). And it would save at least one byte. Then `_format_tag()` can become a regular method.

jonathanpallant:
Ah.

```rust
pub fn fmt_slice<T: Format>(values: &[T]) {
    if let Some((first, remainder)) = values.split_first() {
        usize(&values.len());
        istr(&first.format_tag());
        first._format_data();
        for value in remainder {
            value._format_data();
        } 
    } else {
        usize(0);
    }
}
```

That would be a breaking change to the encoding, which would need a wire format version bump. But it perhaps it is worth it.

thejpster:
We discussed this on the Matrix chat. If you'd like to propose a PR, we can look at exactly how breaking a change it is.

thejpster:
The trick is handling the difference between a slice of T (where the format is interned once) and a slice of &dyn Format, where you need to intern the format for each item in the slice. It wasn't immediately clear how to do that.

jonathanpallant:
We discussed this on the Matrix chat. If you'd like to propose a PR, we can look at exactly how breaking a change it is.

[...]

PR #926, opened 2025-01-29
make Format trait dyn compatible

jordens:
* `Format::_format_tag()` -> `Format::_format_tag(&self)`
* change the wire format for empty slices a `usize(0)`
* change the wire format for empty arrays to be empty

closes: #861

### DIFF
--- a/defmt/src/export/mod.rs
+++ b/defmt/src/export/mod.rs
@@ -185,7 +185,13 @@ pub fn u8_array(a: &[u8]) {
 
 // NOTE: This is passed `&[u8; N]` – it's just coerced to a slice.
 pub fn fmt_array<T: Format>(a: &[T]) {
-    fmt_slice(a)
+    if let Some((first, remainder)) = a.split_first() {
+        istr(&first._format_tag());
+        first._format_data();
+        for value in remainder {
+            value._format_data();
+        }
+    }
 }
 
 /// Implementation detail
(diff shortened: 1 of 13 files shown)


jonathanpallant:
Someone can now write:

```rust
let slice: &[&dyn defmt::Format] = &[&1.0, &2];
defmt::info!("{}", slice);
```

But I see:

```rust
INFO [1.0, 1e-323]
```

I think this is because the bits for `2i32` have been interpreted as an `f32`.

jordens:
Yeah. That's problematic.

jordens:
After a lengthy discussion in the chat:

There is a way to prevent that along the lines of
https://play.rust-lang.org/?version=stable&mode=debug&edition=2021&gist=08d81d35efe5c2cbe41b614f409bc72c

But it requires dropping the blanket `impl Format` for `&` (and `&mut`, `Box` etc) of any unsized type. Instead one would need to explicitly add impls for such containers of known unsized types: `[T]` and `str` (maybe others?).

jonathanpallant:
That would count as a breaking change for anyone currently relying on that impl, and so I don't think we can do that. Thank you for digging through the details though and putting together this PR. It is really appreciated.

jordens:
That leaves the change to the wire format omitting the tag for empty slices and arrays. What are the prospects there?

jordens:
#929

PR #929, opened 2025-01-31
wire format: omit the inner type tag for empty slices and arrays

jordens:
When serializing a zero-length slice/array, the item type tag is currently always emitted.
That doesn't need to be the case. Removing it shortens the wire format (empty slices tend to be relatively common).

* [ ] Wire format bump

Originally this came about as a core part of #926 (making `Format` `dyn` compatible) which ultimately didn't pan out for other reasons. It still looks like a useful improvement independent of that PR.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -50,6 +50,7 @@ We have several packages which live in this repository. Changes are tracked sepa
 ### [defmt-next]
 
 * [#914] Add cargo-deny as a CI action to check crate security and licensing
+* [#929] Change the wire format to omit the inner type tag for empty slices and empty arrays.
 
 ### [defmt-v0.3.10] (2024-11-29)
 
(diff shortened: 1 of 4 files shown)


jonathanpallant:
This looks OK but it needs a wire-format bump so we need to wait until after 1.0 is released, when we have a bunch of stuff to wrap up into Wire Format Version 5.

jonathanpallant:
Still want to get this in, but the process will probably need to be:

* Merge to main
* Bump the Wire Format
* Do a defmt-parser / defmt-decoder release that knows about the new format
* Send a PR to probe-rs
* Get the PR merged
* Wait for the new probe-rs release
* *Then* release a defmt-1.1.0 with the new wire format enabled by default

It's that or we find a way to use version 4 by default and let people opt-in to version 5. We should probably have a road-map so they know when defmt releases are coming and which probe-rs versions they are going to need to have.

PR #935, opened 2025-02-13
clarify setup for static libraries

newcomertv:
Applying the defmt linker args to a cargo package which produces a static lib does not yield any warning.
A user attempting to use this as a module of a larger C project may be surprised to not receive any logs.

After dealing with this issue myself I think its worth putting this into a somewhat prominent user guide section as the final binary gets built and the issue isn't immediately obvious.

### DIFF
--- a/book/src/setup.md
+++ b/book/src/setup.md
@@ -36,7 +36,7 @@ rustflags = [
 
 > Note :
 > If you intend on linking your project as a **static library** the linkerscript contents should be applied to the final binary.
-> Applying `defmt.x` to the static lib has no effect.
+> Adding the `-C link-arg=` option to static libraries has no effect because they are only archived, not linked.
 
 ### `#[global_logger]`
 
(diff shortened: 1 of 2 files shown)


book/src/setup.md
```diff
@@ -34,6 +34,10 @@ rustflags = [
 ]
 ```
 
+> Note :
+> If you intend on linking your project as a **static library** the linkerscript contents should be applied to the final binary.
+> Applying `defmt.x` to the static lib has no effect.
```
jonathanpallant, line 39:
```suggestion
> Adding the `-C link-arg=` option to static libraries has no effect because they are only archived, not linked.
```

jonathanpallant (approved):

jonathanpallant:
I appreciate you taking them time to improve our documentation to cover this usecase, and I apologise for our documentation previously being unclear in this area.

PR #940, opened 2025-02-25
add serial port input to `defmt-print`

brightbloom:


print/src/main.rs
```diff
@@ -65,11 +67,23 @@ enum Command {
         #[arg(long, env = "RTT_PORT", default_value_t = 19021)]
         port: u16,
     },
+    Serial {
+        #[arg(
+            long,
+            env = "SERIAL_PORT",
+            default_value = "/dev/tty.usbserial-1132210"
```
jonathanpallant, line 74: Perhaps /dev/ttyUSB0 is a better default?

jonathanpallant:
This looks great - thank you for the PR.

Can I ask, what's your use-case for getting defmt data over a serial port?

print/src/main.rs
brightbloom: No problem, adjusting that now.

brightbloom:
Of course! I'm working on a low power system in which I am moving the chip in/out of various low power modes. This causes some issues with defmt over RTT when the RAM is powered down and seemingly random data is read back.

And, more importantly, the chip's SWD silicon consumes a significant amount of power on its own and can't be turned on/off easily during execution.

My use case is to redirect all defmt output to a debug UART on the chip, and use a custom defmt handler to manually enable/disable the clock to the peripheral before/after each write. This way, the UART is only consuming current when it is actively logging data and I can see the true current consumption of the system at all other times.

jonathanpallant (approved):

jonathanpallant:
LGTM. Thoughts @Urhengulas ?

PR #943, opened 2025-03-05
defmt-semihosting: use critical_section for synchronization.

driveraid:
This change allows `defmt-semihosting` to be used in multi-core contexts. It updates synchronization to match that of `defmt-rtt`.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -705,6 +705,8 @@ Initial release
 
 ### [defmt-semihosting-next]
 
+* [#943] use critical_section for synchronization.
+
 ### [defmt-semihosting-v0.1.0] (2024-11-27)
 
 Initial release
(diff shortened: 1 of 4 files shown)


jonathanpallant:
Thank you for the PR - I agree this kind of change needs doing.

Unfortunately `defmt_rtt`'s approach here isn't perfect. Instead of a `static mut` there should probably be an `UnsafeCell` and instead of three variables, there should probably be one static `struct Context` to keep it all tidy. I also want to switch to the `semihosting` crate, because it works on more architectures than just Cortex-M.

driveraid:
Do you want all of those in the same PR? Or is it perhaps better in the interest of keeping single-purpose PRs to let this PR merge and cover the other requests in follow-up PRs?

driveraid:
... I mean, since there's some refactoring to do (update `defmt-rtt` to do the same etc)

jonathanpallant (approved):

driveraid:
Draft PR up for one struct Context: https://github.com/knurling-rs/defmt/pull/944

PR #944, opened 2025-03-05
Refactoring: simplify use of global variables in backends.

driveraid:
This changes consolidates the three global variables used in the defmt backends into one structure.


driveraid:
Abandoning the idea since https://github.com/knurling-rs/defmt/pull/945.

Issue #942, opened 2025-02-26
Update defmt-semihosting


jonathanpallant:
The defmt-semihosting crate uses cortex-m-semihosting, however there's nothing cortex-m specific about it. It should use the semihosting crate which supports a wide range of architectures.

It should also stop using `static mut` and instead use `static STATE: critical_section::Mutex<RefCell<State>>`.

* [x] Use `critical-section`
* [ ] Switch to `semihosting` crate
* [ ] Use `UnsafeCell` (or something from `grounded`?) instead of `static mut`

jonathanpallant:
This would be useful for the Ferrous rust-training examples.

PR #945, opened 2025-03-05
Clean up defmt-semihosting.

jonathanpallant:
* Removed the static mut
* Switched to the semihosting crate, for Cortex-R support

Closes: #942

### DIFF
--- a/firmware/defmt-semihosting/src/lib.rs
+++ b/firmware/defmt-semihosting/src/lib.rs
@@ -1,17 +1,18 @@
 //! `defmt` global logger over semihosting
 //!
-//! NOTE this is meant to only be used with QEMU
-//!
-//! WARNING using `cortex_m_semihosting`'s `hprintln!` macro or `HStdout` API will corrupt `defmt`
-//! log frames so don't use those APIs.
+//! WARNING using `semihosting`'s `println!` macro or `Stdout` API will corrupt
+//! `defmt` log frames so don't use those APIs.
 //!
 //! # Critical section implementation
 //!
-//! This crate uses [`critical-section`](https://github.com/rust-embedded/critical-section) to ensure only one thread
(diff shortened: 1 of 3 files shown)


jonathanpallant:
cc @driveraid

firmware/defmt-semihosting/src/lib.rs
```diff
@@ -1,56 +1,142 @@
...
+        }
+    }
+
+    /// Acquire the defmt encoder.
+    fn acquire(&self) {
         // safety: Must be paired with corresponding call to release(), see below
         let restore = unsafe { critical_section::acquire() };
 
-        if TAKEN.load(Ordering::Relaxed) {
+        // NB: You can re-enter critical sections but we need to make sure
```
Urhengulas, line 66: What does `NB` stand for? non-blocking?

firmware/defmt-semihosting/src/lib.rs
```diff
@@ -1,56 +1,142 @@
...
+                    _ = h.write_all(b);
+                }
+            });
+        }
+    }
+}
 
-        // safety: accessing the `static mut` is OK because we have acquired a critical section.
-        unsafe { CS_RESTORE = restore };
+unsafe impl Sync for SemihostingEncoder {}
```
Urhengulas, line 135: Why does `SemihostingEncoder` need to be marked as `Sync`?

Urhengulas (approved):
lgtm, just two questions

firmware/defmt-semihosting/src/lib.rs
jonathanpallant: Because it's used in a `static` variable.

firmware/defmt-semihosting/src/lib.rs
jonathanpallant: [*Nota Bene*](https://en.wiktionary.org/wiki/nota_bene)

PR #949, opened 2025-03-27
Re-work defmt-rtt

jonathanpallant:
* Remove all static mut variables - now we use UnsafeCell, and we wrap all the state into a single static variable.
* Moved all the static variables to the top and put the implementation below.

Tested with the rust-exercises code for the nRF52-DK.

### DIFF
--- a/firmware/defmt-rtt/src/lib.rs
+++ b/firmware/defmt-rtt/src/lib.rs
@@ -61,7 +61,7 @@ const MODE_BLOCK_IF_FULL: usize = 2;
 const MODE_NON_BLOCKING_TRIM: usize = 1;
 
 /// The defmt global logger
-/// 
+///
 /// The defmt crate requires that this be a unit type, so our state is stored in
 /// [`RTT_ENCODER`] instead.
 #[defmt::global_logger]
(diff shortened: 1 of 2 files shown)


jonathanpallant:
The semver-check fail is only for defmt-parser, which should probably just be published.

firmware/defmt-rtt/src/lib.rs
```diff
@@ -39,76 +44,186 @@
...
+                _SEGGER_RTT.up_channel.write_all(b);
+            });
         }
+    }
 
-        // safety: accessing the atomic without CAS is OK because we have acquired a critical section.
-        TAKEN.store(false, Ordering::Relaxed);
+    /// Flush the encoder
+    unsafe fn flush(&self) {
+        if !self.taken.load(Ordering::Relaxed) {
```
Veykril, line 173: I don't think relaxed is correct here? Nothing here enforces an ordering relationship for this load if I see that right

Veykril: Ah wait nevermind, that's the unsafe part, we expect the lock to be acquired here I suppose

Veykril (approved):

firmware/defmt-rtt/src/lib.rs
Veykril: 👍

PR #956, opened 2025-04-02
Link LICENSE-* into defmt/ macros/ and parser/

luigix25:
I'm working on packaging defmt, defmt-macros and defmt-parser for Fedora and it's necessary to have the LICENSE files inside the crate. A link to the file is fine.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -52,6 +52,7 @@ We have several packages which live in this repository. Changes are tracked sepa
 
 ### [defmt-next]
 
+* [#956]: Link LICENSE-* in the crate folder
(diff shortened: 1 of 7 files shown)


jonathanpallant:
Looks ok to me.

Can I ask about the use case for having defmt packaged for Fedora? Are there applications built on Fedora that use defmt?

luigix25:
Thanks for the review! The plan is to package svsm :) `igvm` depends on `defmt`

jonathanpallant:
https://github.com/microsoft/igvm ?

luigix25:
yep! but my bad, `igvm` depends on `bitfield-struct` which then depends on `defmt` :)

jonathanpallant (approved):

PR #957, opened 2025-04-11
add `rust-toolchain.toml` file

vic1707:
Currently the CI jobs fails even on main's schedule
This is might be the known inconsistency problem, which is talked about in trybuild's [README](https://github.com/dtolnay/trybuild?tab=readme-ov-file#troubleshooting).
This change is the proposed fix from trybuild


jonathanpallant:
I think #958 solved this issue?

vic1707:
Looks like it, nice !
Let's close this for now ^^

PR #958, opened 2025-04-11
Update object crate to version 0.36

mcnicholls:
This fixes an issue I was having in probe-rs trying to flash a sealed binary for the RP Pico 2 (rp235x).

I'm using `picotool seal` to modify the binary to have the correct load_map headers, but then probe-rs seems
to have an issue with the resulting binary. It seems that it doesn't read the new ELF sections correctly.

Bumping object to 0.36 in defmt-decoder seems to fix this.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -585,7 +585,7 @@ Initial release
 
 ### [defmt-decoder-next]
 
-* No changes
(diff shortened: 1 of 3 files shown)


jonathanpallant (approved):

jonathanpallant:
Thanks!

PR #959, opened 2025-04-12
fix: missing "unstable-test" cfg in deftmt/tests/encode.rs

gibbz00:
File-wide because all tests in `encode.rs` use either the `check!` or `check_format!` macro, which in turn depends on the feature flagged `defmt::export::fetch_bytes()` function.

https://github.com/knurling-rs/defmt/blob/2c356885ee3b311c3dacff7fb57c01026725b0a3/defmt/src/export/mod.rs#L36-L40

https://github.com/knurling-rs/defmt/blob/2c356885ee3b311c3dacff7fb57c01026725b0a3/defmt/tests/encode.rs#L51-L74

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -52,6 +52,7 @@ We have several packages which live in this repository. Changes are tracked sepa
 
 ### [defmt-next]
 
+* [#959]: Missing "unstable-test" cfg in tests module
 * [#956]: Link LICENSE-* in the crate folder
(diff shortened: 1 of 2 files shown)


gibbz00:
Surely, failing UI tests can't be a result the cfg attribute change?

jonathanpallant:
Try rebasing after #958 goes in. I think this is because Rust changed its error output.

jonathanpallant (approved):

jonathanpallant:
Thanks!

PR #960, opened 2025-04-14
fix: Fix `Format` not accepting multiple helper attribute instances

Veykril:
Quick followup to https://github.com/knurling-rs/defmt/pull/937. I only meant to remove the unwraps but noticed that we regressed in functionality here regarding the derive helper


vic1707:
sorry about that 😓

Issue #964, opened 2025-05-09
Can't run `defmt-print --format=oneline`


jonathanpallant:
`defmt-print` assumes all `--format` options are custom log strings, and doesn't handle `online` or `default` as options.

You get an error about 'log format doesn't include `{s}`', which isn't particularly helpful.

PR #965, opened 2025-05-09
Modify defmt-print so it can handle 'online' as a log format.

jonathanpallant:
Before it was assuming the string was a custom log format. Now it uses the `FormatterFormat::from_string` function to check it for some well-known fixed formats, before treating it like a format string.

Fixes https://github.com/knurling-rs/defmt/issues/964


jonathanpallant:
Edit: oops - I broke it so it didn't support *normal* log strings. Now fixed and force pushed.

Veykril (approved):

PR #972, opened 2025-06-27
macros: Fix env_filter logic

jamesmunns:
A str can't be empty AND contain a separator, this is probably a logic bug, looking at the comment.


jonathanpallant:
https://github.com/knurling-rs/defmt/pull/973 is in the merge queue, so if you rebase in about 10 minutes, that should resolve the clippy issues.

jamesmunns:
@jonathanpallant all green!

PR #974, opened 2025-07-08
typecheck logs disabled by filter

de-vri-es:
This PR makes sure that disabled log statements are still typechecked.

Essentially, it just substitutes `false` for the log filter instead of outputting different code for build-time disabled logs.

I've done some quick verification in a sample project to ensure that the code is still completely removed from the binary: the size of the binary code did not change with or without this commit applied, tested both with the log statements enabled and disabled.

It will almost certainly increase compile times a little for projects with many normally disabled logs, but I believe the trade-off of type-checking disabled logs is worth it.

### DIFF
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -23,7 +23,7 @@ jobs:
           - windows-latest
         toolchain:
           - "stable"
-          - "1.76" # MSRV
+          - "1.81" # host MSRV
     runs-on: ${{ matrix.os }}
     timeout-minutes: 20
     steps:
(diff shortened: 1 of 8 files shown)


de-vri-es:
CI is now picking up an error in the `hints.md` book. I think this is a nice example why I think we need this :smile:

de-vri-es:
Added a UI test to verify that disabled logs are type checked, and run UI tests twice: once with `DEFMT_LOG=trace` and once with `DEFMT_LOG=off`.

de-vri-es:
Let me know if you think this needs more tests regarding binary size or compile times!

thejpster:
This is very interesting! I hope to find time to look at it soon.

de-vri-es:
Polite ping :)

Hoverbear (commented):
My initial impression is that this makes sense. Checking things even when the logs are disabled means the code stays correct, and that's always good.

I'm very glad you tested binary size impact, thank you. I believe @thejpster and I discussed spending a bit of time checking this a tad more before a merge.

This change doesn't appear to break any code that was not already subtly broken. I'm not sure how this interacts with the stability promises of https://ferrous-systems.com/blog/defmt-1-0-released/. Technically, making this change would only reveal already existing bugs in tested with some logs disabled, so we need to ask ourselves if our stability promise gives us permission to break code that already would break, provided the right build time arguments...

jonathanpallant:
Perhaps one path forward is to merge this, *plan* the next release, and do a blog post telling everyone what we're about to release. Then leave it some period of time for anyone to object, and we can then do the release with people having been warned.

Hoverbear:
Circling back, I did some testing with https://github.com/ferrous-systems/rust-exercises/tree/main/nrf52-code/radio-app and noted that the binary sizes did not change, debug or release, or because of `DEFMT_LOG` setting between this branch and `main`. So I think it's fine.

de-vri-es:
Thanks for testing! <3

jonathanpallant:
Thank you for this PR!

PR #978, opened 2025-07-29
add `impl Format for std::sync::Mutex`

rursprung:
this introduces a new optional feature `std` and an initial `Format` implementation just for `Mutex`. the hope is that over time more and more `Format` impls for `std` can be added for those who need it.

the trigger for this is that i broke the `embassy-sync` build for `std` by adding a `Format` derive to something which was using `Mutex` internally and it wasn't covered by CI there (which by now is rectified). in the ensuing conversation on Matrix it was stated that it might be nice to have `defmt` impls for `std` as well, even if only very few users would exist for this (i'm not one of them, which is how i ended up breaking CI).

### DIFF
--- a/defmt/src/impls/mod.rs
+++ b/defmt/src/impls/mod.rs
@@ -33,6 +33,8 @@ mod alloc_;
 mod arrays;
 mod core_;
 mod primitives;
+#[cfg(feature = "std")]
+mod std_;
 mod tuples;
 
 use defmt_macros::internp;
--- a/defmt/src/impls/std_.rs
+++ b/defmt/src/impls/std_.rs
@@ -1,0 +1,37 @@
+use super::*;
+use std::sync::TryLockError;
+
+impl<T> Format for std::sync::Mutex<T>
+where
+    T: ?Sized + Format,
+{
+    #[inline]
+    fn format(&self, fmt: Formatter) {
+        match self.try_lock() {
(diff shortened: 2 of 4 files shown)


defmt/src/impls/mod.rs
```diff
@@ -33,6 +33,8 @@ mod alloc_;
 mod arrays;
 mod core_;
 mod primitives;
+#[cfg(feature = "std")]
```
knoellle, line 36: You also need to disable the `no_std` in `defmt/src/lib.rs` if the `std` feature is enabled.

defmt/src/impls/std_.rs
```diff
@@ -0,0 +1,37 @@
...
+    delegate_format!(T, self, &*self);
+}
+
+#[cfg(test)]
+mod tests {
+    use crate as defmt;
+    use std::sync::Mutex;
+
+    #[test]
+    fn mutex() {
```
knoellle, line 33: This test is currently not executed in the CI.
Please add `std` to the feature sets in the `xtask test-host` `check` and `test` cargo invocations.

knoellle, line 33: Please also add a test for the `MutexGuard` implementation.

knoellle (changes requested):
Thank you for your contribution.

With [`defmt2log`](https://crates.io/crates/defmt2log) being already usable and #1029 on the horizon, I think adding gated support for `std` types makes sense now. @rursprung are you still interested in pushing this forward?

PR #985, opened 2025-08-29
set the _SEGGER_RTT address actively

fan2nd:
When using defmt-print tcp, we can get the address of _SEGGER_RTT directly from elf, and then pass it to the segger server through telnet. After testing, this can make the first line of log print out earlier.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -52,6 +52,7 @@ We have several packages which live in this repository. Changes are tracked sepa
 
 ### [defmt-next]
 
+* [#974]: Ensure typechecking is still performed on disabled log statement.
 * [#960]: Fix `Format` not accepting multiple helper attribute instances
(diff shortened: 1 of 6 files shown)


fan2nd:
Will someone help me fix the CI error?

jonathanpallant (commented):
Looks good - I'll try it out later.

Can you move the CHANGELOG to defmt-print-next?

jonathanpallant:
Looks good to me. I have in the past used defmt-print with OpenOCD's RTT server instead of J-Link's RTT server .... so can we make this optional? Just in case it upsets OpenOCD.

But otherwise good to go.

fan2nd:
You are right, openocd don't support this command.

jonathanpallant:
I did not mean to close this pull request, and I don't know how I did. I was trying to apply a change to your fork, and it told me the fork was out of date with respect to main. Let me open a PR and push these changes into our repo.

Issue #648, opened 2021-12-21
No way to print an hex byte slice compatible with both defmt+log


Dirbaio:
I'm using a [set of macros](https://github.com/embassy-rs/embassy/blob/master/embassy/src/fmt.rs) to forward log calls to `defmt` or `log` based on Cargo features, to make code portable between embedded and std platforms.

I can't find a way to print a hex byte slice that works both in `defmt` and `log`:

```rust
    let id: &[u8] = [0x12, 0x34, 0x56, 0x78];

    // defmt fails with `unknown display hint: "02x?"`
    info!("card id = {:02x?}", id);

    // log fails with `the trait bound `[u8]: LowerHex` is not satisfied`
    info!("card id = {:02x}", id);
```

I'm not sure if it's an explicit goal to make `defmt` log syntax a superset of `std`'s formatting, but it would be very nice if it was, as it allows porting code super easily. As it stands now, it almost is :)

Issue #149, opened 2020-08-29
Compatibility with for `log` logging macros?


birkenfeld:
While defmt is shaping up great for embedded logging, the nice thing about the normal `log` macros is that they can be used in code that's used both in embedded and non-embedded contexts, such as:

* Nontrivial libraries like `smoltcp` that are used in both contexts, where high-bandwidth logging is desirable in both cases.

* Parts of a firmware that are used non-embedded for testing and simulation purposes.  It can be quite time saving to extract the hardware independent parts into a library crate, where they can be tested on normal CI infrastructure, and used to simulate the firmware's behavior.

In both cases, using `defmt` would make this cumbersome, needing to make the logging conditional somehow.

I'm opening this issue for a discussion if it is possible to provide, as a part of `defmt`, a shim that would keep logging usable for both contexts, with acceptable restrictions.  The first thing coming to mind is a macro that makes a transformation like this:

```
debug!("{:u16} and {:str}", a, b);

=>

#[cfg(not(feature = "defmt-to-log"))]
defmt::debug!("{:u16} and {:str}", a, b);
#[cfg(feature = "defmt-to-log")]
log::debug!("{} and {}", a, b);
```

i.e. the format string is rewritten for `log` and the two versions emitted side by side.  Of course this can't expect to deliver the same output for all cases (bitfields come to mind), but IMO a best effort would make life much easier for the overwhelming majority of cases.

jonas-schievink:
I believe we now handle the common cases with https://github.com/knurling-rs/defmt/issues/270. Whether we want to support the entire `core::fmt` grammar is a different question, but this should already help a lot when trying to use both `log` and `defmt`, or when migrating.

huntc:
It’d be great to have the user doc indicate any existing compatibility with the log crate, and perhaps an example. Compatibility was my first question. Thanks.

algesten:
I'm new to embedded programming (trying out an stm32f10 nucleo board). `probe-run` seems like a great idea, and it appeared `defmt` goes hand-in-hand. 

However I stumbled straight away on that `defmt` has no support for the `log` crate. I tried using the equivalent log macros like `defmt::info!()`, but they are different in that they don't support all formatting options I'm used to (I have structs implementing a mix of `Display` and `Debug`) and get errors like `unknown display hint: ".02"`.

I tried working around the problem by implement the glue myself:

```rs
use log::{Level, Log, Metadata, Record};

[...]

PR #987, opened 2025-09-06
Add limited support for fill/align.

iwanders:
Hi! This add limited support for the [Fill/Alignment](https://doc.rust-lang.org/std/fmt/index.html#fillalignment) in the formatting specification, it is fully backwards compatible and currently ensures that only right aligns and `0` is used as a fill character. Ideally, we'd modify the `DisplayHint` struct to support the full [format syntax](https://doc.rust-lang.org/std/fmt/index.html#syntax), but that's a breaking change, this adds support for the fill/align values that can be handled without a breaking change.

For context; I developed a library I'm working on with both `log` and `defmt` as optional. I've always used the `{:0>8x}` notation to get padding, instead of `{:08}`, because I find it easier to interpret without two digits together with different meaning. When I finally switched from `log` to `defmt` to run on bare metal, I encountered a significant number of compilation errors on this, and I thought I'd try to help improve defmt.

But now that I'm writing this PR, I'm actually not sure if this is a good idea, because this may make people think that because `{:0>8x}` is supported, that `{: >8x}` should also work, and we can't make that work without a breaking change as far as I can see. I'll file the PR for discussion, but I'm having second thoughts whether this is a good change, so absolutely fine if we end up closing this.

### DIFF
--- a/parser/src/display_hint.rs
+++ b/parser/src/display_hint.rs
@@ -56,10 +56,78 @@ pub enum DisplayHint {
     Unknown(String),
 }
 
+// https://github.com/rust-lang/rust/blob/99317ef14d0be42fa4039eea7c5ce50cb4e9aee7/library/core/src/fmt/mod.rs#L25
+#[derive(Copy, Clone, Debug, PartialEq, Eq)]
+enum Alignment {
+    /// Indication that contents should be left-aligned.
+    Left,
+
+    /// Indication that contents should be right-aligned.
+    Right,
+
+    /// Indication that contents should be center-aligned.
+    Center,
+}
+impl std::convert::TryFrom<char> for Alignment {
+    type Error = ();
+
+    fn try_from(value: char) -> Result<Self, Self::Error> {
+        match value {
+            '<' => Ok(Alignment::Left),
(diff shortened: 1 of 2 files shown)


iwanders:
Actually, going through my compilation errors I discovered there's more differences between the logging than I expected, for example;
```rust
warn!("Raw write data {:x?}", event.data());
```
But doesn't work with defmt; `error: unknown display hint: "x?"`. Which is unfortunate because even removing the `?` won't make things work on both `log` and `defmt` because Display is missing on a slice.

I assumed the formatting syntaxes to the more similar, but admittedly I assumed this instead of reading the documentation. I found #648 which captures the exact problem I ran into this morning while trying to fix my printing, and #149 also seems applicable, I'll watch those.

This reinforces my belief that this may not be a good change because it only works for some formats. Perhaps a better approach is a feature where the parser can be switched to a best effort mode, where it maps the log/std type format specification to the ones defmt supports and silently ignores things if it doesn't. I'm going to close this for now, if the maintainers believe it is a worthwhile change we can reopen this.

For the record; I am a happy user of defmt, I push the log data through a serial port and use `defmt-print` to consume it :+1:

PR #988, opened 2025-09-10
Add '--set-addr' to defmt-print

jonathanpallant:
A second attempt at https://github.com/knurling-rs/defmt/pull/985

Tested with SEGGER Ozone.

### DIFF
--- a/print/src/main.rs
+++ b/print/src/main.rs
@@ -44,10 +44,6 @@ struct Opts {
     #[arg(long)]
     host_log_format: Option<String>,
 
-    /// Tell Segger J-Link what the RTT address is
-    #[arg(long)]
-    set_addr: bool,
-
     /// Log any malformed defmt frames that are being skipped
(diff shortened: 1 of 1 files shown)


print/src/main.rs
```diff
@@ -106,6 +121,27 @@ impl Source {
         Ok(Source::Serial(ser))
     }
 
+    async fn set_rtt_addr(&mut self, elf_bytes: &[u8]) -> anyhow::Result<()> {
+        let Source::Tcp(tcpstream) = self else {
```
Urhengulas, line 125: Since passing --set-addr only makes sense with tcp we could warn if it is used with serial. Or we could move it into Command::Tcp so it is only available with tcp.

Urhengulas (approved):

print/src/main.rs
jonathanpallant: Good point - fixed.

PR #1005, opened 2025-10-28
Fix available_buffer_size function

Oveln:
In the original implementation, if read < write, the available space between 0 and read would be ignored.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -962,6 +962,7 @@ Initial release
 ---
 
 [#1007]: https://github.com/knurling-rs/defmt/pull/1007
+[#1006]: https://github.com/knurling-rs/defmt/pull/1006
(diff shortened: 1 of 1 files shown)


jonathanpallant:
I'm staring at this and struggling to convince myself that either what we had is correct, or what you propose is correct.

Perhaps we could use some unit tests for that function, covering the various conditions?

Oveln:
Oh... Now this seems like there's something wrong with the testing tool?

jonathanpallant:
The changes here to the algorithm look OK, as do the tests.

Please don't do the version bump though. We'll collect a bunch of changes on the main branch and then do a release PR that combines them together in one bump.

Oveln:
Okay, but how should I modify the changelog?

jonathanpallant (approved):

Issue #1000, opened 2025-10-12
Missing Cargo.lock


N-T-K:
[Why have Cargo.lock in version control?](https://doc.rust-lang.org/cargo/faq.html#why-have-cargolock-in-version-control)

N-T-K:
@Hoverbear I think there's some confusion here. The root workspace (https://github.com/knurling-rs/defmt/blob/main/Cargo.toml) defines multiple binary crates. The link you provided speaks to the philosophy for libraries. If I want to deterministically package `print` (https://github.com/knurling-rs/defmt/blob/main/print/Cargo.toml) I cannot do so without including a lockfile.

PR #1011, opened 2025-11-06
Resolves Missing Cargo.lock #1000

N-T-K:
Fix for the missing Cargo.lock that's suggested per #1000


jonathanpallant (approved):

jonathanpallant:
CI passes, and if I build packages locally the packages pick up the contents of the workspace `Cargo.lock` file (minus the dependencies that particular binary does not require).

I might re-jig the CHANGELOG entry, but that can be another PR.

PR #1012, opened 2025-11-07
Make `qemu-run` suitable for external use

jonathanpallant:
Updates the `qemu-run` binary so we can publish it and use it elsewhere.

### DIFF
--- a/qemu-run/README.md
+++ b/qemu-run/README.md
@@ -22,7 +22,7 @@ Run `qemu-run --help` to see a list of other command-line arguments available.
 
 ## Support
 
-`defmt-print` is part of the [Knurling] project, [Ferrous Systems]' effort at
+`qemu-run` is part of the [Knurling] project, [Ferrous Systems]' effort at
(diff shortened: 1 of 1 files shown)


jonathanpallant:
Fixed formatting, and added a CHANGELOG section for it.

robamu:
> Did we not pin down the version of mdbook we use? We should do that.

sounds like a good thing to do.

This tool could be useful for some of the stuff I did in embedded-alloc as well..

mhatzl (changes requested):
the qemu-run code looks mostly good.
some copy-paste errors and proposed a better clap structure.

when trying to run example binaries in firmware/qemu e.g. `cargo rb log`, I got

```
INFO  Bound UART data socket to 127.0.0.1:41553 (HOST)
Timer with period zero, disabling
ERROR panicked at 'The answer is 42'
```

Not sure if I have a bad QEMU version (6.2.0), or am missing other cfgs.

jonathanpallant:
Github is broken.

PR #1014, opened 2025-11-10
Switch to 32-bit types as specified by SEGGER

### LINKED ISSUE #1013: RTT structure is incorrect for 64-bit targets
The in-memory structures currently used by `defmt` for RTT are incompatible with the reference code when run in 64-bit platforms.

For example, the header uses `usize` as the datatype when defining the maximum number of up- and down-channels:

https://github.com/knurling-rs/defmt/blob/ce8c280ba3ca86a9a2ac69e75a27220417d1e7c2/firmware/defmt-rtt/src/lib.rs#L250-L251

However, the upstream library code uses `int`, which is equivalent to `u32` on all modern platforms: https://github.com/SEGGERMicro/RTT/blob/ad6970d813bb12b8a5e34aa94b3a1999f7bd0b6b/RTT/SEGGER_RTT.h#L353-L354

Additionally, all the fields in `Channel` are defined as `usize` or `AtomicUsize`:

https://github.com/knurling-rs/defmt/blob/ce8c280ba3ca86a9a2ac69e75a27220417d1e7c2/firmware/defmt-rtt/src/channel.rs#L14-L22

However, the official implementation uses `unsigned` as the type, which is also equivalent to a `u32` on all tested platforms: https://github.com/SEGGERMicro/RTT/blob/ad6970d813bb12b8a5e34aa94b3a1999f7bd0b6b/RTT/SEGGER_RTT.h#L340-L343

jonathanpallant:
Switches to types that more closely match the `unsigned` and `int` types used by SEGGER in the reference implementation, even on 64-bit targets.

Closes #1013

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -730,6 +730,7 @@ Initial release
 ### [defmt-rtt-next]
 
 * [#1006] Fix `available_buffer_size` ignoring available buffer space when `read < write`
+* [#1013] Switch to fixed 32-bit types, to support 64-bit targets
 
 ### [defmt-rtt-v1.1.0] (2025-10-09)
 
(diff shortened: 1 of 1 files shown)


jonathanpallant:
@Urhengulas did you have a chance to test this change?

jonathanpallant:
I would leave BUF_SIZE alone, because having a constant that tells you the size of things but that isn't a `usize` usually involves a lot of annoying casts.

PR #1029, opened 2026-01-21
Linux and mac compatibility

aimir:
Currently without correct locations in symbol tables on mac, but everything else should work as expected, with no impact on embedded systems


jonathanpallant:
Nice - I'm excited to look through this in detail soon.

macros/src/construct.rs
```diff
@@ -69,7 +69,13 @@ pub(crate) fn linker_section(for_macos: bool, prefix: Option<&str>, symbol: &str
     };
 
     if for_macos {
-        sub_section = format!(",{:x}", hash(&sub_section));
+        // Use a single section per severity level instead of unique section per log.
```
jonathanpallant, line 72: I wonder if this has an effect on the linker's ability to garbage collect defmt strings in functions that are not called?

defmt/src/export/mod.rs
```diff
@@ -98,9 +99,61 @@ pub fn timestamp(fmt: crate::Formatter<'_>) {
...
+
+/// For Linux (ELF), use the linker-provided `__executable_start` symbol.
+#[cfg(target_os = "linux")]
+#[allow(unused_unsafe)]
+fn binary_base() -> u16 {
+    extern "C" {
+        static __executable_start: u8;
+    }
+    // SAFETY: `__executable_start` is a linker-provided symbol marking the start of the executable.
+    (unsafe { core::ptr::addr_of!(__executable_start) as usize }) as u16
```
jonathanpallant, line 116: I think `core::ptr::addr_of!` is safe, so the `unsafe` keyword (and the `allow(unused_unsafe)` attribute) are not required?

jonathanpallant:
I had a trial run with this, but `defmt-print` cannot open Mach-O files, so this isn't usable on macOS.

How have you been testing it?

tomasbarton-stemcell:
I tested this on Linux and it works well 🥳 

Although I still needed to have a custom linker file that collects all interned strings into a single section:

```
Error: defmt version found, but no `.defmt` section - check your linker configuration
```

Ideally the elf parser could deal with it without the linker file, because the symbols are in the elf file when I print them out.

Alternatively modify the provided `defmt.x` so that it does `INSERT AFTER ...` (somehow, maybe behind a feature flag?) otherwise it cannot be used as by default it replaces the default linker file instead of extending it.

defmt/src/export/mod.rs
aimir: It's considered safe since rust `1.82`, so omitting the `unsafe` would break compatibility with rust 1.81

aimir: hence the `unused_unsafe` too

jonathanpallant:
I think the best solution is to only put items into the `.defmt` section when running bare-metal (when `cfg(target_os = "none")`). Otherwise let linker just put them wherever. And then fix the decoder to be more flexible about where it looks for these symbols.

aimir:
> I had a trial run with this, but `defmt-print` cannot open Mach-O files, so this isn't usable on macOS.
> 
> How have you been testing it?

I just used the decoder directly, but let me go ahead and change this - added a commit to support Mach-O files in `defmt-print` 😁

jonathanpallant:
@knoellle is going to take a look at this one, right after his Robocup competition.

knoellle:
Thank you for your work so far.
I was able to craft a native Linux program and a linker script that worked but don't have a Mac to test with.

What is your plan with the linker scripts? As @tomasbarton-stemcell mentioned already, I think we should either include linker scripts for Linux/Mac or adjust the elf parser to work without linker scripts.
The `defmt/build.rs` already generates/modifies linker scripts depending enabled features. Using a different linker script as a base depending on the platform seems reasonable to me, though not needing one would be preferable I think.

macros/src/construct.rs
knoellle: Would it do much harm if it didn't?
There are some very flash-constrained devices running Linux, but I'm not aware of tiny embedded Mac systems and non-mac targets should be unaffected.

mhatzl:
Thanks for the PR!
Tested it on macOS and it worked without the need for a linker script.
The encoded defmt logs are output on stdout and piping it to `defmt-print` given the binary correctly decoded and displayed the logs.

As mentioned in the PR description, the location information currently does not work for macOS.
@aimir do you have an idea why or maybe even time to fix that?

aimir:
> Thank you for your work so far. I was able to craft a native Linux program and a linker script that worked but don't have a Mac to test with.
> 
> What is your plan with the linker scripts? As @tomasbarton-stemcell mentioned already, I think we should either include linker scripts for Linux/Mac or adjust the elf parser to work without linker scripts. The `defmt/build.rs` already generates/modifies linker scripts depending enabled features. Using a different linker script as a base depending on the platform seems reasonable to me, though not needing one would be preferable I think.

Mac doesn't have linker scripts, and like @mhatzl said it should work without the need for one.

It does mean that things like the default panic handler that's auto-configured by `defmt/build.rs`. I didn't find an elegant workaround yet, and I really want to merge this - is this a must for you?

aimir:
> Thanks for the PR! Tested it on macOS and it worked without the need for a linker script. The encoded defmt logs are output on stdout and piping it to `defmt-print` given the binary correctly decoded and displayed the logs.
> 
> As mentioned in the PR description, the location information currently does not work for macOS. @aimir do you have an idea why or maybe even time to fix that?

The issue is that by default mac location information is scattered across files, and the table parsing doesn't handle this well.

Running without split debug info (`RUSTFLAGS="-C split-debuginfo=off`) seems to resolve the issue - @mhatzl can you check if that works for you?

mhatzl:
Hi @aimir,
thanks for your help!
Unfortunately I am not able to get correct location infos on macOS with `split-debuginfo=off`.

looking at the [docs](https://doc.rust-lang.org/rustc/codegen-options/index.html#split-debuginfo) for split-debuginfo and the [cargo profile docs](https://doc.rust-lang.org/cargo/reference/profiles.html#split-debuginfo), the default for macOS is `packed` except for Cargo profiles that have debug infos enabled it is `unpacked`.

Since the `dev` profile has debug infos enabled, split-debuginfo is set to `unpacked` by default when run via Cargo. Setting it to `packed` creates this `dSym` folder containing DWARF data, but the location info is still missing if this DWARF file is passed to `defmt-print`.

knoellle:
> It does mean that things like the default panic handler that's auto-configured by `defmt/build.rs`. I didn't find an elegant workaround yet, and I really want to merge this - is this a must for you?

I spent some time attempted making the default panic stuff work without linker scripts but didn't find a way without breaking user code. See #1050.

Again, I don't have a mac to test with, but if I understand correctly then attempting to use `defmt::panic` on macOS shouldn't successfully link or run (still not entirely sure how that works over there) unless the user defines a custom defmt panic handler.
`defmt::panic` calls `_defmt_panic`, which is weakly linked to `__defmt_default_panic` but can be overridden using `the defmt::panic_handler` macro.
Without the linker script and no custom implementation, `_defmt_panic` would be an undefined symbol.

What are your thoughts on requiring a custom panic handler for building for macOS specifically?

knoellle:
I'm curious how you tested the linux compatibility.
I have been able to get it to work, but I haven't found a good way of passing the linker script to the linker. My workaround so far was setting `RUSTFLAGS="-C link-arg=-T<path to my repo>/.cargo/defmt.x"`

For embedded platforms, we use flip-link which does a bunch of [magic to find linker scripts](https://github.com/knurling-rs/flip-link/blob/5313058211ee93c001d2b98c49d2f866c318c559/src/main.rs#L240-L269) inside library search paths which other linkers don't do.
So far I have always had to pass an absolute path to my linker script to compile my linux binaries.

- defmt build.rs generates the default linker script but doesn't know if a linker script is going to be used at all.
- end user `.cargo/config.toml` shouldn't contain an absolute path to a file in the repo or worse to the one generated by defmt build.rs
- Requiring a build.rs in the end user project to find either the defmt-generated linker script (somewhere in target/, maybe requires the defmt build.rs to somehow pass it's own OUT_DIR to the end user build.rs) or one provided by the user project and adding the rustflag that way seems very heavy handed.

How do you build your linux binaries?

aimir:
@lakechurch suggested a fix for the symbol table collection that seems to work well - thanks!!

@mhatzl can you take a look now?

aimir:
@knoellle what I ended up doing is using `build.rs` to find the relative path to the linker script, which is definitely not ideal

mhatzl:
@aimir location information now works on macOS. Thank you!

decoder/src/elf2table/mod.rs
```diff
@@ -314,6 +315,23 @@ pub fn get_locations(elf: &[u8], table: &Table) -> Result<Locations, anyhow::Err
     let mut units = dwarf.debug_info.units();
 
     let mut map = BTreeMap::new();
+
+    // macos groups log sections, so we map the secondary location marker's hash to its table index
+    let mut hash_to_index = std::collections::BTreeMap::new();
+    if is_mac {
+        use std::hash::{Hash, Hasher};
```
knoellle: What is your reason for keeping this use statement local in this scope? We don't do that anywhere else.

decoder/src/elf2table/mod.rs
```diff
@@ -78,12 +84,33 @@ pub fn parse_impl(elf: &[u8], check_version: bool) -> Result<Option<Table>, anyh
         }
     }
 
+    let mut defmt_sections = HashMap::new();
+    if is_mac {
+        let defmt_segment = elf.segments().find(|segment| {
+            object::ObjectSegment::name(segment).is_ok_and(|name| name == Some(".defmt"))
```
knoellle:
```suggestion
            ObjectSegment::name(segment).is_ok_and(|name| name == Some(".defmt"))
```

knoellle:
```suggestion
            object::ObjectSegment::name(segment) == Ok(Some(".defmt"))
```

decoder/src/elf2table/mod.rs
```diff
@@ -384,33 +402,44 @@ pub fn get_locations(elf: &[u8], table: &Table) -> Result<Locations, anyhow::Err
...
+                            if let Ok(hash) = u64::from_str_radix(hash_str, 16) {
+                                if let Some(&index) = hash_to_index.get(&hash) {
+                                    let file = file_index_to_path(file_index, &unit, &dwarf)?;
+                                    let module = segments.join("::");
+                                    map.insert(index, Location { file, line, module });
+                                }
+                            }
+                        }
+                    } else {
+                        // --- STANDARD ELF LOGIC (Preserved) ---
```
knoellle: This comment seems to document the diff I'm currently staring at just as much as what the actually code does.

macros/src/construct.rs
```diff
@@ -91,11 +91,28 @@ pub(crate) fn static_variable(
     let section = linker_section(false, prefix, &sym_name);
     let section_for_macos = linker_section(true, prefix, &sym_name);
 
+    // macos restricts segments to 255 sections. we group logs into a single section and emit
+    // a secondary location marker symbol to allow the decoder to resolve addresses via DWARF.
+    use std::collections::hash_map::DefaultHasher;
+    use std::hash::{Hash, Hasher};
```
knoellle: Same here, please avoid local uses.

print/src/main.rs
```diff
@@ -239,9 +238,35 @@ async fn run(opts: Opts, source: &mut Source) -> anyhow::Result<()> {
...
+    // split-debuginfo = "packed". On other platforms DWARF is embedded in the
+    // main ELF, so we just feed `bytes` straight through.
+    #[cfg(not(target_os = "macos"))]
+    let dwarf_bytes = bytes.clone();
+
+    #[cfg(target_os = "macos")]
+    let dwarf_bytes = {
+        let mut dwarf_bytes = bytes.clone();
+        let dsym_bundle = format!("{}.dSYM", elf_path.display());
+        let dwarf_dir = std::path::Path::new(&dsym_bundle).join("Contents/Resources/DWARF");
```
knoellle:
```suggestion
        let dwarf_dir = Path::new(&dsym_bundle).join("Contents/Resources/DWARF");
```

print/src/main.rs
```diff
@@ -239,9 +238,35 @@ async fn run(opts: Opts, source: &mut Source) -> anyhow::Result<()> {
...
+
+        // Attempt to read the first DWARF file within the bundle directory.
+        if let Ok(mut entries) = std::fs::read_dir(dwarf_dir) {
+            if let Some(Ok(entry)) = entries.next() {
+                // Note: defmt-print uses async tokio::fs::read
+                if let Ok(bundle_bytes) = fs::read(entry.path()).await {
+                    dwarf_bytes = bundle_bytes;
+                }
+            }
+        }
```
knoellle: Any particular reason why you chose sync io for the read_dir but async io for reading the file?

print/src/main.rs
```diff
@@ -238,9 +238,35 @@ async fn run(opts: Opts, source: &mut Source) -> anyhow::Result<()> {
...
+
+    // On macOS, DWARF lives in a sibling .dSYM bundle when built with
+    // split-debuginfo = "packed". On other platforms DWARF is embedded in the
+    // main ELF, so we just feed `bytes` straight through.
+    #[cfg(not(target_os = "macos"))]
+    let dwarf_bytes = bytes.clone();
+
+    #[cfg(target_os = "macos")]
+    let dwarf_bytes = {
+        let mut dwarf_bytes = bytes.clone();
```
knoellle: This cloned data is only used when the discovery of sibling DWARF data fails.
Does it even make sense to return the original bytes in that case?

knoellle (changes requested):

print/src/main.rs
aimir: Agree - what I did instead (in 70b625460bec607ecb07c1d6322b987b292fb7e4) was return `locs` from both the linux and macos flows, and have it be optional - which in the failure case simply warns of an issue and gives back `None`.

print/src/main.rs
aimir: Done in 70b625460bec607ecb07c1d6322b987b292fb7e4 (moved this logic to a helper `read_dsym_dwarf` function, in which I use `Path` like you wanted)

print/src/main.rs
aimir: Definitely not a good one 😂

Changed to read using the already-used `tokio::fs` in 70b625460bec607ecb07c1d6322b987b292fb7e4 (also part of the `read_dsym_dwarf` helper function)

macros/src/construct.rs
aimir: Done (24903778409291feb82fb813971186d755fe01ab)

decoder/src/elf2table/mod.rs
aimir: Fixed (24903778409291feb82fb813971186d755fe01ab)

decoder/src/elf2table/mod.rs
aimir: Agreed. Removed it in d370c7584673539dd6d8f88de84ef0d820685c02 - this also allowed me to collapse the `else-if-let` to a single line

decoder/src/elf2table/mod.rs
aimir: Done (163c52ac02328fa056fe052c46b35103d42d2584)

aimir: Done (163c52ac02328fa056fe052c46b35103d42d2584)

aimir:
@knoellle please let me know if this sufficiently addresses your comments

print/src/main.rs
```diff
@@ -354,6 +348,15 @@ async fn run(opts: Opts, source: &mut Source) -> anyhow::Result<()> {
     }
 }
 
+#[cfg(target_os = "macos")]
+async fn read_dsym_dwarf(elf_path: &Path) -> Option<Vec<u8>> {
+    let dsym_bundle = format!("{}.dSYM", elf_path.display());
+    let dwarf_dir = Path::new(&dsym_bundle).join("Contents/Resources/DWARF");
```
knoellle: Can you use [`with_(added)_extension`](https://doc.rust-lang.org/std/path/struct.Path.html#method.with_extension) here?

aimir: `with_added_extension` is the correct behavior that matches `dsymutil`'s logic, but it wasn't stabilized until rust 1.91 so it will break rust 1.83 backcompat

aimir: @knoellle how would you like me to proceed on this?

knoellle:
I recently stumbled upon this project: https://crates.io/crates/defmt2log
Like this PR, it also implements linux-compatible elf-parsing, but unlike here it directly integrates decoding the generated log frames and printing them. It also doesn't rely on a linker script, something I had also been working on for a while.
I can see a use case for either option (sending the defmt stream through a constricted channel or printing directly to the terminal) and I would like to support both of them. However `defmt2log` doesn't work with your changes and I'm not quite sure why.

@aimir @jordens I am open to suggestions for how to nicely integrate the work of both of you. Perhaps the linkerscript-less elf-parsing from defmt2log could be upstreamed such that both linux and mac native builds have built-in support but are agnostic w.r.t. the actual logger implementation?

print/src/main.rs
knoellle: I don't think this is worth bumping MSRV over, resolving.

jordens:
Sure. It'll probably be a while until I get around to touching this again.
But feel free to grab the ELF parsing. Then the synthetic intermediate ELF can go away. 

The one wrinkle of parsing without a linker script is less control over the addresses and thus the reduced u16 indices which then have a different collision risk/scenario. Also I didn't figure out how to easily keep the bitfield info from getting discarded.

aimir:
If @jordens isn't super available, I think I'd be faster to merge this as-is, with the extra control provided by a linker script.

Integrating the `defmt2log` elf-parsing is very useful but tricky, this PR has a lot of logic even as it currently is, and it works well enough on its own.

aimir:
@knoellle what do you think? I opened this PR 4 months ago - I want to understand how to move forward with it

jonathanpallant:
Apologies for the delay - this is a big PR that touches deep into a library we *really* have to be careful about breaking because of the wide audience. The PR also contains some LLM generated code, and that required some internal discussion that was not visible here.

I'm checking out the PR now and going through it.

jonathanpallant:
OK, some testing.

I wrote this program:

```rust
use std::io::Write;
use std::time::Duration;
use std::thread::sleep;

const INTERVAL: Duration = Duration::from_secs(1);

fn main() {
    for x in 0..5 {
        defmt::println!("Hello, x = {=u32}", x);
        sleep(INTERVAL);
    }
    defmt::error!("Not an error - we're done");
}

#[inline(never)]
#[unsafe(no_mangle)]
unsafe fn _defmt_acquire()  {
    eprintln!("Acq!!");
}

#[inline(never)]
#[unsafe(no_mangle)]
unsafe fn _defmt_flush()  {
    eprintln!("Flush");
    let mut stdout = std::io::stdout();
    stdout.flush().unwrap();
}

#[inline(never)]
#[unsafe(no_mangle)]
unsafe fn _defmt_release()  {
    eprintln!("Rel!!");
}

#[inline(never)]
#[unsafe(no_mangle)]
unsafe fn _defmt_write(bytes: &[u8])  {
    eprintln!("Send: {:02x?}", bytes);
    let mut stdout = std::io::stdout();
    stdout.write_all(bytes).unwrap();
}

#[unsafe(export_name = "_defmt_timestamp")]
#[inline(never)]
fn defmt_timestamp(_fmt: defmt::Formatter<'_>) {

}
```

This program compiles and runs on Linux (aarch64-unknown-linux-gnu) and macOS 26 (aarch64-apple-darwin).

Linux:

```console
root@beaac92a9290:/work/defmt-checking# cargo run > linux.log
   Compiling proc-macro2 v1.0.101
   Compiling quote v1.0.41
   Compiling thiserror v2.0.17
   Compiling defmt-macros v1.1.0 (/work/defmt/macros)
   Compiling defmt v1.1.0 (/work/defmt/defmt)
   Compiling syn v2.0.106
   Compiling thiserror-impl v2.0.17
   Compiling defmt-parser v1.0.0 (/work/defmt/parser)
   Compiling defmt-checking v0.1.0 (/work/defmt-checking)
    Finished `dev` profile [unoptimized + debuginfo] target(s) in 3.99s
     Running `target/debug/defmt-checking`
Acq!!
Send: [8b, 39]
Send: [00, 00, 00, 00]
Rel!!
Acq!!
Send: [8b, 39]
Send: [01, 00, 00, 00]
Rel!!
Acq!!
Send: [8b, 39]
Send: [02, 00, 00, 00]
Rel!!
Acq!!
Send: [8b, 39]
Send: [03, 00, 00, 00]
Rel!!
Acq!!
Send: [8b, 39]
Send: [04, 00, 00, 00]
Rel!!
Acq!!
Send: [8a, 39]
Rel!!
root@beaac92a9290:/work/defmt-checking# hexdump -C linux.log
00000000  8b 39 00 00 00 00 8b 39  01 00 00 00 8b 39 02 00  |.9.....9.....9..|
00000010  00 00 8b 39 03 00 00 00  8b 39 04 00 00 00 8a 39  |...9.....9.....9|
00000020
```

macOS:

```console
$ cargo run > macos.log
    Finished `dev` profile [unoptimized + debuginfo] target(s) in 0.10s
     Running `target/debug/defmt-checking`
Acq!!
Send: [01, 00]
Send: [00, 00, 00, 00]
Rel!!
Acq!!
Send: [01, 00]
Send: [01, 00, 00, 00]
Rel!!
Acq!!
Send: [01, 00]
Send: [02, 00, 00, 00]
Rel!!
Acq!!
Send: [01, 00]
Send: [03, 00, 00, 00]
Rel!!
Acq!!
Send: [01, 00]
Send: [04, 00, 00, 00]
Rel!!
Acq!!
Send: [00, 00]
Rel!!
$ hexdump -C macos.log
00000000  01 00 00 00 00 00 01 00  01 00 00 00 01 00 02 00  |................|
00000010  00 00 01 00 03 00 00 00  01 00 04 00 00 00 00 00  |................|
00000020
```

I can decode the data on macOS:

```console
$ cat macos.log | defmt-print -e ./target/debug/defmt-checking
Hello, x = 0
Hello, x = 1
Hello, x = 2
Hello, x = 3
Hello, x = 4
ERROR Not an error - we're done
```

However I cannot decode the log on Linux:

```console
root@beaac92a9290:/work/defmt-checking# cat linux.log | defmt-print -e ./target/debug/defmt-checking
Error: defmt version found, but no `.defmt` section - check your linker configuration
```

I also compiled and ran some bare-metal firmware for the Nordic nRF52-DK, and that seemed to work OK, with all the defmt sections I expect.

jonathanpallant:
If I run this on macOS:

```sh
dsymutil ./target/debug/defmt-checking
```

then I get location info:

```console
$ cat macos.log | defmt-print -e ./target/debug/defmt-checking --format=online
Hello, x = 0
Hello, x = 1
Hello, x = 2
Hello, x = 3
Hello, x = 4
[ERROR] Not an error - we're done (defmt_checking src/main.rs:12)
```

jonathanpallant:
OK, some TODOs:

* [ ] Squash all the commits into one (or a handful) or logical changes - perhaps one for the macro (to use macOS compatible sections) and one for the decoder (to use object, and handle dSYMs)
* [ ] Investigate what's wrong with defmt-print on the Linux binary
* [ ] Add a CHANGELOG entry
* [ ] Add a libstd defmt transport (e.g. that prints to stdout, or a file given as an environment variable)
* [ ] Update the docs to note that you can now run unit tests on macOS and Linux, and show how to do that (including the dSYM trick)
* [ ] Rename the `elf` variable to `executable` or something, because it might be a Mach-O file not an ELF file.

aimir:
Sorry it took me so long to respond!

The linux issue should now be fixed. Host builds don't link with `defmt.x`, so nothing merged the per-statement `.defmt.<tag>.<symbol>` input sections into the single `.defmt` section the decoder looked for. Linux host binaries had defmt data but no `.defmt` section, and decoding bailed. The fix makes the decoder collect all `.defmt*` sections on ELF (and truncate DWARF addresses to 16 bits to match how table indices are computed), so binaries linked without the script decode correctly, locations included. MacOS was unaffected, it uses a different format because of the 16-char length limit on section names.

The `libstd` transport - I added a `defmt-stdout` submodule + explained how to do this as part of `setup.md` - is that what you had in mind? In general I tried to update `setup.md` with info on running/testing on linux and mac - let me know what you think @jonathanpallant.

Addressed the rest of the issues - let me know if any further fixes are needed.

aimir:
@jonathanpallant the issues you flagged should all be fixed - let me know if my fixes match what you had in mind

jonathanpallant:
I apologise for the delay in getting this PR approved. This was one of the first major LLM-assisted PR's we have received in this repository and it triggered a lot of internal discussion.

Unfortunately we have concluded that it is the policy of the knurling project from now on to not accept AI generated contributions. Our policy can be found at https://github.com/knurling-rs#contribution-policy.

I am therefore closing this PR.

PR #1031, opened 2026-01-27
Pin `time` version to fix backwards compatibility

aimir:
Looks like the current version of `main` is compatible except for the `time` version issue, so once this is merged you can change `REVISION_UNDER_TEST` to `origin/main` or to the next release tag.

### DIFF
--- a/xtask/src/backcompat.rs
+++ b/xtask/src/backcompat.rs
@@ -44,7 +44,7 @@ const DISABLED: bool = false;
 
 // use this format: PR <number> - <what feature / change broke compatibility>
 // PR #747 - Bump wire format
-const REVISION_UNDER_TEST: &str = "0e92d3a88aa472377b964979f522829d961d8986";
+const REVISION_UNDER_TEST: &str = "306d45db6ea94972e4acb3e4dec8387f0fc439ff";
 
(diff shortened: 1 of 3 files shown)


xtask/src/backcompat.rs
```diff
@@ -44,7 +44,7 @@ const DISABLED: bool = false;
 
 // use this format: PR <number> - <what feature / change broke compatibility>
 // PR #747 - Bump wire format
-const REVISION_UNDER_TEST: &str = "0e92d3a88aa472377b964979f522829d961d8986";
+const REVISION_UNDER_TEST: &str = "306d45db6ea94972e4acb3e4dec8387f0fc439ff";
```
jonathanpallant, line 47: We need this revision to point to a specific old version of `defmt` to check that we can still work with that version. https://github.com/knurling-rs/defmt/pull/1032 updates that old version to include a lock file, so please rebase after that PR goes through the merge queue.

aimir: I see, thanks

Looks like #1032 then renders this PR unneeded, I'll close it.

aimir:
The issue was fixed in #1032, closing this

PR #1032, opened 2026-01-28
add Cargo.lock file to revision used for backcompat test

japaric:
the revision was missing a lockfile and the workspace resolver defaulted to version 1 which is the non-MSRV (non-`rust-version`) aware resolver

said resolver was choosing package versions incompatible with the MSRV toolchain which caused the build of `qemu-run` to fail after a new package version of `time-core` was published

a commit adding a Rust 1.81 lockfile has been added to the previous under test revision; the resulting revision is tagged
'defmt-v0.3.4-plus-1.81-compatible-lockfile'

### DIFF
--- a/xtask/src/backcompat.rs
+++ b/xtask/src/backcompat.rs
@@ -44,7 +44,7 @@ const DISABLED: bool = false;
 
 // use this format: PR <number> - <what feature / change broke compatibility>
 // PR #747 - Bump wire format
-const REVISION_UNDER_TEST: &str = "0e92d3a88aa472377b964979f522829d961d8986";
+const REVISION_UNDER_TEST: &str = "3d66b3160bdba5ec0f9fa38fcfe80c3495b2da32";
 
 // the target name is in `firmware/qemu/.cargo/config.toml` but it'd be hard to extract it from that file
 const RUNNER_ENV_VAR: &str = "CARGO_TARGET_THUMBV7M_NONE_EABI_RUNNER";


xtask/src/backcompat.rs
```diff
@@ -44,7 +44,7 @@ const DISABLED: bool = false;
 
 // use this format: PR <number> - <what feature / change broke compatibility>
 // PR #747 - Bump wire format
```
Hoverbear, line 46: Does this line also need an update?

japaric: the root cause of the compatibility breakage has not changed so that line should stay the same.

xtask/src/backcompat.rs
```diff
@@ -44,7 +44,7 @@ const DISABLED: bool = false;
...
-const REVISION_UNDER_TEST: &str = "0e92d3a88aa472377b964979f522829d961d8986";
+const REVISION_UNDER_TEST: &str = "3d66b3160bdba5ec0f9fa38fcfe80c3495b2da32";
```
jonathanpallant, line 47: I'm being slow...

```console
$ git checkout 3d66b3160
error: pathspec '3d66b3160' did not match any file(s) known to git
```

Where is this commit?

jonathanpallant:
```console
defmt git:(ja-fix-backcompat) $ cargo xtask test-backcompat
    Finished `dev` profile [unoptimized + debuginfo] target(s) in 0.03s
     Running `target/debug/xtask test-backcompat`
🧪 backcompat
building old qemu-run.. (git revision: 3d66b3160bdba5ec0f9fa38fcfe80c3495b2da32)
stderr:
fatal: Could not parse object '3d66b3160bdba5ec0f9fa38fcfe80c3495b2da32'.

exit-code: 128
error building old qemu-run: `git reset` failed

Error: 😔 some tests failed: [
    "backcompat (building qemu-run)",
]
```

jonathanpallant: Ah, it's a commit that's only available through a tag. `git fetch --tags` makes it appear.

jonathanpallant (approved):

PR #1033, opened 2026-02-02
Add serde Serialize/Deserialize derives to Table and related types

aimir:
This enables caching the parsed defmt Table to disk, avoiding the need to parse the full ELF file for each test process. The cached table can be loaded in milliseconds instead of the ~6 seconds needed to parse a large ELF.

Types modified:
- Table, TableEntry, StringEntry, Tag, Encoding, BitflagsKey (lib.rs)
- Location (elf2table/mod.rs)

The Locations type is already a BTreeMap alias which supports serde.


jonathanpallant:
See #1036 where I proposed a change to the CHANGELOG.

aimir:
made redundant by #1036

PR #1035, opened 2026-02-02
qemu-run: print QEMU command before execution

robamu:
this prints the full QEMU command before executing it


jonathanpallant:
`qemu-run` is used for snapshot tests (checking stdout against reference files) for defmt, so changing the output is difficult....

jonathanpallant:
You could put it behind an optional `-v` (for *verbose*) flag?

robamu:
oh, is it specifying verbose for the snapshot tests as well?

robamu:
I am confused. why does it still fail? The printout is behind the verbose option now.

qemu-run/src/main.rs
```diff
@@ -189,15 +189,22 @@ fn notmain() -> Result<Option<i32>, anyhow::Error> {
     // grab stdout
     command.stdout(Stdio::piped());
 
+    if opts.verbose {
+        println!("Running QEMU command: {}", pretty(&command));
```
Urhengulas, line 193: Printing it to stderr instead of stdout is one option to fix the snapshot test failure. This way you could also always print it and not just in verbose mode.

```suggestion
        eprintln!("Running QEMU command: {}", pretty(&command));
```

Urhengulas: The snapshot tests only care about stdout, not stderr

jonathanpallant:
The tests are run with `--verbose` enabled...

https://github.com/knurling-rs/defmt/blob/5722c22819c6e06561abad90b9fe5c90d5546107/firmware/qemu/.cargo/config.toml#L8

Urhengulas:
> The tests are run with `--verbose` enabled...
> 
> https://github.com/knurling-rs/defmt/blob/5722c22819c6e06561abad90b9fe5c90d5546107/firmware/qemu/.cargo/config.toml#L8

But that does not matter since the tests only care about stdout and not stderr

robamu:
ah okay. just did not find it. maybe removing `--verbose` is the cleaner solution here for the runner?

Urhengulas (dismissed):
The changes look good to me, but someone still needs to fix the carg-deny failure and then rebase this PR on top of it.

CHANGELOG.md
```diff
@@ -968,7 +968,7 @@ Initial release
 
 ### [qemu-run-next]
 
-* No changes
+* Specifying the `--verbose` / `-v` option now also prints the actual QEMU command being run.
```
knoellle, line 971: Please link the PR number here like in the section below.
Don't forget to add the actual link to the list at the bottom too 😉

knoellle:
> The changes look good to me, but someone still needs to fix the carg-deny failure and then rebase this PR on top of it.

This should be done since #1041 was merged. Just need to rebase.

jonathanpallant:
@robamu do you still want to push this forward, or shall we leave it?

robamu:
ah, that went under. i can update this tomorrow

PR #1038, opened 2026-02-26
Update time version to address RUSTSEC-2026-0009

aimir:


aimir:
Looks like we can only get one of {RUSTSEC-2026-0009 mitigation, rust < 1.88 compatibility} 🙁

PR #1039, opened 2026-02-27
improve panic printout

robamu:
Adds a call to the [`PanicInfo::message`](https://doc.rust-lang.org/stable/core/panic/struct.PanicInfo.html#method.message) function, which was added in Rust 1.81. This means doing a defmt write of a `core::panic::PanicInfo` now includes the panic message, if and only if the panic message was a single string literal.

### DIFF
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -43,7 +43,7 @@ jobs:
       matrix:
         toolchain:
           - "stable"
-          - "1.76" # firmware MSRV
+          - "1.81" # firmware MSRV
     runs-on: ubuntu-latest
     steps:
       - uses: actions/checkout@v4
@@ -59,7 +59,7 @@ jobs:
       matrix:
         toolchain:
           - "1.87" # we pin clippy because it keeps adding new lints
-          - "1.76" # firmware MSRV
+          - "1.81" # firmware MSRV
     runs-on: ubuntu-latest
     steps:
       - uses: actions/checkout@v4
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -43,7 +43,7 @@ jobs:
       matrix:
         toolchain:
           - "stable"
-          - "1.76" # firmware MSRV
+          - "1.81" # firmware MSRV
     runs-on: ubuntu-latest
     steps:
       - uses: actions/checkout@v6
@@ -59,7 +59,7 @@ jobs:
       matrix:
         toolchain:
           - "1.87" # we pin clippy because it keeps adding new lints
-          - "1.76" # firmware MSRV
+          - "1.81" # firmware MSRV
     runs-on: ubuntu-latest
     steps:
       - uses: actions/checkout@v6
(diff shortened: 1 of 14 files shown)


jonathanpallant:
This includes an MSRV bump to 1.81, which we need to think carefully about.

jonathanpallant:
We agreed to bump the MSRV to 1.83 to match host-side, so we can accept this.

.github/workflows/ci.yml
```diff
@@ -43,7 +43,7 @@ jobs:
       matrix:
         toolchain:
           - "stable"
-          - "1.76" # firmware MSRV
+          - "1.81" # firmware MSRV
```
jonathanpallant, line 46: You'll need to search the repo for all other instances of `"1.76"` because the MSRV turns up in a bunch of places.

The equivalent Ferrocene release for 1.81 is "Ferrocene 24.11"

CHANGELOG.md
```diff
@@ -61,6 +61,8 @@ We have several packages which live in this repository. Changes are tracked sepa
 ### [defmt-next]
 
 * [#1091](https://github.com/knurling-rs/defmt/pull/1091) Allow inner modules to decrease env filter verbosity
+* [#1039](https://github.com/knurling-rs/defmt/pull/1039) Add support for `PanicInfo::message` if the message is a static string.
+* [#1039](https://github.com/knurling-rs/defmt/pull/1039) Introduced MSRV 1.81 for `defmt`.
```
jonathanpallant: defmt had an MSRV (noted in the README) - this PR just bumped it to 1.81

jonathanpallant (commented):
I am OK with bumping the MSRV to get this API, but it is a little more involved.

robamu:
I added this for defmt-03 as well because it pulls defmt 1 in anyway. I also updated other places where the old MSRV was used.

jonathanpallant:
Should we gate this behind a flag? I worry that someone might be annoyed that a panic log that was only a few bytes might now be a hundred bytes long (if the message was long).

robamu:
should we make the flag opt-in or opt-out?

jonathanpallant:
I think it should be opt-in, so the default behaviour is unchanged when you `defmt::println!("{}", panic_info);`.

CHANGELOG.md
```diff
@@ -61,6 +61,8 @@ We have several packages which live in this repository. Changes are tracked sepa
 ### [defmt-next]
 
 * [#1091](https://github.com/knurling-rs/defmt/pull/1091) Allow inner modules to decrease env filter verbosity
+* [#1039](https://github.com/knurling-rs/defmt/pull/1039) Add support for `PanicInfo::message` if the message is a static string.
+* [#1039](https://github.com/knurling-rs/defmt/pull/1039) Add rust-version 1.81 for `defmt` package configuration.
```
knoellle, line 65: Since we're bumping the MSRV, I think this should be more clearly phrased as a potentially breaking change.

robamu: Like this: "Possibly breaking: ..."?

PR #1044, opened 2026-03-27
feat: add bound attribute

vic1707:
fixes #1043 

Tired my best to create meaningful tests, if you have better ideas please tell me!

PR #1046, opened 2026-04-02
Use general semihosting crate instead of cortex-m specific crate

### LINKED ISSUE #829: defmt-test build fails for non-CortexM thumb targets
Our team is using defmt-test as a foundation for several unit and integration tests. Our target processor is a Cortex R core, and we build for the following architecture targets:

`armv8r-none-eabihf`
`thumbv8r-none-eabihf`

Most of our development occurred while building and testing with `armv8r-...`; we have only recently started using `thumbv8r-...`. When we made the switch to thumb, our test builds started failing with the following error:

```
   Compiling cortex-m v0.7.7
error: failed to run custom build command for `cortex-m v0.7.7`

Caused by:
  process didn't exit successfully: `/workdir/build/debug/build/cortex-m-73e6882762cfa3ea/build-script-build` (exit status: 101)
  --- stderr
  thread 'main' panicked at /opt/cargo/registry/src/index.crates.io-6f17d22bba15001f/cortex-m-0.7.7/build.rs:25:10:
  called `Result::unwrap()` on an `Err` value: Os { code: 2, kind: NotFound, message: "No such file or directory" }
  note: run with `RUST_BACKTRACE=1` environment variable to display a backtrace
```

At first we wondered why the cortex-m crate was even being built for a non-CortexM target architecture (armv8r), but then we looked at the defmt-test dependencies in Cargo.toml:

```
[dependencies]
cortex-m-rt = "0.7"
cortex-m-semihosting = "0.5"
...
```

Looking further into the issue, we found that switching from arm to thumb target invokes [this code](https://github.com/rust-embedded/cortex-m/blob/bb4a78208323260a161e68b2498438867f971bc5/build.rs#L14) in the build of the cortex-m crate:

```
    if target.starts_with("thumb") {
        ...
        fs::copy(
            format!("bin/{}{}.a", target, suffix),
            out_dir.join(format!("lib{}.a", name)),
        )
        .unwrap();
        ...
    }
```

This `unwrap()` is panicking because it is looking for a `.a` file that does not exist, because the crate is being built for an incompatible target architecture. With armv8r (i.e. `target.starts_with("thumb")` is false), this part of the build script is never invoked and the build succeeds.

Overall, we are wondering if you would consider removing this hard dependency on the cortex M crates from defmt-test. Perhaps you could put this behind a default feature flag that could be turned off when using armv8r or similar target architectures. We have had great success using defmt-test so far. That is, our test binaries build and execute successfully with the `main()` provided by defmt-test, when built for armv8r. We would like to continue using and building more tests upon the defmt-test foundation. However, we also would like to build for thumbv8r, and without forking the defmt repo there is currently no way we can think of to do this. In our minds there should be a way for us to "opt out" of the Cortex M requirement here.

Thank you for your consideration.

Siddhant779:
Defmt-test build fails for non-CortexM thumb targets. Instead of using a semihosting cortex-m specific crate, use a general semihosting crate.

Fixes #829

### DIFF
--- a/firmware/qemu/src/bin/panic.rs
+++ b/firmware/qemu/src/bin/panic.rs
@@ -1,7 +1,9 @@
 #![no_std]
 #![no_main]
 
+use cortex_m as _;
 use cortex_m_rt::entry;
+use semihosting::process::ExitCode;
 
 use defmt_semihosting as _; // global logger
 
@@ -20,9 +22,5 @@ fn main() -> ! {
 #[cfg(target_os = "none")]
 #[panic_handler]
 fn panic(_: &core::panic::PanicInfo) -> ! {
-    use cortex_m_semihosting::debug;
-
-    loop {
-        debug::exit(debug::EXIT_SUCCESS)
-    }
+    ExitCode::SUCCESS.exit_process()
 }
--- a/.github/workflows/cargo-semver-check.yml
+++ b/.github/workflows/cargo-semver-check.yml
@@ -39,5 +39,5 @@ jobs:
         uses: obi1kenobi/cargo-semver-checks-action@v2
         with:
           manifest-path: firmware/
-          exclude: defmt-semihosting
+          exclude: defmt-semihosting,defmt-test
           # rust-target: thumbv7em-none-eabihf
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -891,6 +891,7 @@ Initial release
 ### [defmt-test-next]
 
 * [#1049] Pin trybuild to 1.0.89
+* [#1046] Replace `cortex-m-semihosting` with generic `semihosting` crate to support non-Cortex-M thumb targets
 
 ### [defmt-test-v0.4.0] (2025-04-01)
 
@@ -1010,6 +1011,7 @@ Initial release
 [#1049]: https://github.com/knurling-rs/defmt/pull/1049
 [#1048]: https://github.com/knurling-rs/defmt/pull/1048
 [#1047]: https://github.com/knurling-rs/defmt/pull/1047
+[#1046]: https://github.com/knurling-rs/defmt/pull/1046
 [#1044]: https://github.com/knurling-rs/defmt/pull/1044
 [#1041]: https://github.com/knurling-rs/defmt/pull/1041
 [#1036]: https://github.com/knurling-rs/defmt/pull/1036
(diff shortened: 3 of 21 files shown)


firmware/qemu/src/bin/panic.rs
```diff
@@ -1,6 +1,7 @@
 #![no_std]
 #![no_main]
 
+use cortex_m as _;
```
jonathanpallant, line 4: Perhaps `use semihosting::process::ExitCode;` for consistency with the other examples?

jonathanpallant (approved):
One nit that I spotted, but in general this seems fine to me. I use the semihosting crate extensively in the aarch32 repository and it works fine over there, and broadening support beyond cortex-m for this crate I think is useful.

firmware/qemu/src/bin/panic.rs
Siddhant779: Addressed

Siddhant779:
Hi! We're interested in using the changes from this PR in our project. Is there a timeline for when this might get merged and included in a new crates.io release of defmt-test? Currently our only option would be to pull from git, which isn't ideal for our workflow.

Happy to help with anything needed to move this forward. Thanks!

knoellle:
Hello @Siddhant779, thank you for your contribution.
Jonathan already gave his blessing and the changes look fine to me as well.
I'd be happy to push this forward with you, I think we can merge it once the CI is happy.
Checking internally regarding a release.

Siddhant779:
> Hello @Siddhant779, thank you for your contribution. Jonathan already gave his blessing and the changes look fine to me as well. I'd be happy to push this forward with you, I think we can merge it once the CI is happy. Checking internally regarding a release.

Sounds good thanks @knoellle! Looks like the only test that is failing is `host (ubuntu-latest, stable)` which seems to be failing on [main](https://github.com/knurling-rs/defmt/actions/runs/24810451932), so not sure if it's something my PR is introducing. Please keep me updated regarding a release 

Thanks

knoellle:
The main CI breakage should be fixed with
- #1052 

Please rebase and try again. Enable merge when ready when you're happy with it.
We're also going to be making a release in the coming weeks. :)

knoellle (dismissed):
Thank you for your contribution!
Code LGTM but there is a conflict with #1044 in the changelog entries

Siddhant779:
> Thank you for your contribution! Code LGTM but there is a conflict with #1044 in the changelog entries

Could i get re-approval; Just rebased the PR to deal with the conflict

knoellle (approved):
LGTM

knoellle:
@Siddhant779 we just released a new version of the relevant crates.
Feel free to open an issue if you have any problems :)

PR #1048, opened 2026-04-08
Fix qemu-run uart printing

### LINKED ISSUE #1047: qemu-run won't bring UART data after a gap
If your program is run inside `qemu-run` and puts data out on the UART, if there's a gap in the UART data of about 100ms, `qemu-run` stops printing anything to the UART.

I think this is because the `print_loop` function cannot tell the difference between a timeout error and a fatal error, and disconnects regardless.

jonathanpallant:
Fix for #1047 - now we carry on in the event of a UART-over-TCP read timeout.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -969,7 +969,7 @@ Initial release
 
 ### [qemu-run-next]
 
-* No changes
+* [#1048] Fixed UART read timeout issue ([#1047])
 
 ### [qemu-run-v0.1.1]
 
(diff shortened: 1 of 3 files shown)


jonathanpallant:
Tested with https://github.com/ferrous-systems/rust-training/pull/334 - I now see all the output, even with a 500ms gap between strings.

knoellle (approved):
Was able to reproduce the issue and verify the fix.
LGTM

PR #1050, opened 2026-04-20
Refactor avoid-default-panic to not rely on the linker script

knoellle:
#812 introduced a feature to remove the default panicking behavior from for `defmt::panic` in case the user provides [their own panic handler](https://docs.rs/defmt/latest/defmt/attr.panic_handler.html).
I don't know why this was implemented by modifying the linker script instead of just `cfg`ing out the whole function.

Not using the linker script for this also improves compatibility with macOS, see https://github.com/knurling-rs/defmt/pull/1029#issuecomment-4188847484

# Details

- `defmt::panic` logs its error message, then calls `_defmt_panic` to do the actual panicking.
- Previously, `_defmt_panic` wasn't defined anywhere but instead there was `__defmt_default_panic`.
- These two symbols were then combined in the linker script unless the `avoid-default-panic` feature was enabled.
- The new behavior just defines `_defmt_panic` IFF the feature is not enabled

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -69,6 +69,7 @@ We have several packages which live in this repository. Changes are tracked sepa
 * [#983] Add `Format` implementation for `core::num::Wrapping<T>`
 * [#1022] Re-fix accidental breaking change in `Format` macro
 * [#1041] Ignore RUSTSEC-2026-0009
+* [#1050] Refactor avoid-default-panic to not rely on the linker script
 
 ### [defmt-v1.0.1] (2025-04-01)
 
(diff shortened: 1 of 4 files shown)


knoellle:
Damn, didn't think about the duplicate definitions.
That makes it a breaking change because users who were previously overriding the defmt panic handler are now required to use the `avoid-default-panic` feature to even compile.

The `PROVIDES` line in the linker script was effectively doing weak linking, which is supported by rust but hasn't been stabilized for over a decade now. https://github.com/rust-lang/rust/issues/29603

Not sure if there is another way to achieve the same result without using either a linker script or an unstable feature.

jonathanpallant:
@knoellle I'm marking this as draft as it sounds like additional work is required here?

PR #948, opened 2025-03-18
feat: Implement `crate` overwrite for `Format` derive

Veykril:
Closes https://github.com/knurling-rs/defmt/issues/928

PR #1052, opened 2026-04-24
Pin UI tests to host MSRV

knoellle:
### What?

Change the Rust version UI tests are executed on from stable to the project's host MSRV.

### Why?

The UI tests ensure that the error messages generated by the compiler are reasonable for each test case.
But since these error messages also depend on the compiler version, it doesn't make sense to check against multiple compiler version (or we'd have to duplicate the expected output).
#948 limited the UI tests to only run with the stable compiler version.
However, as "stable" as that sounds, it is still a moving target, leading to CI breakage on main and causing unnecessary friction for PR authors.


Hoverbear (approved):

Hoverbear:
Seems like a fair and practical approach. It may inconvenience some developers but it saves us from tests failing between versions.

PR #1053, opened 2026-05-05
defmt-rtt: drop-on-contention

jordens:
This adds the `drop-on-contention` feature to `defmt-rtt`.

Instead of using a critical section to ensure sequential access to the RTT buffer from different execution contexts, this uses an atomic execution context "owner" set from Cortex-M IPSR. It removes the need to disable interrupts during logging at the cost of discarding defmt frames from high priority contexts when a lower priority context is already writing a frame.

It also adds testing (check cross and qemu snapshot) for that feature.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -755,7 +755,7 @@ Initial release
 
 * [#1006] Fix `available_buffer_size` ignoring available buffer space when `read < write`
 * [#1013] Switch to fixed 32-bit types, to support 64-bit targets
-* [#1053] Add the `disable-irq-masking` feature to trade low interrupt latency for reliable RTT delivery
+* [#1053] Add the `drop-on-contention` feature to trade low interrupt latency for reliable RTT delivery
 
 ### [defmt-rtt-v1.1.0] (2025-10-09)
 
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -755,6 +755,7 @@ Initial release
 
 * [#1006] Fix `available_buffer_size` ignoring available buffer space when `read < write`
 * [#1013] Switch to fixed 32-bit types, to support 64-bit targets
+* [#1053] Add the `disable-irq-masking` feature to trade low interrupt latency for reliable RTT delivery
 
 ### [defmt-rtt-v1.1.0] (2025-10-09)
 
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -1019,6 +1019,7 @@ Initial release
 
 ---
 
+[#1053]: https://github.com/knurling-rs/defmt/pull/1053
 [#1055]: https://github.com/knurling-rs/defmt/pull/1055
 [#1052]: https://github.com/knurling-rs/defmt/pull/1052
 [#1049]: https://github.com/knurling-rs/defmt/pull/1049
--- a/firmware/defmt-rtt/Cargo.toml
+++ b/firmware/defmt-rtt/Cargo.toml
@@ -12,7 +12,7 @@ version = "1.2.0"
 
 [features]
 disable-blocking-mode = []
-disable-irq-masking = []
+drop-on-contention = []
 
 [dependencies]
 defmt = { version = "1", path = "../../defmt" }
--- a/firmware/defmt-rtt/Cargo.toml
+++ b/firmware/defmt-rtt/Cargo.toml
@@ -12,6 +12,7 @@ version = "1.2.0"
 
 [features]
 disable-blocking-mode = []
+disable-irq-masking = []
 
 [dependencies]
 defmt = { version = "1", path = "../../defmt" }
--- a/firmware/defmt-rtt/src/channel.rs
+++ b/firmware/defmt-rtt/src/channel.rs
@@ -24,6 +24,7 @@ pub(crate) struct Channel {
     pub flags: AtomicU32,
 }
 
+#[cfg(not(feature = "disable-irq-masking"))]
 impl Channel {
     pub fn write_all(&self, mut bytes: &[u8]) {
         // the host-connection-status is only modified after RAM initialization while the device is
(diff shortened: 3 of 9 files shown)


firmware/defmt-rtt/src/lib.rs
```diff
@@ -218,7 +238,158 @@ impl RttEncoder {
...
+        }
+    }
+
+    /// Flush the encoder.
+    ///
+    /// # Safety
+    ///
+    /// Do not call unless you have called `acquire`.
+    unsafe fn flush(&self) {
+        _SEGGER_RTT.up_channel.flush();
```
knoellle, line 349: Like the other functions here, the docstring says "Do not call unless you have called `acquire`." But is just calling it sufficient? In this implementation the `acquire` may silently fail.
Could this cause issues if a higher priority task calls `flush` while a lower priority task was interrupted during a logging operation?

firmware/defmt-rtt/src/channel.rs
```diff
@@ -178,3 +221,98 @@ mod tests {
         }
     }
 }
+
+#[cfg(all(test, feature = "disable-irq-masking"))]
+mod test_disable_irq_masking {
```
knoellle: These tests are currently not executed in the CI because it never enables this feature. Please make sure they are.

firmware/defmt-rtt/Cargo.toml
```diff
@@ -12,6 +12,7 @@ version = "1.1.0"
 
 [features]
 disable-blocking-mode = []
+disable-irq-masking = []
```
knoellle: I think this feature needs another name. Currently it describes the symptom you're trying to avoid but says nothing about the consequences (dropped logs). Maybe something like `discard/drop-on-contention` but I'm open to suggestions.

knoellle (changes requested):
Thank you for your contribution

firmware/defmt-rtt/Cargo.toml
jordens: I followed the existing "disable-blocking-mode" style which has the exact same problem.

`best-effort` was also on the menu but that is vague. I'm fine with `drop-on-contention`.

firmware/defmt-rtt/src/lib.rs
jordens: Right. `flush()` was the straggler.

firmware/defmt-rtt/src/lib.rs
```diff
@@ -38,6 +38,18 @@
 //! [dependencies]
 //! cortex-m = { version = "0.7.6", features = ["critical-section-single-core"]}
 //! ```
+//!
+//! With feature `drop-on-contention` you do not need a critical section
+//! implementation and interrupts are not disabled. Instead, when execution
+//! contexts collide, frames are dropped. This mode is for bare-metal
+//! Cortex-M use where thread mode is a single execution context. It is not
+//! correct on RTOS or other multi-thread-mode systems because all thread-mode
+//! tasks share `IPSR == 0`, which can misidentify ownership and panic. It can
```
knoellle, line 47: I don't suppose there is a way to detect this and throw a compilation error when building for these platforms with this feature enabled?
If not, I would at least like to adjust the panic message to hint at the fact that the detected reentrancy might stem from using a multi-threading model that is not supported with this feature enabled.

firmware/defmt-rtt/src/lib.rs
knoellle, line 1: Just curious, why did you revert the unsafe removal commit? The CI seemed to be happy with the previous commit.
Where would you have needed to "feature-mux" as you put it?

knoellle (changes requested):
Some minor comments, I think this is in a mergable state otherwise.

CHANGELOG.md
```diff
@@ -747,6 +747,7 @@ Initial release
 
 * [#1006] Fix `available_buffer_size` ignoring available buffer space when `read < write`
 * [#1013] Switch to fixed 32-bit types, to support 64-bit targets
+* [#1053] Add the `drop-on-contention` feature to trade low interrupt latency for reliable RTT delivery
```
knoellle, line 758: The link here will only work once you add a [link reference](https://github.github.com/gfm/#shortcut-reference-link) to the sorted list at the bottom of the file.

firmware/defmt-rtt/src/lib.rs
```diff
@@ -218,7 +238,167 @@ impl RttEncoder {
...
+            core::arch::asm!(
+                "mrs {}, ipsr",
+                out(reg) ipsr,
+                options(nomem, nostack, preserves_flags)
+            );
+            ipsr
+        }
+
+        #[cfg(not(target_arch = "arm"))]
+        0
```
knoellle: This defaulting behavior has the potential for significant breakage, nor is it documented anywhere.
How important is it to you that this feature builds on non-arm platforms?
I would rather emit a `compiler_error` here. If there are other platforms that people want supported, we can add implementations for those as well.
Since we're just introducing this feature, we can always broaden the supported platforms later, but doing the reverse can break people's builds. So I would rather start with a more conservative set of platforms.

jordens: It enables host tests of that feature. I'll nix that and thread it through qemu.

firmware/defmt-rtt/src/lib.rs
jordens: I only added tests for `drop-on-contention` where this did work. There are no tests for the pre-existing code where this doesn't. I'll keep the AtomicRTTEncoder fns safe but their calling context needs to stay unsafe.

firmware/defmt-rtt/src/lib.rs
jordens: That kills the host testing. I'll put in some qemu snapshot and cargo check cross compile tests instead. It didn't feel smooth because there is currently not much defmt-rtt testing at all.

knoellle (dismissed):

Btw, in the future please don't resolve conversations as a PR author; Leave that to the reviewer.
As a reviewer, I want to check off my change requests when reviewing new commits to see if they match what I had intended.

Sadly github offers no control over who can resolve conversations.

knoellle:
No sure why this isn't merging.
My best guess is that your push happened to be right in the affected time span of this incident https://www.githubstatus.com/incidents/gnftqj9htp0g preventing the CI run from being scheduled at all.
I don't see a button anywhere to manually trigger it, can you try poking it again?

jordens:
> Btw, in the future please don't resolve conversations as a PR author; Leave that to the reviewer. As a reviewer, I want to check off my change requests when reviewing new commits to see if they match what I had intended.
> Sadly github offers no control over who can resolve conversations.

If you prefer. The simple practical alternative that I've seen is to use `resolution` as a two-way token. First the author marks it resolved when done and then the reviewer marks as unresolved when unhappy. But yes, the UX is crap in either case.

knoellle (approved):
LGTM

knoellle:
Hmmm, a conflict...
could you rebase and poke it again? :D

PR #1055, opened 2026-05-11
qemu-run supports `--arg foo=bar` to pass `-foo bar` to qemu.

jonathanpallant:
Allows you to start QEMU in SMP mode, or set other useful properties on the virtual machine.

### DIFF
--- a/qemu-run/src/main.rs
+++ b/qemu-run/src/main.rs
@@ -38,6 +38,10 @@ struct Opts {
     #[arg(long, required = false)]
     cpu: Option<String>,
 
+    /// Optionally additional arguments for QEMU
+    #[arg(long, required = false)]
+    arg: Vec<String>,
+
     /// Optionally specify a custom defmt log format
     #[arg(short = 'l', required = false, alias = "log-format")]
     log_format: Option<String>,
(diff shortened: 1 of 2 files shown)


qemu-run/src/main.rs
```diff
@@ -38,6 +38,10 @@ struct Opts {
     #[arg(long, required = false)]
     cpu: Option<String>,
 
+    /// Optionally additional arguments for QEMU
+    #[arg(long, required = false)]
+    arg: Vec<String>,
```
Hoverbear, line 43: Are there ever situations where a user might want to pass `--arg foo` (no `=bar`?)? I see `qemu-system-arm`  has things like `-nographic` and `-usb`.

Maybe taking `--extra-qemu-args "-foo bar"` would be more compatible?

jonathanpallant: yeah, that's true. I don't love the idea of having to escape the space with a double quote, but I should handle the missing `=` case.

jonathanpallant: Right now you'd have to write `--arg nographic=`

jonathanpallant: Fixed

Hoverbear (dismissed):

jonathanpallant:
Fixed the formatting

Hoverbear (approved):

PR #1056, opened 2026-05-12
Bump qemu-run to 0.3.0

jonathanpallant:
We see tags for 0.2.0 but no release, and we see CHANGELOG entries for 0.1.1 but no release, so we're going straight to 0.3.0 to avoid confusion.


knoellle (approved):
LGTM

PR #1059, opened 2026-05-21
Check changelog in CI

knoellle:
We already had a script to manually check the changelog for broken reference links.
This PR improves the output and executes it in CI.

See [here](https://github.com/knurling-rs/defmt/actions/runs/26237813408/job/77215762048?pr=1059) for a failing example.

### DIFF
--- a/check_changelog.sh
+++ b/check_changelog.sh
@@ -4,5 +4,5 @@
 
 for link in `rg '\[(#\d+)\] ' CHANGELOG.md -o | sort | uniq | sed 's/\]/\\\\]/' | sed 's/\[/\\\\[/'`
 do
-    rg "${link}:" CHANGELOG.md -q || echo $link is MISSING
+    grep "${link}:" CHANGELOG.md -q || echo ${link//\\/} is MISSING a link-reference, see https://github.github.com/gfm/#shortcut-reference-link
 done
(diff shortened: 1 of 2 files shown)


check_changelog.sh
```diff
@@ -0,0 +1,15 @@
+#!/bin/bash
+
+# Checks that all the github links in CHANGELOG.md are correctly defined
+
+status=0
+
+for link in `rg '\[(#\d+)\] ' CHANGELOG.md -o | sort | uniq | sed 's/\]/\\\\]/' | sed 's/\[/\\\\[/'`
+do
+    if ! rg "${link}:" CHANGELOG.md -q; then
+        echo ${link//\\/} is MISSING a link-reference, see https://github.github.com/gfm/#shortcut-reference-link
```
jonathanpallant, line 10: What does `${link//\\/}` do? I couldn't parse the slashes.

knoellle: On main the script emits lines like `\[#99999\] is MISSING`.
This syntax is a search and replace of all backslashes to an empty string.
[Reference](https://www.delftstack.com/howto/linux/bash-replace-character-in-string/#method-2---use-bash-parameter-expansion)
`/x/y` replaces the first `x` with a `y`, `//x/y` replaces all `x`s with `y`s.

jonathanpallant: TIL

jonathanpallant (approved):

PR #1040, opened 2026-03-07
fix: use section-relative offsets for .defmt symbol lookup

LoganDark:
**Summary**

- Fix tag lookup failure when the `.defmt` section has a non-zero base address
- Use section-relative offsets (matching what the firmware encodes) instead of raw VMAs as map keys

**Problem**

The firmware side encodes defmt tags by casting the symbol address to `u16`:

```rust
&#var_name as *const u8 as u16
```

This naturally truncates the address to a section-relative offset when the `.defmt` section is aligned (e.g. at `0x20000000` → tag becomes `0x0006`).

However, `elf2table` builds its lookup map using the full VMA (`entry.address()`) as the key. When the decoder extracts tag `6` from the wire format and looks up `entries[6]`, it finds nothing because the entry is actually stored under `entries[0x20000006]`, and the frame is reported as malformed.

This potentially affects any linker configuration where `.defmt` has a non-zero section address, which is sometimes needed for RISC-V (when `.defmt` must be within ±2 GiB of `.text` for `auipc` relocations).

**Fix**

Subtract `defmt_section.address()` from `entry.address()` when building the map keys and when indexing into section data for bitflags values, so the keys match the truncated `u16` tags the firmware sends.

**Testing**

- Verified with a RISC-V bare-metal kernel where `.defmt` is at VMA `0x20000000`
- Before: all frames reported as "malformed frame skipped"
- After: frames decode correctly

---

This patch and commit massage were written with the help of Claude Code, but I have personally (and manually) refined and verified its outputs.

Issue #463, opened 2021-05-02
Support non-embedded targets


Dirbaio:
Currently defmt only works in ELF targets, and requires a custom linker script. Is supporting non-embedded targets in the radar? 

- Targets that don't use ELF: **Windows, iOS, MacOS**
- Targets that do use ELF, but modifying the linker script is not as practical: **Linux, Android**

I think the most portable way would be to not rely en ELF/linker tricks and have the proc macros hit flat files (or a sqlite DB?) to assign indexes to format strings. 

This would be a big change, so if you don't want to do this because the only goal is supporting embedded devices, I understand :) I thought I'd open this issue to confirm.

Urhengulas:
This certainly sounds like a desirable feature and feasible if someone likes to contribute or fund it.

DBLouis:
Is it possible to make defmt works on standard x86 linux? It fails to link with the custom linker script. If I don't use the custom linker script `readelf` still shows a section that contains my log strings. However when running `defmt-print` on the ELF it fails with: `Error: .defmt data not found`.

japaric:
> Is it possible to make defmt works on standard x86 linux?

No OS is supported at the moment; that's what this issue is about. What you are describing are the symptoms of several unimplemented things.

> It fails to link with the custom linker script.

sounds about right. wouldn't know the exact reason off the top of my head but may be related to Linux binaries being relocatable. Firmware images use absolute addresses for everything (relocation=static in linker/compiler lingo). This ("format string 'interning'" is probably the hardest thing to fix / design / implement. More so on platforms that don't use the ELF format: macOS and Windows.

> If I don't use the custom linker script readelf still shows a section that contains my log strings.

(interesting that they "survive" (are not discarded by the linker) the linking process)

> However when running defmt-print on the ELF it fails with: Error: .defmt data not found.

without the custom linker script the log strings won't end in the right place so the .defmt linker section is not created hence the error message

---

The other thing that's missing is a transport crate for Linux and other OSes. That crate defines *how* the encoded/compressed defmt data is output (e.g. framing) and to where (e.g. stdout, a file, a named pipe, a socket, etc.). You would then pass this defmt data to `defmt-print`

DBLouis:
Is there any hope of making it work by setting the right compiler flags so that the log strings are "at the right place"? I don't know much about linking so maybe I am missing the point

japaric:
[...]

PR #1066, opened 2026-06-02
decoder: support load bias in defmt tables

jordens:
## Goal

This PR is the second part to enable support for defmt without any special linker scripts. #1073 was the first part. `defmt2log` exercises this use case by becoming a thin and robust `defmt` sink and `log` source. Having defmt work without linker scripts in turn lowers the adoption barrier of `defmt` logging in embedded libaries as it upholds the advantages of the defmt format hints while still supporting development and testing on host operating systems: no need for the usual [`fmt.rs`](https://github.com/probe-rs/embedded-test/blob/bb99a651e1ff88df1047802f26f595d3d0288e42/src/fmt.rs) feature-mux to switch between `defmt` and `log` macros and no need to restrict the the logging to the common subset (excluding defmt format hints).

The original implementation is from `defmt2log` which can now discard all its synthetic table generation tooling and duplicate elf parsing (https://github.com/quartiq/defmt2log/pull/2).

## Status quo

The decoder assumes that the `Table`/`Locations` addresses have no load bias/load slide and that they never collide.
The truncation from address to `u16` index and the lookup happen in various places as a plain cast and the upcast `u16 -> usize` is naive.

## Problem

With no linker script placing `.defmt*` symbols at zero offset, dynamic loader, address space layout randomization there is an offset (a.k.a. load bias, ELF slide) between the `u16` wire format index sent by the Encoder with that runtime bias and and runtime-independent `usize` `Table` key that the Decoder uses based on the ELF alone.

Also now with unmerged defmt sections the addresses could in theory span more than the `u16` range and could collide since no linker script collocates them or limits their address space.

The lookup of the `u16` in the `usize` Table must account for collisions, for the load bias and the widening.

## Approach

The idea is to separate the `Table` key from the wire format and not enshrine the mapping between them.

The load bias is the difference between `defmt::runtime_anchor()` (known by the Encoder side) and the ELF address of that same anchor symbol (known by the Table/Decoder).
The (*new*) `DecodeContext` holds that load bias.

A map from the (unbiased) wire u16 to the wide usize addresses is added to the Table.
That map is necessary since (a) we don't want to change the `entries` key format to `u16` and (b) due to the absence of the linker script there is no closed form `u16 -> usize` conversion.

Per-frame (*new*) `Table::decode_with_context()` uses that prebuilt `DecodeContext`.
The existing `Table::decode()` uses that with the zero `DecodeContext`.

Since with split defmt sections index collisions are not detected at compile time anymore (through the linker script), the code detects them at runtime. This is not a regression and not a problem in practice: since the linker still places the defmt sections close together the practical collision scenario is the same as before (more than one u16 worth of addresses).

## Testing

* Added unittests (most of this PR is unittests). Happy to cull the unittests tests if they are deemed noisy.
* Exercised and used for a few months now during development of several crates (`minimq`, `miniconf`, `stabilizer`, and others). Those are mid- to high-complexity defmt users. Both on embedded (defmt-rtt) and on host/CI (defmt2log).

---

## CC
* #1029 with linker scripts
* #1040 also addressing a difference between runtime address and ELF address of a defmt symbol in a very specific case
* #463 which was looking at alternative/external defmt table mgmt without linker scripts
* #149 which covers the general defmt/log dichotomy. The usual `fmt.rs` feature mux or `defmt2log` are viable options.


jonathanpallant:
Apologies for the delay in getting to this - on my list for next week

jonathanpallant:
note to self - check for conflicts with #1029

jordens:
I reworked the PR into two commits to help with reviewability and also improve on the modulo/wire format vs table key issue.

Commit 1: section gathering only. This can be reviewed in isolation. Also happy to split it out into its own PR that helps. It focuses on only enabling unmerged defmt sections. No load bias, no modulo/wrapping. It also removes an accidental dependency on `DW_AT_location address == TableEntry raw symbol address == Table.entries key == Frame::index()` which is both not needed and breaks in various modes which are encountered here and in #1029, #1040 etc.

Commit 2: biased/wrapped decoding. This is more interesting because supporting load bias while still detecting collisions previously ran against the existing truncation/modulo policy which was an implicit assumption in different places. The new idea is to separate the Table key from the wire format and not enshrine the u16 index size, bias and modulo into the Table.

- Adds a `DecodeIndex` built from `DecodeOptions`
- `DecodeOptions` separates:
  - `index_type`: how many bytes to read from the frame
  - `modulo_bits`: size of the index space
  - `address_bias`: bias/slide applied to table addresses before modulo

- Per-frame `decode_with_index()` uses the prebuilt map
- `Table::decode()` remains the default u16 direct path
- Colliding wrapped indices are rejected by making ambiguous map entries decode as malformed

There are also some ~200 lines of tests which push the diff size, but all tests are cases that were hit during development. Happy to cull that if deemed noisy.

IMO this is a base/alternative to #1029 (certainly for the non-mach-o/macos parts) and #1040 (most of it?).

jordens:
To make the diff smaller I dropped the eager impl of support for non-u16 indices but kept the DecodeOptions/DecodeIndex layer. The u32/u64 support can be re-added if there is a use case.

jordens:
Converted to draft since this depends on #1073 to be merged first. I split off that one to make it more digestible.

jordens:
@jonathanpallant is this PR (and #1073) just waiting for reviewer manpower or is it in limbo? Can I do something to help/improve/accelerate?

jonathanpallant:
I tried this out and I couldn't make it work ... what am I missing?

```rust
use defmt_stdout as _;

fn main() {
    defmt::println!("Hello, world!");
}

defmt::timestamp!(
    "{=u64:tus}",
    std::time::SystemTime::now()
        .duration_since(std::time::SystemTime::UNIX_EPOCH)
        .unwrap()
        .as_micros() as u64
);
```


(the defmt::timestamp seems mandatory without the default PROVIDE in the defmt.x file)

a basic encoder:

```rust
use std::{io::Write, ops::DerefMut, sync::{Condvar, Mutex}};

struct StdoutEncoder {
    condvar: Condvar,
    encoder: Mutex<StdoutEncoderInner>,
}

struct StdoutEncoderInner {
    encoder: defmt::Encoder,
    stdout: Option<std::io::Stdout>,
}

static ENCODER: StdoutEncoder = StdoutEncoder {
    condvar: Condvar::new(),
    encoder: Mutex::new(StdoutEncoderInner {
        encoder: defmt::Encoder::new(),
        stdout: None,
    })
};

impl StdoutEncoder {
    fn acquire(&self) {
        let mut guard = self.encoder.lock().unwrap();
        while guard.stdout.is_some() {
            guard = self.condvar.wait(guard).unwrap();
        }
        if guard.stdout.replace(std::io::stdout()).is_some() {
            panic!("Double locked output?!");
        }
    }

    fn release(&self) {
        let mut guard = self.encoder.lock().unwrap();
        guard.stdout = None;
        self.condvar.notify_one();
    }

    fn write(&self, bytes: &[u8]) {
        let mut guard = self.encoder.lock().unwrap();
        let guard = guard.deref_mut();
        guard.encoder.write(bytes, |buf| {
            guard.stdout.as_mut().unwrap().write_all(buf).expect("stdout write");
        });
    }

    fn flush(&self) {
        let mut guard = self.encoder.lock().unwrap();
        let guard = guard.deref_mut();
        guard.stdout.as_mut().unwrap().flush().unwrap();
    }
}

#[defmt::global_logger]
struct Logger;

unsafe impl defmt::Logger for Logger {
    fn acquire() {
        ENCODER.acquire();
    }

    unsafe fn flush() {
       ENCODER.flush();
    }

    unsafe fn release() {
        ENCODER.release();
    }

    unsafe fn write(bytes: &[u8]) {
        ENCODER.write(bytes);
    }
}
```

I get some output:

```console
$ cargo run | hexdump -C
00000000  30 a7 0f 39 63 01 49 57  06 82                    |0..9c.IW..|
0000000a
```

But I cannot decode it:

```console
$ cargo run | ../defmt/target/debug/defmt-print -v --show-skipped-frames -e ./target/debug/linux-defmt stdin
    Finished `dev` profile [unoptimized + debuginfo] target(s) in 0.01s
     Running `target/debug/linux-defmt`
$
```

I get no output from the decoder, and no clue as to what went wrong.

jordens:
I see why you want to try this with a minimal stdout encoder and defmt-print. There are two problems:

1. the minimal encoder has broken framing (at least for rzcobs): call `defmt::Encoder::{start,end}_frame()` in `{acquire,release}()`. If you do it, you'll get

```
(HOST) malformed frame skipped
└─ defmt-print @ print/src/main.rs:322
```

Should be the same for both current main and this branch, because:

2. More importantly the `defmt-print` decoder sink doesn't apply -- and can't easily determine -- the (runtime dependent due to PIE, ASLR) load bias of your `linux-defmt` source. If you want this pipe with two processes to work, you'd want the defmt source to figure out its own load bias (e.g. like https://github.com/quartiq/defmt2log/blob/fabe90d2a9d358af6821d8c5d85e44134c6654ae/src/table.rs#L41-L78) and send that as a header over the pipe first. That needs some work on the source/encoder side (the load bias determination and new header encoding) and on the `defmt-print` sink side to support both receiving/decoding the in-band load bias and applying it though `DecodeOptions`.

It's straight forward to implement but I don't think it can be done in a strictly backwards compatible way: a new defmt source with load bias header into an old unaware `defmt-print` would potentially be broken in weird ways. It also makes that bias header disproportionately valuable compared to the rest of the stream. The alternatives (linker script to fix the bias or at least its low bytes, turning of ASLR, or a some way to emit it plus a command line option to defmt-print, or having defmt-print guess/search) don't look feasible or convincing. It's trivial for defmt2log because there the encoder and decoder are the same executable and the decoder can determine the load bias.

Having the source side (export::istr()/Str and defmt-stdout or your minimal encoder) do the debiasing doesn't look simple: the defmt source side needs to figure it out, the decoder still needs to build the u16 -> usize map, and 0 index is a pervasive special case (in istr and FormatSequence).

My proposal would be not to try to get defmt-stdout + defmt-print to work with a ASLR/PIE in this PR. The wire format changes at the core of that (be it a bias header or source-side debiasing) are a separate issue.

decoder/src/lib.rs
```diff
@@ -161,6 +161,65 @@ pub struct Table {
...
+    /// Add `address_bias` when mapping table indices into the wire index namespace.
+    pub const fn address_bias(mut self, address_bias: u64) -> Self {
+        self.address_bias = address_bias;
+        self
+    }
+}
+
+/// Reusable index for decoding frames whose wire indices differ from table indices.
+pub struct DecodeIndex {
+    entries: Option<HashMap<u16, Option<usize>>>,
```
jonathanpallant, line 168: Why is this an `Option` and when would it be none?

jonathanpallant:
I was copying from the defmt-semihosting logger, which doesn't use RZCOBs and is known to work (although I may have made an error when building the new logger).

If the decoder cannot decode anything until the slide is known, how can we build an example to demonstrate that this PR does what we want it to do?

decoder/src/lib.rs
jordens: The plain [`decode()`](https://github.com/jordens/defmt/blob/2dc54e9b6ac0eec8978a0cc0890d2decfd57f0aa/decoder/src/lib.rs#L280) does the conversion from wire `u16` to `usize` `Table` key without bias and without lookup. It does not need an [`entries`](https://github.com/jordens/defmt/blob/2dc54e9b6ac0eec8978a0cc0890d2decfd57f0aa/decoder/src/lib.rs#L205) index map.

Option<DecodeIndex> would be worse since the DecodeIndex is the unique point that implements the u16 read and the u16 -> usize mapping, regardless of whether that mapping needs a HashMap and has been checked for collisions or is just widening.

jordens:
When I tried your code above it used the default rzcobs encoding. The current defmt-semihosting logger does call [`start_frame()`](https://github.com/knurling-rs/defmt/blob/6e6dff329fc3ccd599560cc95098cfcce2f86eb1/firmware/defmt-semihosting/src/lib.rs#L84) and `end_frame()`. Also I believe defmt-semihosting does use rzcobs by default, at least I didn't see how it would override the default.

Here's an example.

`logger.rs`:

```rust
use defmt_decoder::{DecodeContext, Locations, Table};
use std::{cell::RefCell, sync::LazyLock};

struct Info {
    table: Table,
    decode_context: DecodeContext,
    locations: Locations,
}

static INFO: LazyLock<Info> = LazyLock::new(|| {
    let elf = std::fs::read(&std::env::current_exe().unwrap()).unwrap();
    let table = Table::parse(&elf).unwrap().unwrap();
    let decode_context = table
        .new_decode_context_for_runtime_anchor(defmt::runtime_anchor() as _)
        .unwrap();
    Info {
        locations: table.get_locations(&elf).unwrap(),
        table,
        decode_context,
    }
});

defmt::timestamp!("");

#[defmt::global_logger]
struct Logger;

thread_local! {
    static THREAD: RefCell<Vec<u8>> = Default::default();
}

unsafe impl defmt::Logger for Logger {
    fn acquire() {}

    unsafe fn write(bytes: &[u8]) {
        THREAD.with_borrow_mut(|thread| thread.extend_from_slice(bytes));
    }

    unsafe fn release() {
        THREAD.with_borrow_mut(|thread| {
            match INFO.table.decode_with_context(thread, &INFO.decode_context) {
                Ok((frame, consumed)) if consumed == thread.len() => {
                    println!("{}", frame.display(false));
                    if let Some(location) = INFO.locations.get(&frame.index()) {
                        println!("  at {} @ {:?}", location.module, location);
                    }
                }
                Ok((_frame, _consumed)) => println!("partial frame"),
                Err(err) => println!("{err:?}"),
            }

            thread.clear();
        });
    }

    unsafe fn flush() {}
}
```

`main.rs`:

```rust
mod logger;

fn main() {
    defmt::println!("Hello, world!");
    defmt::debug!("Foo");
    defmt::warn!("Bar");
}
```

```
$ DEFMT_LOG=debug cargo run
Hello, world!
  at linux_defmt::main @ /home/rj/src/defmt-linux/src/main.rs:4
DEBUG Foo
  at linux_defmt::main @ /home/rj/src/defmt-linux/src/main.rs:5
WARN Bar
  at linux_defmt::main @ /home/rj/src/defmt-linux/src/main.rs:6
```

jordens:
I added aeade3f32716992f04d65fb50ab8d4d44b839e15:

* Add warning log msgs for colliding u16 indices.
* Removes `DecodeOptions` and just uses the bias directly. The struct seemed too heavy.
* Adds `defmt::runtime_anchor()` which yields the runtime address of a well-known anchor symbol.
* Adds support to `Table::new_decode_index_for_runtime_anchor()` to compute the load bias and the decode index given that runtime anchor address (load bias is runtime address minus elf address). That way the two parts (runtime and elf addresses) of the load bias cleanly remain at their respective wire ends (encoder and decoder respectively).
* It makes the example (and also defmt2log) shorter and completely standalone and removes the need for findshlibs/mmap parsing.

With this change we could later introduce a change/addition to the wire format where the encoder subtracts the runtime anchor and the decoder subtracts the elf anchor address. That's more elegant and clean than a wire format header or other way to shuttle the actual load bias. It would enable `defmt-stdout | defmt-print`.

~CI is broken via #1087~

jordens:
> (the defmt::timestamp seems mandatory without the default PROVIDE in the defmt.x file)

#1089 keeps the bitflags metadata and timestamp from being GC-ed. Without it when used with a non-empty `defmt::timestamp!()`, the timestamp data is present in the stream but the decoder doesn't expect it (since it's not in the ELF) and fails with partial/malformed frames.

decoder/src/lib.rs
```diff
@@ -159,6 +159,53 @@ pub struct Table {
...
+                .or_insert(Some(index));
+        }
+
+        Self {
+            entries: Some(entries),
+        }
+    }
+
+    fn read_raw(&self, bytes: &mut &[u8]) -> Result<u16, DecodeError> {
+        Ok(bytes.read_u16::<LE>()?)
```
jonathanpallant, line 200: What is this method for? It doesn't use `self`?

jonathanpallant:
Sorry for the ongoing delays - it's been a busy summer, and it takes a while to get my head back around a PR like this one.

I think I'm just about satisified now, but I was confused by a method that didn't seem to be related to the type it was added to.

jonathanpallant:
Github is also reporting conflicts - I suspect this is in the CHANGELOG (again).

decoder/src/lib.rs
jordens: The idea was to localize this part of the the wire format in a single place. `DecodeIndex` knows how to read the raw index from the wire (this method) and knows how to `resolve()` it into a table key. `Decoder` just drives those. Before this change `Decoder` knew about `read_u16::<LE>()` in three places.
In theory, some future more powerful `DecodeIndex` (using the other potential `DecodeOptions` fields mentioned in https://github.com/knurling-rs/defmt/pull/1066#issuecomment-4798103047) could decide that raw indices are to be read in a different way it. It could then make use of `self`.

jordens:
No problem! Thanks for having a look.
Conflicts resolved.
Re `read_raw()`: let me know. I'm happy to follow your preference (be that as-is, inlining, or making it an associated func).

decoder/src/lib.rs
jonathanpallant: OK, I think some of my confusion is coming from 'index' meaning "a numeric value identifying a defmt string", and 'DecodeIndex' meaning "a look-up table that is used to decode indicies".

I would change `read_raw` to be an associated function, and I would change `DecodeIndex` to be `IndexDecoder` or `IndexLookupMap` or something. And everywhere you read a raw index I would call that variable `let raw_index =`, to try and make the distinction between a raw index (as read off the wire) and a decoded index (after decoding using the HashMap) a bit clearer.

decoder/src/lib.rs
```diff
@@ -227,13 +274,48 @@ impl Table {
...
+    /// the low 16 bits of the corresponding loaded runtime symbol address.
+    /// `load_bias` is the runtime symbol address minus the same symbol's ELF
+    /// address, and is added before truncating to the wire index.
+    ///
+    /// Decoding fails for wire indices that match multiple table entries.
+    pub fn new_decode_index(&self, load_bias: u64) -> DecodeIndex {
+        DecodeIndex::new(self, load_bias)
+    }
+
+    /// Build a reusable index from the runtime address of the defmt anchor.
```
jonathanpallant: Another use of 'index' to mean "a place we look things up" as opposed to "a numeric ID that we can use to look something up".

> Build a reusable index decoder from...

or

> Build a reusable index look-up map from...

jordens: `index` is ambiguous. I'll avoid it. Also found a couple more things that I'm not happy with in hindsight.

jordens:
Looking at the code again I did an overhaul.

* `s/DecodeIndex/DecodeContext/` and avoid the `index` terminology due to ambiguity pointed out.
* inlined `read_raw()` (went back to two occurences of `read_u16::<LE>()`) for simplicity
* (bugfix) Ensure that a u16 "wire index" that -- when biased -- becomes zero is not confused with the magic format sequence terminator.
* Use `DEFMT_VERSION` as anchor, not `DEFMT_ENCODING`, since the former appears more fundamental. And added a test to tie down the behavior.
* `s/elf_address_anchor/image_anchor_address/` since it's not really ELF-specific (despite the `elf2table` name)
* Moved the `u16 -> usize` map as `u16_to_address` into `Table`: With proper u16 modulo math we can just split and commute the biasing and the lookup and don't need (runtime dependent) biased lookup keys. That way `Table` including its `u16_to_address` map stays independent of runtime load bias. The map is now always generated but the cost is tiny compared to the other things going on. And `Table` needs custom serde helpers. Making `Table.entries: HashMap<u16, TableEntry>` would be a radical, semver-breaking but possible change to make the lookup one-step.

PR #1070, opened 2026-06-12
Use `syn::Error`

### LINKED ISSUE #1067: RUSTSEC-2026-0173
`defmt-macros` depends on `proc-macro-error2` which is unmaintained, which in turn now triggers advisory pipelines:

https://rustsec.org/advisories/RUSTSEC-2026-0173

Hoverbear:
Get rid of an unnecessary dependency.

Closes #1067

### DIFF
--- a/macros/src/derives/format.rs
+++ b/macros/src/derives/format.rs
@@ -1,9 +1,8 @@
 use codegen::DefmtAttr;
 use proc_macro::TokenStream;
 use proc_macro2::Span;
-use proc_macro_error2::{abort, abort_call_site};
 use quote::quote;
-use syn::{parse_macro_input, parse_quote, Data, DeriveInput, Generics, Ident, WhereClause};
+use syn::{parse_macro_input, parse_quote, Arm, Data, DeriveInput, Generics, Ident, WhereClause};
 
 mod codegen;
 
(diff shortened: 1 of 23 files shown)


macros/src/function_like/log/env_filter/parse.rs
```diff
@@ -1,6 +1,4 @@
 use defmt_parser::Level;
-#[cfg(not(test))]
-use proc_macro_error2::abort_call_site as panic;
```
Hoverbear, line 3: This is the only place I'm rather unsure about -- Removing this means we fall back to panics, which results in worse errors.

jonathanpallant (dismissed):

macros/src/function_like/log/env_filter/parse.rs
knoellle: As far as I can tell, there were four uses of the panic macro in this file. The first you replaced by a `syn::Error` and the other three are only related to parsing the `DEFMT_LOG` environment variable. For those, generating an error with a span doesn't really make sense anyway I think.

macros/src/derives/format.rs
```diff
@@ -77,22 +86,27 @@ pub(crate) fn expand_transparent(
...
-                    abort_call_site!(
-                        "Transparent format can only be applied \
-                        when all variants have exactly one field."
-                    )
+                    return Err(syn::Error::new(
+                        Span::call_site(),
+                        format!(
+                            "Transparent format can only be applied \
+                            when all variants have exactly one field (got {})",
+                            fields.len(),
```
knoellle: You are measuring the length of the iterator after already calling `next` on it twice, which would cause the number of fields to be underreported.

```suggestion
                            v.fields.len(),
```

macros/src/derives/format.rs
```diff
@@ -77,22 +86,27 @@ pub(crate) fn expand_transparent(
...
                 let mut fields = v.fields.iter();
                 let field = fields.next();
                 let one_or_less = fields.next().is_none();
                 let Some(field) = field.filter(|_| one_or_less) else {
-                    abort_call_site!(
-                        "Transparent format can only be applied \
-                        when all variants have exactly one field."
-                    )
+                    return Err(syn::Error::new(
+                        Span::call_site(),
```
knoellle:
```suggestion
                        v.fields.span(),
```
yields much more useful errors. Also check other `Span::call_site` calls.
```
error: Transparent format can only be applied when all variants have exactly one field (got 0)
  --> src/bin/test.rs:31:8
   |
31 |     Baz(u32, u32),
   |        ^^^^^^^^^^
```
vs. error when using `call_site`:
```
error: Transparent format can only be applied when all variants have exactly one field (got 0)
  --> src/bin/test.rs:27:10
   |
27 | #[derive(defmt::Format)]
   |          ^^^^^^^^^^^^^
   |
   = note: this error originates in the derive macro `defmt::Format` (in Nightly builds, run with -Z macro-backtrace for more info)
```

knoellle (approved):
LGTM

tisonkun:
Thanks for working on this! Just need this today :D

PR #1073, opened 2026-07-06
decoder: gather unmerged defmt sections

jordens:
(Split from #1066)

## Goal

This PR is the first part to enable support for defmt on Linux without any special linker scripts. #1066 is the second part. The goal is to get defmt working-by-default without any linker gymnastics (on linux at least). `defmt2log` exercises that by becoming a thin and robust `defmt` sink and `log` source. Having defmt work on linux without linker scripts in turn lowers the adoption barrier of `defmt` logging in embedded libaries as it upholds the advantages of the defmt format hints while still supporting development and testing on host operating systems. No need for the usual [`fmt.rs`](https://github.com/probe-rs/embedded-test/blob/bb99a651e1ff88df1047802f26f595d3d0288e42/src/fmt.rs) feature-mux to switch between `defmt` and `log` macros and no need to restrict the the logging to the common subset (excluding defmt format hints).

## Status quo

1. The [defmt linker script](https://github.com/knurling-rs/defmt/blob/main/defmt/defmt.x.in) merges the defmt symbols into a single `.defmt` section. That makes the symbols occupy a contiguous address range, aligns them and allows detecting collisions in the u16 wrapped index range already at compile time.
2. `get_locations()` relies on `DW_AT_location address == TableEntry raw symbol address == Table.entries key == Frame::index()`.

## Problem

1. When used without a merging linker script (linker orchestration is extremely platform- and linker-dependent and notoriously tricky to get robust, universal and correct) the symbols are still there (since they are used) just not merged into a single `.defmt` section. The linker is in principle free to scatter them but in practice keeps them contiguous. But they are usually not aligned. On linux at least each `.defmt*` symbol gets its own section.
2. Using `DW_AT_location` is both not needed and breaks in various modes. Some of them are encountered here, in #1066, and in https://github.com/knurling-rs/defmt/pull/1029, https://github.com/knurling-rs/defmt/pull/1040 etc. It only works for a a merged section without a load slide. It fails for Host PIE, shared object loading, stale GC'ed symbols (DWARF can still mention variables whose corresponding defmt symbol was discarded), location expressions that are not simple addresses, non-zero VMA, and non-zero section base addresses.

## Approach

1. This extends the defmt symbol `Table` to include all defmt sections named `.defmt.*`. This then covers both the merged (classic `defmt.x`) as well as the unmerged (linker-script-less, e.g. on Linux) case.

2. It also removes the `DW_AT_location` usage. Instead going through "DWARF linkage name -> table entry raw_symbol -> table index -> Location" using `DW_AT_linkage_name` as the join key is simpler and robust.

* Note that this PR only includes the unmerged sections in the table and fixes the `DWARF/Table/Locations/Frame::index()` join chain. It doesn't cover load bias or runtime index collision detection yet. That's not a regressions in this PR: collision detection is so far the responsibility of the linker script. #1066 detects and handles runtime collisions in the load bias + unmerged sections case.

* Given the new sections Vec, rewrite the tuple match for the `bail!()` prerequisite checks to be a bit more readable.

* Use `BITFLAGS_VALUE_SIZE` instead of magic constant.

## Testing

* Added unittests (most of this PR is unittests).
* Exercised and used for a few months now during development of several crates (`minimq`, `miniconf`, `stabilizer`, and others). Those are mid- to high-complexity defmt users. Both on embedded (defmt-rtt) and on host/CI (defmt2log).

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -626,6 +626,8 @@ Initial release
 
 ### [defmt-decoder-next]
 
+* [#1073] Gather symbols from unmerged `.defmt.*` sections, don't use `DW_AT_location`
+
 ### [defmt-decoder-v1.1.0] (2026-01-20)
 
 * [#1004] decoder: add `Send + Sync` bound to returned `StreamDecoder`
@@ -1033,6 +1035,7 @@ Initial release
 
 ---
 
+[#1073]: https://github.com/knurling-rs/defmt/pull/1073
 [#1070]: https://github.com/knurling-rs/defmt/pull/1070
 [#1053]: https://github.com/knurling-rs/defmt/pull/1053
 [#1055]: https://github.com/knurling-rs/defmt/pull/1055
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -254,6 +254,15 @@ source = "registry+https://github.com/rust-lang/crates.io-index"
 checksum = "773648b94d0e5d620f64f280777445740e61fe701025087ec8b57f45c791888b"
 
 [[package]]
+name = "crc32fast"
+version = "1.5.0"
+source = "registry+https://github.com/rust-lang/crates.io-index"
+checksum = "9481c1c90cbf2ac953f07c8d4a58aa3945c425b7185c9154d67a65e4230da511"
+dependencies = [
+ "cfg-if",
+]
+
+[[package]]
 name = "data-encoding"
 version = "2.9.0"
 source = "registry+https://github.com/rust-lang/crates.io-index"
@@ -421,6 +430,12 @@ source = "registry+https://github.com/rust-lang/crates.io-index"
 checksum = "52051878f80a721bb68ebfbc930e07b65ba72f2da88968ea5c06fd6ca3d3a127"
 
 [[package]]
+name = "foldhash"
+version = "0.1.5"
+source = "registry+https://github.com/rust-lang/crates.io-index"
+checksum = "d9c4f5dac5e15c24eb999c26181a6ca40b39fe946cbe4c263c7209467bc83af2"
+
+[[package]]
 name = "fsevent-sys"
 version = "4.1.0"
 source = "registry+https://github.com/rust-lang/crates.io-index"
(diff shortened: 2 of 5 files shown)


jonathanpallant:
Can you explain in the PR description *why* this PR is required - what the observed issues are currently, and what this PR does to resolve them?

jordens:
done

jonathanpallant:
Right, ok! Now I understand (/remember) what this is about. Thank you!

To give this some more testing, I rebuilt `qemu-run` and used it to run one of my training examples:

```console
$ DEFMT_LOG=trace cargo run
   Compiling defmt-macros v1.1.0
   Compiling defmt v1.1.0
   Compiling defmt-semihosting v0.3.0
   Compiling qemu-common v0.1.0 (/Users/jonathan/Documents/ferrous-systems/rust-training/example-code/qemu-common)
   Compiling qemu-aarch32v8r v0.1.0 (/Users/jonathan/Documents/ferrous-systems/rust-training/example-code/qemu-aarch32v8r)
    Finished `dev` profile [unoptimized + debuginfo] target(s) in 1.24s
     Running `qemu-run --machine mps3-an536 --cpu cortex-r52 --log-format '{[{L}]%bold} {s} {({ff}:{l:1})%dimmed}' target/armv8r-none-eabihf/debug/defmt`
Hello, world!
[ERROR] This is an error log (bin/defmt.rs:10)
[WARN ] This is a warn log (bin/defmt.rs:11)
[INFO ] This is an info log (bin/defmt.rs:12)
[DEBUG] This is a debug log (bin/defmt.rs:13)
[TRACE] This is a trace log (bin/defmt.rs:14)
[ERROR] PANIC at src/bin/defmt.rs:18: Testing panic!() in fn main (src/lib.rs:23)
```

All still seems to work OK.

jonathanpallant (approved):

jonathanpallant:
I think I'm happy with this, but I haven't looked at the ELF decoder code before so I'll ping for a colleague.

jonathanpallant:
OK I gave it a second pass and I'm happy. Thank you for this PR.

PR #1078, opened 2026-07-20
Support named arguments with inline capture in format strings ({owner:?})

aimir:
## Summary

Adds support for named arguments in defmt format strings, mirroring `core::format_args!`:

```rust
let owner = 42u8;
defmt::info!("{owner:?}");                          // inline capture, std Debug syntax works unchanged
defmt::info!("{owner=u8:#x} then {count=u16:#x}");  // defmt types and hints compose with names
defmt::info!("explicit {x=u8}", x = 1);             // explicit `name = expr` arguments
defmt::info!("mixed pos={=u8} named={owner=u8} first={0=u8}", 7);
```

Named parameters without a matching `name = expr` argument capture the identifier from the surrounding scope. Repeated use of a name serializes the argument only once. The feature applies to all log macros, `println!`, `write!` (and therefore `Format` impls), `timestamp!`, and the `assert!`/`panic!` family, which share the codegen path.

## Design

**No wire format or runtime changes.** Names never exist at runtime; they live only inside the interned format string, and the wire still carries the u16 index plus arguments in index order.

- **defmt-parser**: the argument slot of a parameter now accepts an identifier (`argument := integer | identifier`). `Parameter` gained a `name: Option<String>` field (breaking change for defmt-parser/decoder consumers). Named parameters are assigned the argument indices *after* all positional ones, in order of first appearance — the same rule `core::format_args!` uses to append named arguments. This assignment is load-bearing: defmt-macros serializes arguments in this order at compile time and defmt-decoder re-derives the same order from the interned string at decode time, both through `defmt_parser::parse`, so it lives in exactly one place.
- **defmt-macros**: formatting arguments parse as `expr` or `name = expr` (a leading `ident =`, but not `ident ==`, denotes a named argument, like std). A new `resolve_args` step maps each named parameter to its explicit expression or to a capture of the identifier; the capture's span is set to the format string literal, so a missing variable produces rustc's native ``cannot find value `owner` in this scope`` pointing at the format string. Diagnostics for duplicate names, unused named arguments, positional-after-named, and keyword captures (`{fn}`) are included; the existing "format string requires N arguments but only M were provided" wording is preserved for pure-positional format strings.
- **defmt-decoder**: no logic changes (only the added `Parameter::name` field in constructions).

## Compatibility

Old decoders (defmt-print/defmt-decoder built before this change) fail to parse format strings that use the new syntax, producing a per-frame decode error; frames not using named arguments are unaffected. This matches how previous grammar additions (`:cbor`, `:iso8601`) shipped, so this PR does **not** bump the wire format version — happy to change that if you prefer a bump.

## Testing

- New `named_args` QEMU snapshot test exercising capture, hints, repeats, explicit and mixed arguments end-to-end (target-side encoding → host-side decoding).
- Parser unit tests for the new grammar (index assignment, repeats, mixing with positional/explicit indices, `{owner:?}`, error cases like `{_}` and `{owner.field}`).
- Decoder unit test decoding a frame whose format string mixes positional and named arguments, pinning the argument-order contract.
- Compile tests for all macro forms under `unstable-test`.
- `cargo xtask -d test-host --skip-ui-tests`, `cargo xtask test-snapshot` (all bins), and `cargo xtask test-lint` all pass locally (the latter with the clippy fix from #1079, which current clippy requires on `main` too).

## Remaining work (why this is a draft)

- [x] trybuild UI tests for the new error messages (blessed on 1.83, the host MSRV)
- [x] Book documentation (`macros.md` grammar section)
- [x] Decision on wire-format version bump vs. documented soft-compatibility (see above; the backcompat test now skips the named_args snapshot for this reason)
- [ ] Possibly squash the decoder-adaptation commit into the parser commit so every commit builds standalone

🤖 Generated with [Claude Code](https://claude.com/claude-code)

https://claude.ai/code/session_01VWapkqhDJF6z7b3UQ34Qh6

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -60,7 +60,7 @@ We have several packages which live in this repository. Changes are tracked sepa
 
 ### [defmt-next]
 
-* Support named arguments in format strings, like `core::format_args!`: `{owner}`, `{owner:?}` and `{owner=u8:#x}` capture `owner` from the surrounding scope, and arguments can be passed by name (`defmt::info!("{x}", x = expr)`).
+* Support named arguments in format strings. A parameter may now name a variable instead of using a positional index; the variable is captured from the surrounding scope, following the same rules as `core::format_args!`. This means `format!`-style strings such as `"{owner}"` and `"{owner:?}"` work unchanged, and defmt's type and display-hint syntax composes with names: `"{owner=u8:#x}"`. Arguments can also be passed by name: `defmt::info!("{x}", x = expr)`. Decoders older than this change cannot decode log statements that use the new syntax (other log statements are unaffected).
 
 ### [defmt-v1.1.1] (2026-06-26)
 
@@ -711,7 +711,7 @@ Initial release
 
 ### [defmt-parser-next]
 
-* Support named arguments (`{owner}`, `{owner=u8:#x}`) in the format string grammar. Named parameters are assigned the argument indices after all positional ones, in order of first appearance. `Parameter` gained a `name` field (breaking change).
+* Support named arguments (`{owner}`, `{owner=u8:#x}`) in the format string grammar. Named parameters are assigned the argument indices after all positional ones, in order of first appearance. Breaking changes (v2.0.0): `Parameter` gained a `name` field and `Error` gained an `InvalidArgumentName` variant.
 * [#956] Link `LICENSE-*` in the crate folder
 * [#1028] Clarify that MSRV is 1.76
 
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -452,7 +452,7 @@ Initial release
 
 ### [defmt-macros-next]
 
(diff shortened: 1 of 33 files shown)


CHANGELOG.md
```diff
@@ -451,6 +452,8 @@ Initial release
 
 ### [defmt-macros-next]
 
+* Support named arguments with inline capture (`{owner:?}`) and explicit `name = expr` formatting arguments.
```
aimir:
```suggestion
* [#1078] Support named arguments with inline capture (`{owner:?}`) and explicit `name = expr` formatting arguments.
```

jonathanpallant:
I apologise for the delay in getting back to you. This was one of the first major LLM-assisted PR's we have received in this repository and it required a lot of internal discussion.

Unfortunately we have concluded that it is the policy of the knurling project from now on to not accept AI generated contributions. Our policy can be found at https://github.com/knurling-rs#contribution-policy.

I am therefore closing this PR.

PR #1079, opened 2026-07-20
decoder: use slice::from_ref instead of cloning an Arg for a one-element slice

aimir:
Used `Frame::format_args` cloned an `Arg` - which can hold `Vec`s and nested format sequences - solely to build a one-element slice.

Use `std::slice::from_ref(arg)` instead, which produces the `&[Arg]` directly from the existing reference, avoiding the clone.

This is also what clippy suggests in version 1.90 and above (`clippy::cloned_ref_to_slice_refs`, warn-by-default), so `cargo clippy -- -D warnings` on a current toolchain fails on `main` without it.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -626,6 +626,8 @@ Initial release
 
 ### [defmt-decoder-next]
 
+* [#1079] Avoid cloning `Arg`s when formatting format sequences
+
 ### [defmt-decoder-v1.1.0] (2026-01-20)
 
 * [#1004] decoder: add `Send + Sync` bound to returned `StreamDecoder`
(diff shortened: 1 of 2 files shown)


jonathanpallant:
slice::from_reg was added in 1.63, which meets our MSRV. So this seems fine, thanks!

jonathanpallant (dismissed):

jonathanpallant:
I suspect this won't merge because of a conflict on CHANGELOG.md, but let's see.

jonathanpallant (approved):

jonathanpallant:
Sorry, github won't let me rebase your branch. Let me just push this again.

PR #1083, opened 2026-07-27
parser: reject bare-number bitfield specs instead of panicking

Ddystopia:
`parse_range` sliced `s[start_digits..start_digits + 2]` to look for the `..` separator without checking the length first, so a type fragment consisting of a number with fewer than two bytes after it panicked instead of being reported as an invalid type specifier. Affects `{=8}`, `{=0}`, `{=127}` and `{=0.}`.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -711,6 +711,7 @@ Initial release
 
 ### [defmt-parser-next]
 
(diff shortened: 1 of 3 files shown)


jonathanpallant (approved):

jonathanpallant:
Good catch - thanks!

PR #1084, opened 2026-07-27
Fix stray close brace

Ddystopia:
`push_literal` tracked a parity toggle per brace kind rather than checking that each brace is doubled, so two stray braces of the same kind cancelled each other out whenever no plain character sat between them to trip the odd-parity check. `}{{}` was accepted as the literal `}{}`.

Found by sweeping every string over `{`, `}` and `a` up to length 6 through both this parser and rustc's `format!`: of 1092 inputs, 13 were accepted here and rejected by rustc, and the pairing check removes the discrepancy in both directions. All 13 are covered.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -452,6 +452,8 @@ Initial release
 
 ### [defmt-macros-next]
 
+* [#1084] Report a format string that a future release will reject, such as `}{{}`, as a deprecation warning at the macro call site.
+
 ### [defmt-macros-v1.1.1] (2026-06-26)
 
 * [#1070] Swap from `proc-macro-error2` to `syn::Error`
@@ -713,6 +715,7 @@ Initial release
 ### [defmt-parser-next]
 
 * [#1083] A bitfield spec whose leading number is not followed by a complete `..end`, such as `{=8}` or `{=0.}`, now returns `Err(InvalidTypeSpecifier)` instead of panicking.
+* [#1084] Add `parse_with_warnings`, which reports the format strings that a future release will reject. A stray `}` in a literal that a later stray `}` cancels out, such as `}{{}`, is one: `format!` rejects it, so it warns with `Warning::UnmatchedCloseBracket` while still parsing as before.
 * [#956] Link `LICENSE-*` in the crate folder
 * [#1028] Clarify that MSRV is 1.76
 
@@ -1040,6 +1043,7 @@ Initial release
 ---
 
 [#1089]: https://github.com/knurling-rs/defmt/pull/1089
+[#1084]: https://github.com/knurling-rs/defmt/pull/1084
 [#1083]: https://github.com/knurling-rs/defmt/pull/1083
 [#1073]: https://github.com/knurling-rs/defmt/pull/1073
 [#1070]: https://github.com/knurling-rs/defmt/pull/1070
--- a/defmt/tests/ui/deprecated-braces.rs
+++ b/defmt/tests/ui/deprecated-braces.rs
@@ -1,0 +1,16 @@
+#![deny(deprecated)]
+
+fn main() {
+    defmt::info!("}{{}");
+    defmt::println!("}{{}");
+}
+
+struct S;
+
+impl defmt::Format for S {
+    fn format(&self, f: defmt::Formatter) {
+        defmt::write!(f, "}{{}");
+    }
+}
+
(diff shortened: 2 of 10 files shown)


jonathanpallant:
I think this makes sense, but I'd like a second opinion from @knoellle because I suppose there's a risk someone was (accidentally) relying on our bad parsing and we might break existing code with this fix?

knoellle:
I suppose there is a risk of [spacebar heating](https://xkcd.com/1172/).
Effectively, I see three options here: 
1. Use new parser and risk breakage 
2. Just keep the broken parser
3. Keeping both parsers around until the next major release and emit a warning if only the new parser implementation complains.

The proper thing to would probably be 3., but spacebar heating seems unlikely enough to me, so I'm fine with 1. too. We can always revert the change and push out a new patch version if somebody complains.

On the other hand, this diff only affects a tiny piece of code, so extracting it into two functions for 3. is relatively low overhead and still results in relatively clean code.
I`m thinking something like
```rs
// comment explaining why there are two impls here
old_impl(unescaped_literal)?;
if let Err(error) = new_impl(unescaped_literal) {
  // emit warning
};
```

2. is out in my opinion.

Ddystopia:
Note that #1085 is kind of similar

Ddystopia:
@jonathanpallant did you decide which option, 1 or 3?

jonathanpallant:
I just had some production code break because RTIC did an update that started disallowing some code. Technically that code was wrong, but they didn't warn me and just made it a hard break. That is then very difficult to deal with.

I would not want to do that to any defmt user, especially over something as trivial as bracket alignment in a format string. So option 3 is really the only option (and I note that there will hopefully never be another major release of the parser, and we can keep it all backwards compatible).

So the steps are:

1. Detect the bad string and warn about it
2. At some point in the future, after maybe a year, turn the warning into an error

jonathanpallant:
Also, is there anything on the firmware side in the macro to validate that the format string is balanced?

Ddystopia:
> Also, is there anything on the firmware side in the macro to validate that the format string is balanced?

I didn't quite get what you mean. It should not be hard to detect those cases in code and warn about them, if this is what you mean.

Also, how do you want to warn? At runtime? I doubt it will be trivial at all to warn at compile time.

jonathanpallant:
Well the format string is already parsed in the macro at compile time to verify all the format arguments have been provided.

If you compile a firmware with a bad format string (with no warning), put the firmware in production, and then find the parser is failing with an error once a particular message is received over your cellular modem link (a mode of operation that people are using) then you'd be pretty annoyed.

knoellle:
IMO both the macro and printer should warn when the parser detects a "deprecated"/broken format string.

Ddystopia:
@jonathanpallant @knoellle `cortex-m-rt` published 0.7.6 with MSRV bump and CI resolves to it automatically (Cargo.lock isn't commited and --locked isn't used), but defmt-MSRV rustc can't even parse the manifest and fails. You should either raise MSRV in sync with `cortex-m-rt` or pin cortex-m-rt. But it is certainly out of this PR's scope.

https://github.com/rust-embedded/cortex-m/blob/master/cortex-m-rt/CHANGELOG.md

Should I wait for you and rebase?

Ddystopia:
> IMO both the macro and printer should warn when the parser detects a "deprecated"/broken format string.

Btw, which of you makes the final decision I should follow?

knoellle:
Sorry for the wait, after some discussion about versioning and compatibility, we have decided to leave the warning generation in the macro only to avoid API issues when turning the warning into a hard error later.
`parse_with_warnings` has to be `pub` because `pub(workspace)` isn't a thing and we need to use it from the macro. But we don't want this in our public API, so I suggest to make it `#[doc(hidden)]`, and `#[deprecated]` while selectively ignoring the deprecation warning at the macro callsite.

- `cortex-m-rt` has been pinned in #1088.
- `cortex-m` has been pinned in #1093.

Ddystopia:
Okay, do I understand my next actions correctly?
- Rebase against master
- Rebase to make `parse_with_warnings` `#[doc(hidden)]`
- *Not* add any runtime logs

That's it? If yes, I'll do that later, as well as with the other PR.

knoellle:
Yes, that seems correct to me. Thank you

Ddystopia:
@knoellle It should be ready, along with #1085

knoellle (approved):

Issue #1080, opened 2026-07-22
Feature request: support on-device rendering of defmt frames (deferred to a low-priority task)


Ddystopia:
## Motivation

In real-time firmware, replacing `log`-style formatting with defmt removes three costs from high-priority tasks: formatting cycles on the real-time path, formatting stack (which can be large, and every high-priority task's stack has to be sized for it), and blocking on the transport like UART or RTT. With defmt, a log call is just an index plus raw argument bytes into a RAM queue, so the high-priority path stays cheap and shallow, and because the encoding is compact the same buffer holds far more messages before anything is dropped. A low-priority task drains the queue when there is time.

The gap is what that low-priority task can do with the frames. Today the only supported option is shipping them off-device to a host decoder that has the matching ELF. But in many real-time deployments the natural place to finish the job is on the device: the idle-time task renders the frames to text into an output buffer, and from there the text goes out through whatever slow peripheral the product uses. The deferral that matters here is across *priority levels*, not across the device boundary. defmt's architecture already does the hard part, it just can't complete the last step locally.

Shipped products illustrate this. A device that outputs readable text can be debugged with any serial terminal, the widespread tools everyone already has. The alternative is dedicated software for this one device that picks the right ELF for whatever firmware version a unit is running, or offloading that ELF matching problem to the user. And sometimes getting the logs *quickly* is critical. On-device rendering gives products that field story while the firmware internally keeps all of defmt's benefits on the high-priority path.

So the use case is: keep defmt's call-site cost and wire compactness exactly as they are, accept paying flash for the format strings, and do the `defmt-decoder` work on-target at low priority, as well as wait for the transport.

Related ground has been covered in #463 (non-embedded targets), #854 (metadata outside the ELF), and the std-rendering idea in #888, but this exact shape, a no_std, no_alloc decoder running on-target with formatting deferred to a low-priority task, doesn't seem to have been written down as a request.

## What the code shows

`defmt-parser` is nearly no_std already. ~640 lines, single dependency (`thiserror`, which supports no_std in v2). The only `std` items are `Cow`, `Range`, `FromStr`, and `ParseIntError`, all of which are available from `core` or replaceable. Porting is mostly `use std::` to `use core::`.

[...]

PR #1085, opened 2026-07-27
Fix param body rescan

Ddystopia:
The scan loop never advanced past a parameter it had just consumed, so it resumed one byte after the opening `{` and treated any `{` in the body as the start of another parameter. `{:{}` produced two parameters terminated by the same `}`, inventing an argument that the format string does not have.

Found by sweeping every string over `{`, `}`, `:` and `a` up to length 5 in `ForwardsCompatible` mode: 9 of 1364 inputs gained a phantom parameter or reported an error from re-parsing the body.

Only the last commit is a part of this PR, others are for easy fast forward merge (kind of stacked PR).

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -452,6 +452,8 @@ Initial release
 
 ### [defmt-macros-next]
 
+* [#1084] Report a format string that a future release will reject, such as `}{{}`, as a deprecation warning at the macro call site.
+
 ### [defmt-macros-v1.1.1] (2026-06-26)
 
 * [#1070] Swap from `proc-macro-error2` to `syn::Error`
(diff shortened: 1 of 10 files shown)


jonathanpallant:
Thank you for this PR and apologies for the delay in replying.

I'm generally very wary about changing the parser in a way that might break existing code. Arguably if the parser was broken in defmt 1.0, then it should remain that way from now on.

Do you have examples of code that *should work* but currently *does not* work that this PR would fix? Or is only about rejecting code that currently *does work* but *should not work*?

Ddystopia:
This is the same kind as my #1084.

jonathanpallant:
Oh right, sorry. We've been around this loop already in that PR.

knoellle:
I'm not sure I understand the purpose of this PR. I think it improves the parser logic, but just to confirm: the warning added in this PR is never shown to users of the macro nor the decoder, right?
The macro uses the `Strict` parsing mode which already rejected the affected format strings previously and while the decoder uses `ForwardsCompatible` mode, it discards the warnings as discussed in #1084.

Ddystopia:
The purpose of these two PR's is: I created an implementation, as requested, for #1080 , and it includes no_std parser that is a lot faster (on host the decoding speed doesn't really matter, but for no_std I needed to improve the performance *a lot*), and I found a couple of bugs. I decided to first upstream bug fixes, then to pr the no_std parser impl and other stuff, maybe.

PR #1089, opened 2026-08-03
macros: add #[used] to timestamp and bitflags metadata

jordens:
Without a linker script, currently the timestamp and bitflags metadata is optimized away during linking since nothing references it.

This PR adds `#[used]` to those symbols.

In the case of no special linker script this prevents the symbols from being GC-ed. When used with a linker script (which already KEEPs them) the flag doesn't change the behavior.

This works with Rust >= 1.89 when `#[used]` [started to mean](https://github.com/rust-lang/rust/pull/140872) `#[used(linker)]` (the latter remaining unstable as of 1.97).

`DEFMT_{ENCODING,VERSION}` already carry `used`.

Bitflags metadata doesn't get the used flag on msvc because the resulting COFF `/INCLUDE:`  argument [breaks](https://github.com/knurling-rs/defmt/actions/runs/30815407100/job/91691813156#step:5:859) due to json symbol names.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -60,6 +60,7 @@ We have several packages which live in this repository. Changes are tracked sepa
 
 ### [defmt-next]
 
+* [#1089] Retain timestamp and bitflags metadata when linking without `defmt.x`.
 * [#1068] Adding `Format` impl for `core::str` errors
 
 ### [defmt-v1.1.1] (2026-06-26)
@@ -1038,6 +1039,7 @@ Initial release
 
 ---
 
+[#1089]: https://github.com/knurling-rs/defmt/pull/1089
 [#1083]: https://github.com/knurling-rs/defmt/pull/1083
 [#1073]: https://github.com/knurling-rs/defmt/pull/1073
 [#1070]: https://github.com/knurling-rs/defmt/pull/1070
--- a/macros/src/items/bitflags.rs
+++ b/macros/src/items/bitflags.rs
@@ -84,7 +84,9 @@ fn codegen_flag_statics(input: &Input) -> Vec<TokenStream2> {
                 #(#cfg_attrs)*
                 // These values are decoder metadata and have no runtime reference that would
                 // otherwise keep them when no defmt linker script is used.
-                #[used]
+                // MSVC represents `#[used]` as `/INCLUDE:<symbol>`, but `link.exe` fails to
+                // resolve defmt's JSON symbol names when they are forced this way.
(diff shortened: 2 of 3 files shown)


jonathanpallant (dismissed):
I think this makes sense to me, but letting @knoellle decide whether to merge or not.

CHANGELOG.md
```diff
@@ -1040,6 +1041,7 @@ Initial release
 
 [#1083]: https://github.com/knurling-rs/defmt/pull/1083
 [#1073]: https://github.com/knurling-rs/defmt/pull/1073
+[#1089]: https://github.com/knurling-rs/defmt/pull/1089
```
knoellle: This list is (almost) sorted, please keep it that way :)

knoellle:
> This works with Rust >= 1.89

Does this mean native linux builds don't work properly below this version?

We currently have two MSRVs in this project, 1.76 for cross-compilation and 1.83 for host crates, e.g. decoder.
This change means we either introduce a third MSRV or bump the host-MSRV to 1.89... or we ignore this for now while native Linux builds are not fully supported yet. I'd prefer having it documented somewhere though. @jonathanpallant opinions?

jordens:
> Does this mean native linux builds don't work properly below this version?

Native linux (i.e. "target", the defmt source side) builds work fine as long as at least one of the following holds:

* No timestamp metadata (`timestamp!("")`, see https://github.com/knurling-rs/defmt/pull/1066#issuecomment-5167239362)
* metadata is present in the ELF:
  * linker script with `KEEP` OR
  * this PR plus #1066 plus Rust >= 1.89

Rust < 1.89 or linker script is not affected (negatively or positively) by this PR. Nor is the decoder (host side) in any scenario.

If you want to phrase it this way: "#1066 plus no linker script plus linux plus non-trivial timestamp depends on this PR plus Rust >= 1.89". But in any case, this PR is strictly increasing support, not decreasing. I wouldn't see this as introducing another MSRV.

Could well be that this PR also resolves issues in other PRs (of which there are several that pertain to linker scripts and other platforms) but I didn't check that.

jordens:
And also the same argument holds for the `#[used]` usage on both [`DEFMT_VERSION`](https://github.com/knurling-rs/defmt/blob/2e300dfb6b948a3b732926e3f5d81b472d3a53ef/defmt/src/lib.rs#L34) and `DEFMT_ENCODING` which has always depended on rust >= 1.89 to [become relevant](https://github.com/rust-lang/rust/pull/140872).

knoellle (approved):

PR #1096, opened 2026-08-18
defmt: implement Format for alloc Errors

wojciech-graj:
The following errors have been omitted because they do not expose a way to get the information required to make the Format implementation consistent with the Display implementation: FromUtf16Error, FromVecWithNulError, TryReserveError.


jonathanpallant:
I worry that this will drag in the alloc crate for projects that aren't currently using the alloc crate, making compile times longer.

Perhaps we should gate this behind a feature flag? Or, can you show that it doesn't affect compile times or binary sizes?

wojciech-graj:
> I worry that this will drag in the alloc crate for projects that aren't currently using the alloc crate, making compile times longer.

defmt::impls::alloc_ is already gated behind the alloc feature, so this shouldn't be an issue.

https://github.com/knurling-rs/defmt/blob/main/defmt/src/impls/mod.rs#L31-L32

jonathanpallant:
fair point!

Just the conflicts to resolve then, but otherwise looks OK.

jonathanpallant (approved):

PR #1102, opened 2026-09-21
Bump syn to v3

### LINKED ISSUE #1092: Upgrade to syn v3
`defmt` currently depends on `syn` v2. Upgrading to `syn` v3 would help reduce duplicate `syn` versions in downstream dependency trees.

knoellle:
Fixes #1092 

Some syn v2 still remains in dependency tree, but all transitive deps of `defmt` use v3 now.
syn v2 is still part of `defmt`'s dev dependencies and is also used for `defmt-print` & `defmt-parser`.


Hoverbear (approved):

xtqqczze:
Is this intended for `defmt` v1.1.2?

PR #1103, opened 2026-09-22
Drain cleanup

knoellle:
Found an old TODO in the decoder saying to refactor that part once `Vec::drain_filter` was stabilized.
There is no `drain_filter` but there is `retain`, which I think is identical except that the predicate is flipped.

Also removed an unnecessary (de)reference nearby.


decoder/src/decoder.rs
```diff
@@ -294,22 +294,7 @@ fn merge_bitfields(params: &mut Vec<Parameter>) {
...
-                        } else {
-                            i += 1; // we haven't removed a bitfield -> move i forward
-                        }
-                    }
-                    _ => {
-                        i += 1; // we haven't removed a bitfield -> move i forward
-                    }
-                }
-            }
+            params.retain(|param| !matches!(param.ty, Type::BitField(_)) || param.index != index);
```
jonathanpallant: Perhaps flipping this boolean might read better?

```suggestion
            params.retain(|param| !(matches!(param.ty, Type::BitField(_)) && (param.index == index)));
```

or even

```suggestion
            let should_remove = |param| matches!(param.ty, Type::BitField(_)) && (param.index == index);
            params.retain(|param| !should_remove(param));
```

knoellle: Me likey!
I used the second form.

jonathanpallant (approved):

PR #1104, opened 2026-09-22
Update printers page

### LINKED ISSUE #1063: printers page is out of date
https://defmt.ferrous-systems.com/printers is out of date - it should mention probe-rs, and note that qemu-run is now on crates.io.

knoellle:
This PR
- adds `probe-rs` to the list of printers as a preferred option
- marks `probe-run` as deprecated
- applies git warning to all tools
- redirects `qemu-run` link to crates.io
- fixes #1063


book/src/printers.md
```diff
@@ -3,17 +3,27 @@
 *Printers* are *host* programs that receive log data, format it and display it.
 The following printers are currently available:
 
-- [`probe-run`], parses data sent over RTT (ARM Cortex-M only).
-  > 💡 If you are using the git version of defmt, make sure you also install the tool from git and not crates.io.
-  
-  Since v0.3.3, `probe-run` has now a [`--json`] flag to format the output. The main goal of `--json` is to produce machine readable output, that can be used to changing the human-readable format, a question [addressed here] for example.
+> 💡 If you are using the git version of defmt, make sure you also install the printer from a compatible git source and not crates.io.
```
jonathanpallant: Our wire format is stable, so I don't *know* that this is required?

book/src/printers.md
```diff
@@ -3,17 +3,27 @@
...
+
+- [`probe-rs`], comprehensive embedded toolkit, decoding and displaying defmt data sent over RTT.
+
+  > 💡 This is the recommended way to get started with embedded Rust and defmt.
 
-- [`defmt-print`], a generic command-line tool that decodes defmt data passed into its standard input.
 - [`qemu-run`], parses data sent by QEMU over semihosting (ARM Cortex-M only).
```
jonathanpallant: No longer Cortex-M only - also works on Cortex-R and Cortex-A

book/src/printers.md
```diff
@@ -3,17 +3,27 @@
...
 
+  Available on crates.io; install with `cargo install qemu-run`.
+  Requires `qemu-system-arm` to be installed separately.
+
+- [`defmt-print`], a generic command-line tool that decodes defmt data passed into its standard input.
+- [`probe-run`], parses data sent over RTT (ARM Cortex-M only).
+
+  **Deprecated:** Use [`probe-rs`] instead.
+
+  Since v0.3.3, `probe-run` has now a [`--json`] flag to format the output. The main goal of `--json` is to produce machine readable output, that can be used to changing the human-readable format, a question [addressed here] for example.
```
jonathanpallant: I think we can remove this note?

book/src/printers.md
```diff
@@ -3,17 +3,27 @@
...
+
+- [`probe-rs`], comprehensive embedded toolkit, decoding and displaying defmt data sent over RTT.
+
+  > 💡 This is the recommended way to get started with embedded Rust and defmt.
 
-- [`defmt-print`], a generic command-line tool that decodes defmt data passed into its standard input.
 - [`qemu-run`], parses data sent by QEMU over semihosting (ARM Cortex-M only).
-  > 💡 Used for internal testing and won't be published to crates.io
 
+  Available on crates.io; install with `cargo install qemu-run`.
```
jonathanpallant, line 12: There's a note somewhere else in this tree that qemu-run is for internal use only - we should find that and fix it

book/src/printers.md
knoellle: True, but there may be newer features or changes to the interning symbol format.
We guarantee that old ELF/wire-formats can be read by newer printer tools, but do we also guarantee that old tools can read all new binary formats forever?

I'd be happy to move this to the bottom of the list.

jonathanpallant: I think it's OK to say that you *may* need to use a git version printer rather than *make sure* you use one.

book/src/printers.md
knoellle: The closest I could find was this:
https://github.com/knurling-rs/defmt/blob/main/book/src/setup.md?plain=1#L51

Rephrased slightly

book/src/printers.md
knoellle: Removed

book/src/printers.md
knoellle: Removed caveat

book/src/printers.md
jonathanpallant: OK, no, it was *this* note I was thinking of.

book/src/printers.md
knoellle: Weakened claim and move to the bottom of the page.

book/src/setup.md
```diff
@@ -48,7 +48,7 @@ The following `global_logger`s are provided as part of the project:
 
 - [`defmt-rtt`], logs over RTT. Note that this crate can *not* be used together with `rtt-target`.
 - [`defmt-itm`], logs over ITM (Instrumentation Trace Macrocell) stimulus port 0.
-- [`defmt-semihosting`], logs over semihosting. Meant only for testing `defmt` on a virtual Cortex-M device (QEMU).
+- [`defmt-semihosting`], logs over semihosting. Meant for testing on a virtual Cortex-M device (QEMU).
```
jonathanpallant: > Meant for testing on a virtual device (e.g. on QEMU) where semihosting has acceptable performance. Not recommend on real hardware.

knoellle: Done

jonathanpallant (approved):

